PR #1, opened 2014-04-30
Windows batch file

dgsmith2:
Some folks may want/have to develop on Windows


henryju:
Outdated since we are now relying on official Gradle plugin from Jetbrains for the build.

PR #2, opened 2014-04-30
Support the presence of sonar-project.properties

dgsmith2:
Allows doing things such as specifying sonar.java.source to override the default of 1.5 for PMD


henryju:
This is now possible in SonarLint for IntelliJ to pass additional properties to the analysis so 4m closing this P/R.

PR #8, opened 2016-02-16
SLI-43 NPE when reporting global issue

dbmeneses:
unfortunately it's hard to test without using intellij testing framework


henryju:
LGTM

PR #18, opened 2016-04-18
Increased maximum length of SonarQube server configuration items

aguther:
I tested the new version 2.0.0 and discovered that with that version we are unable to use our SonarQube in the company due to the restriction of the Server URL to 50 characters.

Therefore this pull-request to increase this. What I asked myself is the reason to limit this setting to 50. Is this a limitation of the core library?


henryju:
@aguther @dbmeneses AFAIK there is no limit in sonarlint-core, so we can even consider the 150 limit to be an issue. Why having a limit in the first place?

dbmeneses:
For most fields, the limit is just to sanitize the input and avoid very long names that could be a problem for certain parts of the UI.
For the name, the limit is important as it is used to name a directory and OSs have limits for path lengths (and also because it's displayed in several places in the UI). Could you change the P/R and only set the 50-char length limit to the name?

aguther:
I've made the changes and additionally renamed the variable to be more speaking.

Additional info:
I also made a short check on the eclipse plugin but there seems no limitation.

dbmeneses:
Thanks @aguther

PR #19, opened 2016-04-20
get component from Application

meiping:


dbmeneses:
Thanks for the contribution. The fix was done at the same time for 2.0.1.

PR #22, opened 2016-04-29
Set read timeout when connecting to SonarQube server to 30 seconds to…

runeks:
… support slow response time from api/plugins/installed endpoint

Our SonarQube installation uses for some reason about 20 seconds to respond to a request to http://sonar.eva.lokal/api/plugins/installed.   Increasing the default timeout makes it possible for us to bind to our SonarQube server and fetch settings.


dbmeneses:
Thanks but we are already on track to change this (under review: https://jira.sonarsource.com/browse/SLI-70), which includes also the creation of tokens: https://github.com/SonarSource/sonarlint-intellij/commit/e351cdf718c8393be830ce5913fa59084b72bcd5

PR #23, opened 2016-04-29
Read timeout as a server configuration value

runeks:
Added possibility to configure read timeout on server configuration in case it is needed..  default value is used if not set.


dbmeneses:
The timeout had already been set to 30s. 
I don't see an interest in making it configurable - we should try to make things work with minimal configuration.

runeks:
That's ok, 30s should be sufficient.  Thanks for developing a useful plugin!

PR #29, opened 2016-07-12
SLI-94 Do not loose issues when analysis an unparseable file

dbmeneses:


src/main/java/org/sonarlint/intellij/issue/IssueProcessor.java
```diff
@@ -83,22 +81,27 @@ public void process(final SonarLintAnalyzer.SonarLintJob job, final Collection<I
...
    * Transforms issues and organizes them per file
    */
-  private Map<VirtualFile, Collection<IssuePointer>> transformIssues(Collection<Issue> issues, Collection<VirtualFile> analysed) {
+  private Map<VirtualFile, Collection<IssuePointer>> transformIssues(Collection<Issue> issues, Collection<VirtualFile> analysed, Collection<ClientInputFile> failedAnalysisFiles) {
     Map<VirtualFile, Collection<IssuePointer>> map = new HashMap<>();
+    Set<VirtualFile> failedVirtualFiles = failedAnalysisFiles.stream().map(f -> ((VirtualFile) f.getClientObject())).collect(Collectors.toSet());
 
-    for (VirtualFile f : analysed) {
-      map.put(f, new ArrayList<IssuePointer>());
+    for(VirtualFile f : analysed) {
```
henryju: Formatting issue

travis.sh
```diff
@@ -15,6 +15,7 @@ function strongEcho {
   echo "================ $1 ================="
 }
 
+build_snapshot "SonarSource/sonarlint-core"
```
henryju: Please use a QA build of sonarlint-core before merging (in fact I think we should stop using SNAPSHOTs except locally).

henryju:
Few comments but LGTM

PR #31, opened 2016-07-18
SLI-86 When analysing a source file for the first time, issue date sh…

dbmeneses:
…ould be set to '-' and not to 0s


henryju:
LGTM

PR #33, opened 2016-07-19
Log debug analysis trigger type

dbmeneses:


src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalConfigurable.java
```diff
@@ -67,6 +69,17 @@ public SonarLintGlobalConfigurable() {
     serverManager.reloadServers();
   }
 
+  @CheckForNull
+  public SonarLintGlobalSettings getCurrentSettings() {
+    if(serversPanel != null) {
```
henryju: Code style

henryju:
Code style issue but LGTM

PR #35, opened 2016-07-21
Update Gradle plugin "org.jetbrains.intellij" to 0.30.0

Godin:
This allows to remove special configuration for JaCoCo.


Godin:
`./gradlew clean test jacoco` before this change:

before

after:

after

henryju:
Merged, thanks!

PR #38, opened 2016-08-22
SLI-93: Replace getModules by Extension Point API.

Johnnei:
The JavaAnalysisConfigurator was never detected which caused that the
bytecode would NOT be supplied to the Sonar Java plugin which causes
more FPs and FNs to occur.

### DIFF
--- a/src/main/java/org/sonarlint/intellij/analysis/SonarLintAnalyzer.java
+++ b/src/main/java/org/sonarlint/intellij/analysis/SonarLintAnalyzer.java
@@ -19,6 +19,7 @@
  */
 package org.sonarlint.intellij.analysis;
 
+import com.intellij.openapi.extensions.ExtensionPointName;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.roots.ContentEntry;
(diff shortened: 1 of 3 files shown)


Johnnei:
Hi,

We've been running into an issue which was introduced with IntelliJ SonarLint 2.3. Some of my team started noticing FPs on squid:UnusedPrivateMethod which don't appear in the SonarQube results. Today a post was made over at SonarQube's group with a similar report with a reproducer: https://groups.google.com/forum/#!topic/sonarqube/56TMG6Hyyxk
So armed with the reproducer I started bisecting on the repository since the 2.2 release. At the end the bug was introduced with https://github.com/SonarSource/sonarlint-intellij/commit/58352a5256c8fe6cc32bc47363013aff368b779b 
Once that was down (and confirmed) I started comparing the verbose logs of the Analysis between SLI-93 and the commit before that and noticed that the bytecode paths were no longer supplied. So I've dug through the new system to figure out what was going wrong which in the end lead me to the `module.getComponents(AnalysisConfigurator.class);` call. The issue manifested in this call by NOT returning the `JavaAnalysisConfigurator` in the array and thus the binaries paths were not configured. 
The `#getComponents()` call is deprecated with the hint to use extension points instead. I've been looking through the documentation of that and replaced the deprecated call. After that change the JavaAnalysisConfigurator is correctly detected and returned in the list.
I haven't tested this with CLion so that will still be up to you to see if the intented fix of SLI-93 is still working.

dbmeneses:
Thanks @Johnnei. I will check it and do all tests today.

dbmeneses:
I've tested it on CLion and IntelliJ 15 and 2016.2 - working fine.
I made a small change on it and merge it. Thanks @Johnnei !

PR #42, opened 2016-10-12
Adds Modified files scope.

deepy:


janos-ss:
Hi! Thank you for this contribution. We're considering to include it. We'll keep you posted.

dbmeneses:
@deepy Thanks again for the suggestion. We have specified and implemented the feature: https://jira.sonarsource.com/browse/MMF-512
Having a "changed files" scope wasn't feasible for large change sets because of the work required to keep it up-to-date in all situations, so we dropped the concept of scopes and introduced a new tab, which only displays results on-demand.

PR #47, opened 2016-10-18
SLI-83 fix matching by line hash

janos-ss:
* Rename line-hash to text-range-hash in the old matchers, as that's what they really did
* Add step 6: match issues with same rule, same line and same line hash
* Add step 7: match issues with same rule and same same line hash


src/main/java/org/sonarlint/intellij/issue/LocalIssuePointer.java
```diff
@@ -19,22 +19,33 @@
...
+import org.apache.commons.codec.digest.DigestUtils;
 import org.sonarsource.sonarlint.core.client.api.common.analysis.Issue;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 public class LocalIssuePointer implements IssuePointer {
+
   private static final AtomicLong UID_GEN = new AtomicLong();
+
+  private static final MessageDigest md5Digest = DigestUtils.getMd5Digest();
```
SonarTech: CRITICAL Use a stronger hashing algorithm than MD5. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AS2070)

SonarTech:
SonarQube analysis reported 1 issue
* CRITICAL 1 critical

Watch the comments in this conversation to review them.

src/main/java/org/sonarlint/intellij/issue/LocalIssuePointer.java
janos-ss: No can do!

dbmeneses: naming convention is to call it something like MD5_DIGEST

PR #53, opened 2016-10-21
Fix default values from protobuf

dbmeneses:


src/main/java/org/sonarlint/intellij/issue/LocalIssueTrackable.java
```diff
@@ -49,12 +53,14 @@ public LocalIssueTrackable(Sonarlint.Issues.Issue issue) {
...
+  @CheckForNull
   @Override public String getServerIssueKey() {
-    return issue.getServerIssueKey();
+    return issue.getServerIssueKey().isEmpty() ? null : issue.getServerIssueKey();
   }
 
+  @CheckForNull
   @Override public Long getCreationDate() {
-    return issue.getCreationDate();
+    return issue.getCreationDate() != 0 ? issue.getCreationDate() : null;
```
janos-ss: The methods with the ternary operators are inconsistent with each other. `getServerIssueKey` returns `null` in the `if` branch, but `getCreationDate` and `getLine` return `null` in the `else` branch. I'd align them all, consistently, but it's a minor thing.

PR #55, opened 2016-10-24
SLI-111 Changing/Updating binding should trigger an analysis of all o…

dbmeneses:
…pen files

### DIFF
--- a/src/main/java/org/sonarlint/intellij/util/SonarLintAppUtils.java
+++ b/src/main/java/org/sonarlint/intellij/util/SonarLintAppUtils.java
@@ -1,0 +1,47 @@
+/*
+ * SonarLint for IntelliJ IDEA
+ * Copyright (C) 2015 SonarSource
+ * sonarlint@sonarsource.com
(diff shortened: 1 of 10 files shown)


src/main/java/org/sonarlint/intellij/util/SonarLintAppUtils.java
```diff
@@ -0,0 +1,47 @@
...
+public class SonarLintAppUtils extends ApplicationComponent.Adapter {
+  protected SonarLintAppUtils() {
+    super();
+  }
+
+  public boolean shouldAnalyzeAutomatically(@Nullable VirtualFile file, @Nullable Module module) {
+    return SonarLintUtils.shouldAnalyzeAutomatically(file, module);
+  }
+
+  public boolean shouldAnalyze(@Nullable VirtualFile file, @Nullable Module module) {
```
janos-ss: Who would call this with a `null` file? The question "should analyze?" makes no sense if there is no file to analyze.

PR #58, opened 2016-10-27
SLI-109 SonarLint analyzers should rely on IntelliJ VFS

dbmeneses:


build.gradle
```diff
@@ -68,8 +68,8 @@ configurations {
 }
 
 dependencies {
-  compile 'org.sonarsource.sonarlint.core:sonarlint-client-api:2.5.0.782'
-  compile 'org.sonarsource.sonarlint.core:sonarlint-core:2.5.0.782'
+  compile 'org.sonarsource.sonarlint.core:sonarlint-client-api:2.5-SNAPSHOT'
+  compile 'org.sonarsource.sonarlint.core:sonarlint-core:2.5-SNAPSHOT'
```
janos-ss: Don't forget to update that

src/main/java/org/sonarlint/intellij/analysis/DefaultClientInputFile.java
```diff
@@ -50,6 +59,21 @@
     return charset;
   }
 
+  @Override public InputStream inputStream() throws IOException {
+    if(doc == null) {
```
janos-ss: A later commit will probably put a space in the `if(`, which is noise... Better fix it now ;-)

PR #67, opened 2016-11-21
Windows

dbmeneses:


dbmeneses:
Yes, good catch. With test licenses are not checked.

PR #80, opened 2016-12-08
More improvements

dbmeneses:
Won't compile until we have SLCore and SQ QA'ed

### DIFF
--- a/src/main/java/org/sonarlint/intellij/util/SonarLintUtils.java
+++ b/src/main/java/org/sonarlint/intellij/util/SonarLintUtils.java
@@ -36,6 +36,7 @@ import com.intellij.openapi.vfs.VfsUtil;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.util.net.HttpConfigurable;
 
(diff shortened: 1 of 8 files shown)


src/main/java/org/sonarlint/intellij/core/ServerUpdateTask.java
```diff
@@ -29,6 +29,7 @@
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.ui.Messages;
 
+import com.intellij.util.net.ssl.CertificateManager;
```
janos-ss: Unused import?

src/main/java/org/sonarlint/intellij/util/SonarLintUtils.java
```diff
@@ -36,6 +36,8 @@
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.util.net.HttpConfigurable;
 
+import com.intellij.util.net.ssl.CertificateManager;
+import com.intellij.util.net.ssl.ConfirmingTrustManager;
```
janos-ss: Unused import?

PR #105, opened 2017-05-08
Several tickets

dbmeneses:


src/main/java/org/sonarlint/intellij/tasks/ServerUpdateTask.java
```diff
@@ -106,6 +119,37 @@ public void run(@NotNull ProgressIndicator indicator) {
...
+        .append(", minimum: ")
+        .append(p.minimumVersion())
+        .append(")");
+    }
+
+    return builder.toString();
+  }
+
+  private static boolean tooOld(SonarAnalyzer analyzer) {
+    if (analyzer.minimumVersion() != null && analyzer.version() != null) {
```
janos-ss: When would an analyzer not have a version?

dbmeneses: In the case where we fail to parse it from the filename.

PR #107, opened 2017-05-10
Added server issues panel

asputa:
- view sonarQube server issues
- performance improvement
- other languages support (css,xml,hmtl,...) - server issues
- comparing with server issues


janos-ss:
Thank you @asputa for this contribution! It's an interesting idea. However, providing great UX, as well as a usage pattern that is in line with our long-term vision is no simple matter, and it takes careful planning on our part.

Someday we may implement similar functionality, to let users view and perhaps even manage their SonarQube issues from SonarLint. At the moment we don't have the bandwidth, but this is definitely on our radar. If and when we do that, I don't know if we'll be able to use this contribution, but we'll certainly keep it in mind. I close this PR for now.

PR #112, opened 2017-06-09
SLI-199 Don't save files before analysis in standalone mode

janos-ss:
It seems too simple, am I missing something, or is there a better way?


dbmeneses:
Doesn't the python plugin need to be updated?

janos-ss:
@dbmeneses Good call, updated, and Java too while at it.

PR #120, opened 2017-07-25
Add mnemonics to the Review dialog

tomsquest:
The current Review dialog cannot be accessed using the keyboard.
This commit introduces **mnemonics**, the same as found in the Intellij Commit/Review dialog.

Screenshot of the change:

20170725_selection_003

Default Intellij commit dialog:

20170725_selection_002


dbmeneses:
That's great, thanks!
Ticket for it: https://jira.sonarsource.com/browse/SLI-229

PR #121, opened 2017-10-06
Create librerias

kathe1102:


dbmeneses:
Hi @kathe1102, I've no idea what is this about. If you think it's something that should be merged, please re-open with an explanation about what it does and what is the goal.

PR #134, opened 2018-02-26
Extend read timeout to 10 minutes

treyder:
For slow SonarQube server and big projects (we have around 4 mil. lines of code) SonarLint times out. Can we extend timeout to 10 minutes?


treyder:
@janos-ss could you review the change? Thank you very much in advance. We need to extend timeout to be able to use SonarLint on our project.

treyder:
@dbmeneses could you also review the change?

dbmeneses:
Thanks @treyder

PR #141, opened 2018-04-04
Made RulePanel allow horizontal scrolling

synthomat:
For some reason horizontal scrolling was disabled in the Rule Panel. I enabled it as some rules have quite extensive text that goes beyond the window borders. Having word wrap on HTML Editor Panes seems to be more complicated.


janos-ss:
Oops, had to revert this :( The problem is that this change _removes word wrapping completely_. That makes most if not all rule descriptions unreadable. 

We have a ticket to address this issue, and we plan to do it in the next release:
https://jira.sonarsource.com/browse/SLI-237

synthomat:
Shouldn't have happened. But great that you plan to address this issue sometime! :)

PR #143, opened 2018-05-07
Update sonarlint.org link in readme

juboucher:


janos-ss:
Thanks! Btw, oddly, there were CRLF line endings instead of LF. I fixed it before merging ;-)

PR #145, opened 2018-05-24
Organizations

dbmeneses:


src/main/java/org/sonarlint/intellij/config/global/wizard/OrganizationStep.java
```diff
@@ -60,18 +62,32 @@ public OrganizationStep(WizardModel model) {
...
+            Messages.showErrorDialog("Failed to fetch organization from SonarQube server: " + task.getException().getMessage(), "Connection Failure");
+            continue;
+          } else {
+            Messages.showErrorDialog(String.format("Organization '%s' not found. Please enter the key of an existing organization.", organizationKey), "Organization Not Found");
+            continue;
+          }
         }
       }
-    });
+      break;
```
janos-ss: This may be a matter of taste, I would find it simpler to explicitly `break` when ready. That is, `if (found) break` and reduce the nesting, and then also when `task.organization() != null` then `break`. And the 2 `continue` can be dropped.

PR #148, opened 2018-06-13
SLI-287 - Update plugin description to fit value proposition

n-b-sonar:
HTML paragraphs inspired by other [open-source plugins](https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/b0c8a9f494b30d076f754f5b7712063a68b44384/google-cloud-tools-plugin/resources/META-INF/plugin.xml) (see [repo entry](https://plugins.jetbrains.com/plugin/8079-google-cloud-tools)).

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -19,7 +19,7 @@
     Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
 
 -->
-<idea-plugin url="http://intellij.sonarlint.org/">
+<idea-plugin url="http://www.sonarlint.org">
     <id>org.sonarlint.idea</id>
     <name>SonarLint</name>
     <version>${project.version}</version>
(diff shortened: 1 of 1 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -19,17 +19,26 @@
     Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
 
 -->
-<idea-plugin url="http://intellij.sonarlint.org/">
+<idea-plugin url="http://www.sonarlint.org">
     <id>org.sonarlint.idea</id>
     <name>SonarLint</name>
     <version>${project.version}</version>
     <vendor email="sonarlint@googlegroups.com" url="http://www.sonarsource.com">SonarSource</vendor>
```
n-b-sonar: @dbmeneses : do you know if I can just dump the _email_ attribute here, or is it mandatory to have one ?

dbmeneses: You probably can, I see other plugins without it.

PR #149, opened 2018-06-14
Locally ignore rules

YuleZ:
Feature that allows to locally ignore rules without changing the server rule set.


YuleZ:
oh, I did not know... Maybe you can use my stuff for that. See my new PR https://github.com/SonarSource/sonarlint-intellij/pull/150

PR #150, opened 2018-06-14
Ignore rules locally

YuleZ:
Feature that allows you to ignore rules locally without changing the server's ruleset.

I found some users requested something like this on StackOverflow.

### DIFF
--- a/src/main/java/org/sonarlint/intellij/actions/ExcludeRuleAction.java
+++ b/src/main/java/org/sonarlint/intellij/actions/ExcludeRuleAction.java
@@ -1,0 +1,87 @@
+/*
+ * SonarLint for IntelliJ IDEA
+ * Copyright (C) 2015 SonarSource
+ * sonarlint@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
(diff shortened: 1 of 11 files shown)


dbmeneses:
Hi @YuleZ thanks for the contribution but we implemented a similar feature that allows to configure the ruleset when not connected to SonarQube/SonarCloud. It replicates what had been done for SonarLint Eclipse and uses a new API in sonarlint core to actually prevents those rules from running. It will be included in the upcoming release.

YuleZ:
Hi @dbmeneses, does this mean you can ignore rules from a connected server only locally? In my case I want to ignore rules like "Don't use NOSONAR" only locally. I cannot change the ruleset in my project and I don't want to see some issues (like "Don't use NOSONAR") when committing, but "real" issues.

janos-ss:
Hi @YuleZ, the rule configuration feature we added is used in projects that are *not connected* to a SonarQube server. When you are connected to a SonarQube server, you can configure the rules on the server.

YuleZ:
Thats my problem. I cannot change them. Thats why I want to ignore them locally.

henryju:
If the server side rules don't feed your need, why not simply removing the connection, and use SonarLint as a standalone tool? The connected mode should help everybody to share the same rules/settings. Allowing people to override server side config while still pretending they are in connected mode would be confusing IMO.

PR #158, opened 2018-07-27
Bugfix

dbmeneses:


src/test/java/org/sonarlint/intellij/ui/nodes/IssueNodeTest.java
```diff
@@ -20,53 +20,16 @@
...
 import org.sonarsource.sonarlint.core.client.api.common.analysis.Issue;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 public class IssueNodeTest {
   private IssueNode node;
-  private TreeCellRenderer renderer = mock(TreeCellRenderer.class);
```
janos-ss: Why is it ok to drop these tests?

dbmeneses: It's not great but the methods in `TreeCellRenderer` are now final and we can't verify their invocation anymore.

PR #159, opened 2018-09-24
SLI-313 Improve matching between local files and files in SonarQube/S…

dbmeneses:
…onarCloud

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,3 @@
-version=3.6-SNAPSHOT
+version=4.0-SNAPSHOT
 org.gradle.jvmargs=-XX:MaxPermSize=256M
 sonarlintCoreVersion=4.0.0.1971


src/main/java/org/sonarlint/intellij/config/module/SonarLintModuleSettings.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.intellij.openapi.components.PersistentStateComponent;
+import com.intellij.openapi.components.State;
+import com.intellij.openapi.components.Storage;
+import com.intellij.openapi.module.ModuleComponent;
+import com.intellij.util.xmlb.XmlSerializerUtil;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@State(name = "SonarLintProjectSettings", storages = {@Storage("sonarlint-module.xml")})
```
henryju: I'm a bit confused, should the name be `SonarLintModuleSettings`?

dbmeneses: Yes, it was wrong and I've fixed it already

PR #161, opened 2018-11-05
SLI-315 - Add Kotlin & Ruby to supported languages

n-b-sonar:


n-b-sonar:
I think the alphabetical order was mostly a coincidence so far. Our intent is more to give a notion of dominance/relevance, even if there is no strong logic (see for example [Atom](https://www.sonarlint.org/atom/)), and then have ordering in Marketplaces correspond to order in sonarlint.org IDE pages ([work in progress](https://github.com/SonarSource/www.sonarlint.org/pull/65) for IntelliJ).

PR #165, opened 2018-12-12
Expand All button in Report Panel

itanoss:
Sometimes we need to expand all even though the number of issues is more than 30.

image


src/main/java/org/sonarlint/intellij/ui/SonarLintAnalysisResultsPanel.java
```diff
@@ -56,14 +57,15 @@ public SonarLintAnalysisResultsPanel(Project project, ProjectBindingManager proj
     subscribeToEvents();
   }
 
-  private static SimpleActionGroup createActionGroup() {
+  private SimpleActionGroup createActionGroup() {
```
janos-ss: Why drop the `static` keyword here?

janos-ss:
Thank you for this suggestion. I'm not sure this is super useful. How exactly is this useful to you? (CC @christophelevis)

henryju:
Thanks you for this suggestion, but it seems to not have received much interest so far. I'm closing this PR. If you (or other) have something to add, like explaining the value of this addition, feel free to comment here, or on our community forum.

PR #179, opened 2019-10-02
SLI-338 Update to latest SLCORE 4.6, add ide_version to telemetry

jblievremont:


src/main/java/org/sonarlint/intellij/config/global/rules/RulesTreeNode.java
```diff
@@ -118,7 +118,7 @@ public String type() {
     }
 
     public String language() {
```
henryju: Maybe rename the method `languageKey` to make it more explicit?

PR #181, opened 2019-10-08
SLI-342 Mark overloaded constructors as @Deprecated to avoid exception in 2019.3

jblievremont:


src/main/java/org/sonarlint/intellij/analysis/LocalFileExclusions.java
```diff
@@ -62,6 +62,10 @@ public LocalFileExclusions(Project project, SonarLintGlobalSettings settings, So
     this(project, settings, projectSettings, appUtils, projectRootManager, PowerSaveMode::isEnabled);
   }
 
+  /**
+   * @deprecated in 4.2 to silence a check in 2019.3
```
henryju: I would add a TODO to switch to NotInjectable when upgrading IJ dependency to 2019.3

PR #183, opened 2019-10-14
SLI-337 Keep issues, log only when there is no parse error

jblievremont:
Reverts most of the changes initially pushed, just avoid the "0 issues" log when all files have parse errors (e.g single "Current file" analysis).


src/main/java/org/sonarlint/intellij/issue/IssueProcessor.java
```diff
@@ -73,7 +73,9 @@ public void process(final SonarLintJob job, ProgressIndicator indicator, final C
     });
 
     Set<VirtualFile> failedVirtualFiles = asVirtualFiles(failedAnalysisFiles);
-    if (! failedVirtualFiles.containsAll(job.allFiles())) {
+    if (failedVirtualFiles.containsAll(job.allFiles())) {
+      console.debug("");
```
henryju: Is it needed to log an empty string?

jblievremont: No, sorry I pushed a bit too fast.

PR #184, opened 2019-10-15
SLI-346 Adapt to dynamic languages in 2019.3 EAP

jblievremont:


src/main/java/org/sonarlint/intellij/SonarApplication.java
```diff
@@ -26,31 +26,47 @@
...
+import java.util.concurrent.ConcurrentMap;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.jetbrains.annotations.NotNull;
 import org.sonarlint.intellij.core.SonarLintProjectNotifications;
 import org.sonarlint.intellij.core.SonarQubeEventNotifications;
 import org.sonarlint.intellij.editor.SonarExternalAnnotator;
 import org.sonarsource.sonarlint.core.client.api.util.FileUtils;
 
+import static org.sonarsource.sonarlint.shaded.com.google.common.base.Predicates.not;
```
henryju: Not sure this is a good thing to rely on this import...

src/main/java/org/sonarlint/intellij/SonarApplication.java
```diff
@@ -26,31 +26,47 @@
...
+
 public class SonarApplication extends ApplicationComponent.Adapter {
   private IdeaPluginDescriptor plugin;
+  private ConcurrentMap<Language, LanguageExtensionPoint> annotatorsByLanguage;
```
henryju: Not sure using Language instance as key is a good thing. The class doesn't implement hashCode. If you look at IJ code, they are either using the class, or the ID as key. If the same language is unloaded/reloaded, the instance may be different, even if this is the same language.

henryju: In fact, maybe in this case we want to unregister/register again the annotator, for the new Language instance.

src/main/java/org/sonarlint/intellij/SonarApplication.java
jblievremont: Yup, I had the same impression while doing my own review.

src/main/java/org/sonarlint/intellij/SonarApplication.java
```diff
@@ -99,7 +131,7 @@ public String getComponentName() {
 
   private static void cleanOldWorkDir() {
     Path oldWorkDir = Paths.get(PathManager.getConfigPath()).resolve("sonarlint").resolve("work");
-    if (!Files.isDirectory(oldWorkDir)) {
+    if (!oldWorkDir.toFile().isDirectory()) {
```
henryju: You did that because of SQ issue right? The performance issue only affect JDK 8 (and maybe not the latest versions). And performance is an issue only when calling this a lot of times. I would keep the previous syntax, that is more future proof to me.

PR #187, opened 2019-11-19
Hardening

henryju:


src/main/java/org/sonarlint/intellij/tasks/ServerUpdateTask.java
```diff
@@ -154,7 +154,7 @@ private void updateProjects(ServerConfiguration serverConfiguration, TaskProgres
     for (Map.Entry<String, List<Project>> entry : projectsPerProjectKey.entrySet()) {
       try {
         updateProject(serverConfiguration, entry.getKey(), entry.getValue(), monitor);
-      } catch (Exception e) {
+      } catch (Throwable e) {
```
henryju: When project is disposed, IntelliJ is raising an AssertionError, which is not an exception... I added a check, but there is still a chance of race condition, so I prefer to catch all here.

src/main/resources/images/type/bug-ui.svg
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
```
jblievremont: Inkscape added a lot of useless stuff, I suggest to give a quick round of [SVGOMG](https://jakearchibald.github.io/svgomg/) to reduce file sizes and/or to exclude those files from the generated JAR if they are not used.

henryju: Good catch, I will remove the SVG

PR #197, opened 2020-01-07
SLI-363 Bump embedded SonarJava to 6.0

jblievremont:


henryju:
Be careful of bug https://jira.sonarsource.com/browse/SONARJAVA-3268. Maybe we should wait for 6.0.1?

jblievremont:
Yup, this PR is more an experiment on my side that I was not planning to merge for SLI 4.4. We'll see which version to embed when we tackle 4.5.

jblievremont:
Updated to 6.1 in #200

PR #201, opened 2020-02-18
SLI-367 Resolve CertificateManager service/component with factory method

jblievremont:


src/main/java/org/sonarlint/intellij/telemetry/TelemetryManagerProvider.java
```diff
@@ -58,7 +58,7 @@ public TelemetryManager get() {
   }
 
   private static TelemetryClientConfig getTelemetryClientConfig() {
-    CertificateManager certificateManager = SonarLintUtils.get(CertificateManager.class);
+    CertificateManager certificateManager = CertificateManager.getInstance();
```
jblievremont: I'm not a big fan of static factory methods, but here:
- it passes tests on 201(7,8,9).x
- it runs smoothly on both 2019.3 and 2020.1 EAP

PR #202, opened 2020-02-18
SLI-374 Force icon size in rule description pane

jblievremont:


jblievremont:
I tested on Ubuntu on a 3K@280dpi screen and on a Retina iMac, works well :+1:

PR #206, opened 2020-02-20
Project closing issues

henryju:


src/main/java/org/sonarlint/intellij/util/SonarLintAppUtils.java
```diff
@@ -35,11 +36,26 @@
...
+      if (!project.isOpen()) {
+        return null;
+      }
+      return ProjectFileIndex.SERVICE.getInstance(project).getModuleForFile(file, false);
+    });
+  }
+
+  private static <G> G runReadAction(Computable<G> computable) {
+    Application application = getApplication();
+    if (application == null || application.isReadAccessAllowed()) {
```
jblievremont: What is the purpose of `application.isReadAccessAllowed` here? IIRC it gets called by `application.runReadAction` anyway.

henryju: I found this pattern in multiple places in the code, so I though it was a good idea. But your're right, this is useless.

PR #212, opened 2020-03-20
SLI-263 Enable TS analysis

jblievremont:


.gitignore
```diff
@@ -1,4 +1,5 @@
 src/main/resources/plugins/*.jar
+src/main/resources/typescript/*.zip
```
henryju: Is it still needed?

src/main/java/org/sonarlint/intellij/core/AnalysisRequirementNotifications.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private AnalysisRequirementNotifications() {
+    // NOP
+  }
+
+  public static final String GROUP_ANALYSIS_PROBLEM = "SonarLint: Unmet Analysis Requirement";
+
+  public static void notifyNodeCommandException(Project project) {
+    Notification notification = new Notification(GROUP_ANALYSIS_PROBLEM,
+      "<b>SonarLint - Unmet Analysis Requirement</b>",
```
henryju: I doesn't like the title too much (at least it doesn't appear to look like something I have already seen). In SLE we simply says as title: `SonarLint - Node.js required`

henryju:
Don't forget the changelog as well ;)

PR #218, opened 2020-04-10
Rework Java analyzer classpath computation

henryju:


src/main/java/org/sonarlint/intellij/analysis/JavaAnalysisConfigurator.java
```diff
@@ -59,92 +68,158 @@
...
   private static final char SEPARATOR = ',';
 
-  private static final String JAR_REGEXP = "(.*)!/";
-  private static final Pattern JAR_PATTERN = Pattern.compile(JAR_REGEXP);
-
   @Override
   public Map<String, String> configure(@NotNull Module ijModule) {
+    JavaModuleClasspath moduleClasspath = new JavaModuleClasspath();
+    moduleClasspath.dependentModules().add(ijModule);
+    collectModuleClasspath(moduleClasspath, ijModule, true, false);
```
damien-urruty-sonarsource, line 75: I think we could simplify those boolean flags. 

In collectModuleClasspath if topLevel==false we do nothing inside the loop, so in this case all we do in the method is call processCompilerOutput. So instead of calling collectModuleClasspath with topLevel=false, we could directly call processCompilerOutput (probably a null check should also be moved). We can now get rid of topLevel that is now always true.

The second flag can then also be simplified in collectModuleClasspath because it is now always false.

In the end this would simplify collectModuleClasspath.

src/main/java/org/sonarlint/intellij/analysis/JavaAnalysisConfigurator.java
```diff
@@ -59,92 +68,158 @@
...
+          addProductionClasspathEntry(moduleClasspath, libPath);
+        }
+      });
+  }
+
+  private static boolean isOnlyForTestClasspath(DependencyScope scope) {
+    return !scope.isForProductionRuntime() && !scope.isForProductionCompile();
+  }
+
+  private static void processCompilerOutput(JavaModuleClasspath moduleClasspath, @NotNull Module module, boolean topLevel, boolean testModule) {
```
damien-urruty-sonarsource, line 166: I think these flags could also be reworked, it would simplify the nesting of conditionals

src/main/java/org/sonarlint/intellij/analysis/JavaAnalysisConfigurator.java
henryju: > if topLevel==false we do nothing inside the loop

I double checked, and I disagree. We do nothing only if topLevel == false AND entry is not exported.

damien-urruty-sonarsource: OK, got it... I wanted to try to simplify things a bit, but it seems we can't

PR #219, opened 2020-04-14
SLI-391 Fix compatibility with older versions of TypeScript analyzer

jblievremont:


src/main/java/org/sonarlint/intellij/core/SonarLintEngineFactory.java
```diff
@@ -170,6 +170,6 @@ private static Path getWorkDir() {
   }
 
   private Map<String, String> prepareExtraProps() {
-    return Collections.singletonMap("sonar.typescript.internal.typescriptLocation", application.getPluginPath().resolve("typescript").resolve("lib").toString());
+    return Collections.singletonMap("sonar.typescript.internal.typescriptLocation", application.getPluginPath().toString());
```
henryju, line 173: @saberduck is it a safe change in the long run? Will the new plugin always support this path?

saberduck: I had a chat with @jblievremont , we simply take this path verbatim and set `NODE_PATH` . Ultimetely it's nodejs resolution algorithm which uses this path. It seems that indeed it should be parent of `typescript` directory

henryju: Thanks!

henryju:
For the record, this is indeed the path to the `node_modules` folder (so parent of the `typescript` directory) that we are using in our medium tests: https://github.com/SonarSource/sonarlint-core/blob/f9f3b683ddcf0f16ff6f656871e1475435fca707/core/src/test/java/org/sonarsource/sonarlint/core/mediumtest/StandaloneIssueMediumTest.java#L99

PR #221, opened 2020-04-21
SLI-392 Fix scroll of rule description panel in settings

jblievremont:


jblievremont:
Agreed. This [minor bug](https://jira.sonarsource.com/browse/SLI-392) was already there in previous versions.

PR #225, opened 2020-04-29
SLI-395 support rule parameters in global settings

kirill-knize-sonarsource:


src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -47,13 +48,52 @@
...
+  @Deprecated
+  private final Set<String> excludedRules = new HashSet<>();
+  private final Map<String, Rule> rules = new HashMap<>();
 
   public static SonarLintGlobalSettings getInstance() {
     return ApplicationManager.getApplication().getComponent(SonarLintGlobalSettings.class);
   }
 
+  public void setRuleParam(String ruleKey, String paramName, String paramValue) {
+    if (!rules.containsKey(ruleKey)) {
```
henryju, line 62: you can use computeIfAbsent here

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -47,13 +48,52 @@
...
+      rules.put(ruleKey, new Rule(true));
+    }
+    rules.get(ruleKey).params.put(paramName, paramValue);
+  }
+
+  public String getRuleParam(String ruleKey, String paramName) {
```
henryju, line 68: Maybe naming the method `getRuleParamValue` or even `getRuleParamUserValue` would be clearer.

henryju: I would return Optional<String>, or at least annotate `@CheckForNull`

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -47,13 +48,52 @@
...
+  public void enableRule(String ruleKey) {
+    setRuleActive(ruleKey, true);
+  }
+
+  public void disableRule(String ruleKey) {
+    setRuleActive(ruleKey, false);
+  }
+
+  private void setRuleActive(String ruleKey, boolean active) {
+    if (!rules.containsKey(ruleKey)) {
```
henryju, line 84: Use `computeIfAbsent`

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -47,13 +48,52 @@
...
+      rules.put(ruleKey, new Rule(false));
+    }
+    rules.get(ruleKey).isActive = active;
+  }
+
+  public boolean isRuleActive(String ruleKey) {
+    if (!rules.containsKey(ruleKey)) {
+      return true;
```
henryju, line 92: Returning `true` is misleading. When a rule has not explicitely be activated or disactivated, it should be the default activation from the analyzer.
I see two options here:
* returns `Optional<Boolean>`, with `Optional.empty()` meaning there is no explicit state
* change the method to `isRuleExplicitlyDisabled` (since it is used in the quick fix to decide if we should offer to disable the rule)

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -62,6 +102,12 @@ public SonarLintGlobalSettings getState() {
   @Override
   public void loadState(SonarLintGlobalSettings state) {
     XmlSerializerUtil.copyBean(state, this);
+    includedRules.forEach(it -> {
+      rules.put(it, new Rule(true));
+    });
+    excludedRules.forEach(it -> {
+      rules.put(it, new Rule(false));
+    });
```
henryju, line 110: If you don't clear (or better set to null) the two sets after the migration, it means the next time you'll reopen your IDE, the migration will happen again, and override possible different values added in the meantime by the user in the rules map.

src/test/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettingsTest.java
```diff
@@ -47,10 +51,37 @@ public void testRoundTrip() {
...
+  public void testRuleParamAccessors() {
+    SonarLintGlobalSettings settings = new SonarLintGlobalSettings();
+    settings.setRuleParam(RULE, PARAM, VALUE);
+    assertThat(settings.getRuleParam(RULE, PARAM)).isEqualTo(VALUE);
+  }
+
+  @Test
+  public void testRuleIsNotLoaded() {
+    SonarLintGlobalSettings settings = new SonarLintGlobalSettings();
+    assertThat(settings.getRuleParam(RULE, PARAM)).isEqualTo(null);
```
henryju, line 70: s/`isEqualTo(null)`/`isNull()`

src/test/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettingsTest.java
```diff
@@ -47,10 +51,37 @@ public void testRoundTrip() {
...
+  }
 
+  @Test
+  public void testDisableRule() {
+    SonarLintGlobalSettings settings = new SonarLintGlobalSettings();
+    settings.disableRule(RULE);
+    assertThat(settings.isRuleActive(RULE)).isEqualTo(false);
```
henryju, line 77: s/`isEqualTo(false)`/`isFalse`

src/test/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettingsTest.java
```diff
@@ -47,10 +51,37 @@ public void testRoundTrip() {
...
+  }
+
+  @Test
+  public void testEnableRule() {
+    SonarLintGlobalSettings settings = new SonarLintGlobalSettings();
+    settings.enableRule(RULE);
+    assertThat(settings.isRuleActive(RULE)).isEqualTo(true);
```
henryju, line 84: `isTrue()`

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
jblievremont: I believe we need a `resetRuleParam(String ruleKey, String paramName)` method, to remove the explicitly set value in case a rule is "reset" to its default behavior.

src/test/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettingsTest.java
```diff
@@ -47,10 +55,73 @@ public void testRoundTrip() {
...
-    super.register(app, SonarLintGlobalSettings.class, instance);
-    assertThat(SonarLintGlobalSettings.getInstance()).isEqualTo(instance);
+    SonarLintGlobalSettings settings = new SonarLintGlobalSettings();
+    super.register(app, SonarLintGlobalSettings.class, settings);
+    assertThat(SonarLintGlobalSettings.getInstance()).isEqualTo(settings);
+
+  }
+
+  @Test
+  public void testLoadState() throws Exception {
```
jblievremont, line 65: Nitpick: this test only checks that the old format is correctly handled, I think it would be great to add another one to check that the new format is correctly loaded/saved.

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -62,6 +105,18 @@ public SonarLintGlobalSettings getState() {
   @Override
   public void loadState(SonarLintGlobalSettings state) {
     XmlSerializerUtil.copyBean(state, this);
+    if(includedRules != null) {
```
henryju, line 108: Formatting issue. Please ensure you are using the SonarSource configuration, and apply a reformat.

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -47,13 +49,54 @@
   private boolean autoTrigger = true;
   private List<SonarQubeServer> servers = new LinkedList<>();
   private List<String> fileExclusions = new LinkedList<>();
+  @Deprecated
   private Set<String> includedRules = new HashSet<>();
```
henryju, line 53: I wonder if the default value should not be changed to `null`?

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -47,13 +49,54 @@
...
+  @Deprecated
   private Set<String> excludedRules = new HashSet<>();
```
henryju, line 55: Same here (`null` by default ?)

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -84,17 +139,40 @@ public String getComponentName() {
   }
 
   public Set<String> getIncludedRules() {
-    return includedRules;
+    if (includedRules != null) {
+      return includedRules;
```
henryju, line 143: Should never occurs, or it means the migration failed?

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -84,17 +139,40 @@ public String getComponentName() {
...
+    this.rules = rules;
   }
 
   public void setIncludedRules(Set<String> includedRules) {
     this.includedRules = includedRules;
   }
 
+  public Set<String> getExcludedRules() {
+    if (excludedRules != null) {
+      return excludedRules;
```
henryju, line 167: Same, remove this condition.

src/test/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettingsTest.java
```diff
@@ -47,10 +56,89 @@ public void testRoundTrip() {
...
+
+  }
+
+  @Test
+  public void testLoadStateNewFormat() throws Exception {
+    SonarLintGlobalSettings state = new SonarLintGlobalSettings();
+    SonarLintGlobalSettings settings = new SonarLintGlobalSettings();
+    HashMap<String, SonarLintGlobalSettings.Rule> rules = new HashMap<>();
+    rules.put(RULE, new SonarLintGlobalSettings.Rule(false));
+    Field rulesField = state.getClass().getDeclaredField("rules");
```
henryju, line 98: I suggest to add a package private accessor, to be used by the test, instead of using reflection.

src/test/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettingsTest.java
```diff
@@ -47,10 +56,89 @@ public void testRoundTrip() {
...
+  }
+
+  @Test
+  public void testLoadStateOldFormat() throws Exception {
+    SonarLintGlobalSettings state = new SonarLintGlobalSettings();
+    HashSet<String> includedRules = new HashSet<>();
+    includedRules.add(RULE);
+    HashSet<String> excludedRules = new HashSet<>();
+    excludedRules.add(RULE1);
+    Field includedRulesField = state.getClass().getDeclaredField("includedRules");
```
henryju, line 72: Don't use reflection. We usually relax encapsulation by allowing classes in the same package to access attributes for testing purposes.

jblievremont:
Rebased and merged into sprint branch as @93082392

PR #226, opened 2020-05-05
Initial version of the UI

henryju:


src/main/resources/META-INF/plugin.xml
```diff
@@ -102,7 +102,7 @@
     </change-notes>
 
     <!-- please see http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/build_number_ranges.html for description -->
-    <idea-version since-build="181.3007.14"/>
+    <idea-version since-build="183.2940.10"/>
```
henryju, line 105: This is to allow JBTextArea

PR #228, opened 2020-05-07
Migrate tests to 2020.1

henryju:
* use official test framework in many cases
* stop modifying the global static Application in tests
* stop compiling with multiple IJ versions in QA


src/main/java/org/sonarlint/intellij/util/SonarLintAppUtils.java
```diff
@@ -98,7 +100,7 @@ public String getPathRelativeToProjectBaseDir(Project project, VirtualFile file)
     if (project.getBasePath() == null) {
       return null;
     }
-    String relativePathToProject = VfsUtil.getRelativePath(file, project.getBaseDir());
+    String relativePathToProject = FileUtil.getRelativePath(ProjectUtil.guessProjectDir(project).getPath(), file.getPath(), '/');
```
henryju, line 103: Not sure about this one. Please have a close look if you have time.

src/test/java/org/sonarlint/intellij/issue/IssueManagerTest.java
```diff
@@ -45,13 +46,12 @@
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
-public class IssueManagerTest extends SonarTest {
+public class IssueManagerTest extends LightPlatformCodeInsightFixture4TestCase {
```
jblievremont, line 49: Why use the parent class instead of our `AbstractSonarLintLightTest` here? IMHO, sticking to our own test parent can help with migrations in the future.

henryju: :+1: This is a mistake.

PR #229, opened 2020-05-11
SLI-378 store password and token in PasswordSafe not in xml

kirill-knize-sonarsource:


jblievremont:
Reworked as part of #237

PR #232, opened 2020-05-13
SLI-399 SLI-400 add release notes, change versions of PHP and Python …

kirill-knize-sonarsource:


src/main/resources/META-INF/plugin.xml
```diff
@@ -50,6 +50,7 @@
 
     <change-notes><![CDATA[
       <ul>
+        <li>4.8 - Support of rules parameters configuration for IntelliJ. Fewer false positives and performance improvement for PHP. 4 new vulnerability detection rules for Python.</li>
```
jblievremont, line 53:
```suggestion
        <li>4.8 - Support offline configuration of rule parameters. Performance improvement and fewer false positives for PHP. 4 new vulnerability detection rules for Python.</li>
```

jblievremont: @christophelevis could you please have a look at the suggested release notes for SLI 4.8?

src/main/resources/META-INF/plugin.xml
```diff
@@ -50,6 +50,7 @@
 
     <change-notes><![CDATA[
       <ul>
+        <li>4.8 - Support offline configuration of rule parameters. Performance improvement and fewer false positives for PHP. 4 new vulnerability detection rules for Python.</li>
```
henryju, line 53: Is `offline` the new name for `standalone`? I don't find it very explicit. Maybe `local` instead? Or instead `Support configuration of rules parameters in IDE settings`. No need to emphasis this is not for connected mode IMO.

PR #233, opened 2020-05-14
SLI-387 Cache the dirty state to avoid costly computation

henryju:


src/main/java/org/sonarlint/intellij/config/global/rules/RuleConfigurationPanel.java
```diff
@@ -177,6 +185,7 @@ private void restoreDefaults() {
       r.getCustomParams().clear();
     });
     updateModel();
+    recomputeDirtyState();
```
jblievremont, line 188: Another option could be to store the info that `isDirty` must be recomputed, and lazily recompute it if needed when it is queried (a kind of "dirty dirty").

henryju: In this case the UI will take longer to refresh. Now `isModified` is always instantaneous.

PR #234, opened 2020-05-14
SLI-387 Ensure migrated rule activations are persisted on first load

jblievremont:


henryju:
@jblievremont I just pushed our common refactoring. I let you handle the merge once the pipeline is green.

PR #237, opened 2020-05-19
SLI-378 Use the IDE-provided password safe to store sensitive information

jblievremont:


jblievremont:
Stale branch, will need more work especially on cross platform validation.

PR #240, opened 2020-06-11
SLI-368 Declare the AnalysisConfiguration EP dynamic

henryju:


henryju:
I haven't find any other way.

PR #241, opened 2020-06-15
Remove null check in MakeTrigger

henryju:
CompileContext::getProject was changed to @NotNull in
https://github.com/JetBrains/intellij-community/commit/07e8f9dacd61811216a109e89e1e20fd89f126f5#diff-1442c013914b9019d787cfcaef28615c


damien-urruty-sonarsource:
Shouldn't we keep it for backward compatibility ? Otherwise LGTM

henryju:
The change was introduced in IDEA 2019.1 so this is above the min version we support.

PR #249, opened 2020-06-26
SLI-368 Fix plugin uninstall

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/core/SonarLintEngineManager.java
```diff
@@ -141,5 +143,6 @@ public void dispose() {
       standalone.stop();
       standalone = null;
     }
+    Loggers.setTarget(null);
```
henryju, line 146: I don't know why, but I don't like this change. We can see in the next sprint if we can do better.

damien-urruty-sonarsource: :+1: sure we can do better

PR #252, opened 2020-07-09
SLI-415 Fix duplicated annotations for PHP when language is registered.

damien-urruty-sonarsource:


jblievremont:
Nice!

Tested in PhpStorm 2020.1 and 2019.1, and IDEA Ultimate + PHP plugin 2020.1 and 2019.3 - all good! :+1:

PR #253, opened 2020-07-10
SLI 412 Display severity from quality profile in rule panel

damien-urruty-sonarsource:


damien-urruty-sonarsource:
@henryju I think it would be simpler to make the review commit by commit in chronological order. I would also like you to play with it because I may have missed a corner case
FYI @jblievremont & @kirill-knize-sonarsource

src/main/java/org/sonarlint/intellij/core/ConnectedSonarLintFacade.java
```diff
@@ -89,8 +89,21 @@ protected AnalysisResults analyze(Path baseDir, Path workDir, Collection<ClientI
...
+  public String getDescription(String ruleKey) {
+    ConnectedRuleDetails details = ruleDetails(ruleKey);
+    if (details == null) {
+      return null;
+    }
+    String extendedDescription = details.getExtendedDescription();
+    if (extendedDescription.isEmpty()) {
+      return details.getHtmlDescription();
+    }
+    return details.getHtmlDescription() + "<br/><br/>" + details.getExtendedDescription();
```
henryju, line 106:
```suggestion
    return details.getHtmlDescription() + "<br/><br/>" + extendedDescription;
```

damien-urruty-sonarsource: :+1:

PR #258, opened 2020-07-29
SLI-300 Highlight first flow when selecting an issue

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/editor/SonarLintHighlighting.java
```diff
@@ -73,47 +68,46 @@ private void stopBlinking() {
...
-    HighlightInfo primaryInfo = createHighlight(rangeMarker, message);
-
-    List<HighlightInfo> infos = flows.stream()
-      .flatMap(f -> f.locations().stream()
+  public void highlightIssue(LiveIssue issue) {
+    RangeMarker issueRange = issue.getRange();
+    if (issueRange == null) {
+      return;
+    }
+    List<HighlightInfo> highlights = issue.flows().stream().findFirst()
```
henryju: Remember the special case of issues having only flows with a single location (like cognitive complexity). Here all (single) locations of all flows should be highlighted.

src/main/java/org/sonarlint/intellij/issue/LiveIssue.java
```diff
@@ -208,6 +209,19 @@ public void setType(@Nullable String type) {
     return flows;
   }
 
+  public String getFlowsDescription() {
```
henryju, line 212: Maybe this could be cached to avoid recomputing it every time the annotator runs?

src/main/java/org/sonarlint/intellij/editor/SonarLintHighlighting.java
```diff
@@ -73,55 +68,65 @@ private void stopBlinking() {
...
+    updateHighlights(createFlowHighlights(flow), flow.locations().get(0).location().getDocument());
+  }
 
-    infos.add(primaryInfo);
+  public void highlightIssue(LiveIssue issue) {
+    RangeMarker issueRange = issue.getRange();
+    if (issueRange == null) {
+      return;
+    }
+    List<HighlightInfo> highlights = issue.flows().stream().findFirst().map(SonarLintHighlighting::createFlowHighlights).orElse(new ArrayList<>());
```
henryju, line 84: Only `findFirst` when there are more than one location per flows.

PR #261, opened 2020-08-05
SLI-428 Display secondary location indexes in editor

damien-urruty-sonarsource:


damien-urruty-sonarsource:
Hey guys, could you also test it please ? Thanks!

damien-urruty-sonarsource:
Thanks a lot @jblievremont for fixing the rounded corners and for improving the dark theme!

I had seen the antialising option but hadn't tested it. Actually there is something in the IntelliJ API about this: `com.intellij.util.ui.GraphicsUtil.setupAntialiasing`. I don't know if it does the same thing though, but maybe this could simplify a bit ?

jblievremont:
Good catch! Actually, `setupAntialiasing` only enables font AA. However, I made good use of `com.intellij.util.ui.GraphicsUtil#setupRoundedBorderAntialiasing` which does what it says on the tin :smile:

PR #262, opened 2020-08-06
SLI-433 Relax search criteria in rule configuration filter

jblievremont:
I made sure that:
- search by rule key finds the exact rule (to avoid breaking the link in the rule description)
- all terms from the search criterion match (AND search)
- sub-words are matched (e.g "null" will match "nullpointerexception", "notnull", "nullable")


damien-urruty-sonarsource:
Tested :+1:

PR #271, opened 2020-09-01
SLI-388 Protect against duplicate file submission in case of problematic module setup

jblievremont:


damien-urruty-sonarsource (approved):
Tested on IntelliJ CE 2020.2 :+1:

PR #272, opened 2020-09-02
SLI-437 Project exclusion filters settings are never loaded after restart

damien-urruty-sonarsource:


damien-urruty-sonarsource:
@jblievremont could you have a look please and also test it ?

Classes that manage persistence were both services and state holders. At first I did split them and had the impression that it fixed the bug. It was not the case, the real fix is just a one-liner but I decided to keep separate classes for services and states as it is cleaner IMO. I also simplified the way we access them.

damien-urruty-sonarsource:
FYI @henryju @kirill-knize-sonarsource, the last commit which contains the real fix shows what should be avoided in state classes for persistence

src/main/java/org/sonarlint/intellij/messages/GlobalConfigurationListener.java
```diff
@@ -21,6 +21,7 @@
 
 import com.intellij.util.messages.Topic;
 import java.util.List;
+
```
jblievremont, line 24: Nitpick: unneeded change

src/main/java/org/sonarlint/intellij/config/Settings.java
```diff
@@ -0,0 +1,45 @@
...
+import org.sonarlint.intellij.config.global.SonarLintGlobalSettings;
+import org.sonarlint.intellij.config.global.SonarLintGlobalSettingsStore;
+import org.sonarlint.intellij.config.module.SonarLintModuleSettings;
+import org.sonarlint.intellij.config.module.SonarLintModuleSettingsStore;
+import org.sonarlint.intellij.config.project.SonarLintProjectSettings;
+import org.sonarlint.intellij.config.project.SonarLintProjectSettingsStore;
+
+import static org.sonarlint.intellij.util.SonarLintUtils.getService;
+
+public class Settings {
```
jblievremont, line 33: Since this class only exposes static stuff, I think it should have a private "no-op" constructor.

src/main/java/org/sonarlint/intellij/config/Settings.java
```diff
@@ -0,0 +1,45 @@
...
+  public static SonarLintModuleSettings getSettingsFor(Module module) {
+    return getService(module, SonarLintModuleSettingsStore.class).getState();
+  }
+
+  public static SonarLintProjectSettings getSettingsFor(Project project) {
```
jblievremont, line 38: :+1:

src/main/java/org/sonarlint/intellij/config/Settings.java
```diff
@@ -0,0 +1,45 @@
...
+    return getService(project, SonarLintProjectSettingsStore.class).getState();
+  }
+
+  public static SonarLintGlobalSettings getGlobalSettings() {
```
jblievremont, line 42: :+1:

jblievremont (approved):
LGTM :+1: tested in IC 2020.1 and IU 2020.2!

src/main/java/org/sonarlint/intellij/messages/GlobalConfigurationListener.java
damien-urruty-sonarsource: :+1:

src/main/java/org/sonarlint/intellij/config/Settings.java
damien-urruty-sonarsource: I know that this is a common convention among SonarSourcers, it is also raised as a major code smell by next. I usually don't like this convention, for 3 reasons:
* adding a private constructor results in dead code, it is inserted "just in case", so it should not be here. As an exception maybe I can see the point of inserting a private constructor when it's a public API, here it's just internal stuff
* I usually expect people to understand that it's a utility class when all the methods are static. At least personally I don't feel the need to have a private constructor with a comment to explain it
* anyway if you misunderstand and instantiate the class, you can't do anything with it. You basically have a raw object

It could be an opportunity to talk about it on Discuss! 
That said I guess I need to apply the convention anyway

jblievremont: This might be an argument in favor of a switch to e.g Kotlin: first class functions, instead of utility class methods.

PR #274, opened 2020-09-04
SLI-420 Use latest blocker icon from sonar-ui-common

jblievremont:


damien-urruty-sonarsource (approved):
Tested on IntelliJ 2020.2 CE :+1:

PR #275, opened 2020-09-08
SLI-416 Show notification link on same line as message

jblievremont:


damien-urruty-sonarsource (approved):
Tested on CE 2020.2 :+1:

PR #278, opened 2020-09-22
SLI-451 Simplify test file detection

henryju:
Based on a discussion on the JB slack channel.


henryju:
I have tested a Java/Maven project in IDEA. If you have the opportunity to test with other languages/other flavors, please do :)

jblievremont:
Will test with PHP.

On a broader note: this is going to be fun with languages like Rust where the test code is in the same _files_ as production code :innocent:

jblievremont (approved):
Tested in PhpStorm 2020.2.1 on code tested with PhpUnit.

Test source file was correctly classified as test, on-the-fly analysis did not raise any issue, but "analyze all files" correctly raised a test-related issue - maybe a side effect of the symbol table.

LGTM! :+1:

PR #279, opened 2020-09-23
Node.js requirement

henryju:
Better support for Node.js and JRE requirements

### DIFF
--- a/src/main/java/org/sonarlint/intellij/core/AnalysisRequirementNotifications.java
+++ b/src/main/java/org/sonarlint/intellij/core/AnalysisRequirementNotifications.java
@@ -54,6 +54,10 @@ public class AnalysisRequirementNotifications {
     // NOP
   }
 
+  public static void resetCachedMessages() {
+    alreadyNotified.clear();
+  }
+
   public static void notifyOnceForSkippedPlugins(AnalysisResults analysisResults, Collection<PluginDetails> allPlugins, Project project) {
     Set<Language> attemptedLanguages = new HashSet<>(analysisResults.languagePerFile().values());
     attemptedLanguages.forEach(l -> {
--- a/src/main/java/org/sonarlint/intellij/core/AnalysisRequirementNotifications.java
+++ b/src/main/java/org/sonarlint/intellij/core/AnalysisRequirementNotifications.java
@@ -24,25 +24,96 @@ import com.intellij.notification.Notification;
 import com.intellij.notification.NotificationGroup;
 import com.intellij.notification.NotificationListener;
 import com.intellij.notification.NotificationType;
+import com.intellij.openapi.options.ShowSettingsUtil;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.wm.ToolWindow;
 import com.intellij.openapi.wm.ToolWindowManager;
 import com.intellij.ui.content.Content;
 import com.intellij.ui.content.ContentManager;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import javax.annotation.Nullable;
 import javax.swing.event.HyperlinkEvent;
 import org.jetbrains.annotations.NotNull;
+import org.sonarlint.intellij.config.global.SonarLintGlobalConfigurable;
 import org.sonarlint.intellij.ui.SonarLintToolWindowFactory;
+import org.sonarsource.sonarlint.core.client.api.common.Language;
+import org.sonarsource.sonarlint.core.client.api.common.PluginDetails;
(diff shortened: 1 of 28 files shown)


src/main/java/org/sonarlint/intellij/core/AnalysisRequirementNotifications.java
```diff
@@ -24,25 +24,96 @@
...
+import org.sonarsource.sonarlint.core.client.api.common.PluginDetails;
+import org.sonarsource.sonarlint.core.client.api.common.SkipReason;
+import org.sonarsource.sonarlint.core.client.api.common.analysis.AnalysisResults;
 
 public class AnalysisRequirementNotifications {
 
-  private static final NotificationGroup ANALYSIS_PROBLEM_GROUP = NotificationGroup.balloonGroup("SonarLint: Analysis Requirement");
+  private static final NotificationGroup ANALYZER_REQUIREMENT_GROUP = NotificationGroup.balloonGroup("SonarLint: Analyzer Requirement");
+
+  private static final Set<String> alreadyNotified = new HashSet<>();
```
jblievremont, line 51: This means that users will have to reload SonarLint / restart the IDE to have the checks performed again, right?

Let's imagine someone has the warning about Node, installs it while the IDE is running (e.g with a package manager that puts it on the `$PATH`); then technically the requirement is met, but SonarLint will only notice it on restart.

I believe we have to either:
- mention that the IDE must be restarted in the message about Node
- make sure to clean this cache when the path to Node is updated in the settings

src/main/java/org/sonarlint/intellij/core/AnalysisRequirementNotifications.java
```diff
@@ -24,25 +24,96 @@
...
+              runtimeRequirement.getMinVersion(), l.getLabel(), runtimeRequirement.getCurrentVersion());
+            createNotificationOnce(project, title, content, new NotificationListener.UrlOpeningListener(true));
+          } else if (runtimeRequirement.getRuntime() == SkipReason.UnsatisfiedRuntimeRequirement.RuntimeRequirement.NODEJS) {
+            StringBuilder content = new StringBuilder(String.format("SonarLint requires Node.js runtime version %s or later to analyze %s code.", runtimeRequirement.getMinVersion(), l.getLabel()));
+            if (runtimeRequirement.getCurrentVersion() != null) {
+              content.append(String.format(" Current version is %s.", runtimeRequirement.getCurrentVersion()));
+            }
+            content.append("<br>Please configure the Node.js path in the <a href=\"#edit-settings\">SonarLint settings</a>.");
```
jblievremont, line 74: I guess the new settings will come in another PR?

src/main/java/org/sonarlint/intellij/core/ProjectBindingManager.java
```diff
@@ -68,14 +68,14 @@ public synchronized SonarLintFacade getFacade(boolean logDetails) throws Invalid
...
     SonarLintConsole console = SonarLintUtils.getService(myProject, SonarLintConsole.class);
     if (projectSettings.isBindingEnabled()) {
-      String serverId = projectSettings.getServerId();
+      String connectionId = projectSettings.getServerId();
       String projectKey = projectSettings.getProjectKey();
-      checkBindingStatus(notifications, serverId, projectKey);
+      checkBindingStatus(notifications, connectionId, projectKey);
       if (logDetails) {
-        console.info(String.format("Using configuration of '%s' in server '%s'", projectKey, serverId));
+        console.info(String.format("Using connection '%s' for project '%s'", connectionId, projectKey));
```
jblievremont, line 75: :+1:

jblievremont (approved):
Looks good on the functional side, see my comments about the "cache" of checks + settings declaration.

src/main/java/org/sonarlint/intellij/core/AnalysisRequirementNotifications.java
henryju: Cache cleaned in the next commit, when the setting is updated

src/main/java/org/sonarlint/intellij/core/AnalysisRequirementNotifications.java
henryju: in fact I added it to the same PR, sorry if it makes the review more difficult

henryju:
@jblievremont there is the latest changes added in the last commits. To ease testing you can update the embedded SonarJS plugin to 6.5.0.13345 (thanks to https://github.com/SonarSource/SonarJS/pull/2172)

jblievremont (approved):
LGTM :+1:

PR #286, opened 2020-10-20
SLI-457 Endpoint for getting IDE name and version

kirill-knize-sonarsource:


src/main/java/org/sonarlint/intellij/core/ProjectBindingManager.java
```diff
@@ -27,6 +27,7 @@
 import org.sonarlint.intellij.config.global.SonarQubeServer;
 import org.sonarlint.intellij.config.project.SonarLintProjectSettings;
 import org.sonarlint.intellij.exception.InvalidBindingException;
+import org.sonarlint.intellij.server.SonarLintHttpServer;
```
damien-urruty-sonarsource, line 30: I suppose this import can be removed

src/main/java/org/sonarlint/intellij/util/SonarLintUtils.java
```diff
@@ -297,6 +301,30 @@ public static boolean isPhpFile(@NotNull PsiFile file) {
...
+      return false;
+    } catch (Exception e) {
+      if (e.getMessage().contains("refused")) {
+        return true;
+      }
+      LOG.error("Troubles checking if port is open", e);
+      throw new PortAvailabilityCheckException("Troubles checking if port is open: " + portNumber, e);
+    }
+  }
+
```
damien-urruty-sonarsource, line 327: I think this method is not used anymore

src/main/java/org/sonarlint/intellij/util/SonarLintUtils.java
```diff
@@ -297,6 +301,30 @@ public static boolean isPhpFile(@NotNull PsiFile file) {
     return "php".equalsIgnoreCase(file.getFileType().getName());
   }
 
+  public static File getIdeIconForOpenInIde() {
+    try {
+      return ((ApplicationInfoImpl) getAppInfo()).getApplicationSvgIconFile();
+    } catch (NullPointerException e) {
+      return null;
```
damien-urruty-sonarsource, line 308: Do we really need to catch this NPE ? Can it happen ? If yes, couldn't we check if something is null instead of dereferencing the null reference ?

src/main/java/org/sonarlint/intellij/exception/PortAvailabilityCheckException.java
```diff
@@ -0,0 +1,27 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package org.sonarlint.intellij.exception;
+
+public class PortAvailabilityCheckException extends RuntimeException {
```
damien-urruty-sonarsource, line 22: I think this class is not used

src/main/java/org/sonarlint/intellij/server/SonarLintHttpServer.kt
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    fun tryToStart(numberOfAttempts: Int) {
+        try {
+            actualStart(STARTING_PORT + numberOfAttempts)
+        } catch (e: BindException) {
+            if (numberOfAttempts < PORT_RANGE) {
+                tryToStart(numberOfAttempts + 1)
+            } else {
+                throw RuntimeException("Couldn't start SonarLint server in port range: $STARTING_PORT - ${STARTING_PORT + PORT_RANGE}")
```
damien-urruty-sonarsource, line 75: Is this exception caught somewhere ? How would we show the user that it failed ?

src/main/java/org/sonarlint/intellij/server/SonarLintHttpServer.kt
```diff
@@ -0,0 +1,182 @@
...
+            val bossGroup: EventLoopGroup = NioEventLoopGroup(1)
+            val workerGroup: EventLoopGroup = NioEventLoopGroup()
+            try {
+                val b = ServerBootstrap()
+                b.group(bossGroup, workerGroup)
+                        .channel(NioServerSocketChannel::class.java)
+                        .handler(LoggingHandler(LogLevel.INFO))
+                        .childHandler(HttpSnoopServerInitializer())
+                val ch = b.bind(port).sync().channel()
+                System.err.println("Open your web browser and navigate to http://localhost:$port/")
```
damien-urruty-sonarsource, line 92: Some traces like this one can be removed I guess

src/main/java/org/sonarlint/intellij/util/SonarLintUtils.java
```diff
@@ -297,6 +301,30 @@ public static boolean isPhpFile(@NotNull PsiFile file) {
     return "php".equalsIgnoreCase(file.getFileType().getName());
   }
 
+  public static File getIdeIconForOpenInIde() {
```
damien-urruty-sonarsource, line 304: This method is never used

src/main/java/org/sonarlint/intellij/server/SonarLintHttpServer.kt
```diff
@@ -0,0 +1,182 @@
...
+
+    private val buf = StringBuilder()
+    override fun channelReadComplete(ctx: ChannelHandlerContext) {
+        ctx.flush()
+    }
+
+
+    private fun processRequest(request: HttpRequest): String {
+        if (request.uri() == ENVIRONMENT_ENDPOINT && request.method() == HttpMethod.GET) {
+            return SonarLintUtils.getIdeVersionForTelemetry() ?: UNKNOWN_INTELLIJ_FLAVOR
```
damien-urruty-sonarsource, line 128: The handler for the request and the response that we return are hidden in the middle of this big server class, mixed with details about the server setup. Could we find a way to extract this somewhere else ?

src/main/java/org/sonarlint/intellij/server/SonarLintHttpServer.kt
```diff
@@ -0,0 +1,175 @@
...
+        ctx.flush()
+    }
+
+    public override fun channelRead0(ctx: ChannelHandlerContext, msg: Any?) {
+        if (msg is HttpRequest) {
+            request = msg
+            val request = request as HttpRequest
+            if (HttpUtil.is100ContinueExpected(request)) {
+                send100Continue(ctx)
+            }
```
damien-urruty-sonarsource, line 134: I am not sure we need to handle such cases, have you tested without that ?

src/main/java/org/sonarlint/intellij/core/SonarLintEngineManager.java
```diff
@@ -42,6 +44,13 @@
 
   public SonarLintEngineManager() {
     this(new SonarLintEngineFactory());
+    if (!ApplicationManager.getApplication().isUnitTestMode()) {
+      startSonarLintWebServer();
+    }
```
damien-urruty-sonarsource, line 49: Could we avoid using this if ? Maybe this is the sign that the start should not happen here, but in a different place that we would not reach from unit tests ?

src/main/java/org/sonarlint/intellij/server/SonarLintHttpServer.kt
```diff
@@ -0,0 +1,175 @@
...
+            response.headers().setInt(HttpHeaderNames.CONTENT_LENGTH, response.content().readableBytes())
+            response.headers()[HttpHeaderNames.CONNECTION] = HttpHeaderValues.KEEP_ALIVE
+        }
+        ctx.write(response)
+        return keepAlive
+    }
+
+    override fun exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable) {
+        // TODO change to LOGGER
+        cause.printStackTrace()
```
damien-urruty-sonarsource, line 163: You can maybe try to use the SonarLintConsole service here ?

PR #287, opened 2020-10-23
SLI-458 Fetch and display a security hotspot

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/ui/SonarLintHotspotsListPanel.java
```diff
@@ -0,0 +1,75 @@
...
+  private final DefaultTreeModel treeModel;
+  private final Project project;
+  private HotspotNode hotspotNode;
+
+  public SonarLintHotspotsListPanel(Project project) {
+    this.project = project;
+    treeModel = new DefaultTreeModel(null);
+    hotspotsTree = new Tree(treeModel);
+    hotspotsTree.setCellRenderer(new HotspotCellRenderer());
+    hotspotsTree.getEmptyText().setText("No hotspot to display, open one from SonarQube");
```
henryju, line 45: Typo
```suggestion
    hotspotsTree.getEmptyText().setText("No hotspots to display, open one from SonarQube");
```

src/main/resources/META-INF/plugin.xml
```diff
@@ -118,7 +118,7 @@
 
     <extensions defaultExtensionNs="com.intellij">
         <codeInsight.linkHandler prefix="#sonarissue/" handlerClass="org.sonarlint.intellij.editor.SonarLinkHandler"/>
-        <toolWindow id="SonarLint" anchor="bottom" icon="/images/ico-sonarlint-13.png" factoryClass="org.sonarlint.intellij.ui.SonarLintToolWindowFactory"/>
+        <toolWindow id="SonarLint" anchor="bottom" icon="/images/ico-sonarlint-13.png" factoryClass="org.sonarlint.intellij.ui.SonarLintToolWindowFactory" canCloseContents="true"/>
```
henryju, line 121: Does it mean all tabs could be closed now? Or only hotspots?

src/main/java/org/sonarlint/intellij/issue/hotspot/LocalHotspot.java
```diff
@@ -0,0 +1,76 @@
...
+  public String getRuleKey() {
+    return remote.rule.key;
+  }
+
+  public String getAuthor() {
+    return remote.author;
+  }
+
+  public String getStatusDescription() {
+    return remote.status.description + (remote.resolution == null ? "" : (" as " + remote.resolution.description));
```
kirill-knize-sonarsource, line 52: I'm don't think you need internal parentheses. This should work the same:
return remote.status.description + (remote.resolution == null ? "" : " as " + remote.resolution.description);

src/main/java/org/sonarlint/intellij/ui/HotspotCellRenderer.java
```diff
@@ -0,0 +1,86 @@
...
+    panel.add(probabilityLabel);
+
+    panel.add(new JLabel(hotspot.getMessage()));
+
+    VirtualFile file = hotspot.primaryLocation.file;
+    JLabel fileNameLabel = new JLabel();
+    panel.add(fileNameLabel);
+
+    Integer lineNumber = hotspot.getLineNumber();
+    JLabel lineLabel = new JLabel(file.getName() + (lineNumber != null ? (":" + lineNumber.toString()) : ""));
```
kirill-knize-sonarsource, line 81: Same. No need in parentheses inside ternary operator.

kirill-knize-sonarsource (approved):

src/main/java/org/sonarlint/intellij/actions/IssuesViewTabOpener.java
```diff
@@ -96,4 +97,14 @@ public void showIssueDescription(LiveIssue liveIssue) {
   public void showIssueLocations(LiveIssue liveIssue) {
     showIssue(liveIssue, SonarLintIssuesPanel::selectLocationsTab);
   }
+
+  public void show(LocalHotspot localHotspot) {
+    openTab(SonarLintToolWindowFactory.TAB_HOTSPOTS);
+    ContentManager contentManager = getToolWindow().getContentManager();
+    Content content = contentManager.findContent(SonarLintToolWindowFactory.TAB_HOTSPOTS);
+    if(content.getComponent() instanceof SonarLintHotspotsPanel) {
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: You have a sharp vision ! :)
It can't be something else indeed, I got inspired by similar code in the class. I will remove the check and directly cast the result.

src/main/resources/META-INF/plugin.xml
damien-urruty-sonarsource: No because I also call `setCloseable(false);` on tabs that we don't want to be closeable. In the end only the hotspots tab is closeable

src/main/java/org/sonarlint/intellij/issue/hotspot/LocalHotspot.java
damien-urruty-sonarsource: That's how I had written it at the beginning because they are indeed not needed. I changed it after next complained about it, because of [this rule](https://rules.sonarsource.com/java/RSPEC-864) if I'm correct. I am a bit dubious about it, but we can maybe discuss it later

src/main/java/org/sonarlint/intellij/ui/HotspotCellRenderer.java
damien-urruty-sonarsource: See previous comment

src/main/java/org/sonarlint/intellij/ui/SonarLintHotspotsListPanel.java
damien-urruty-sonarsource: :+1:

src/main/java/org/sonarlint/intellij/core/SecurityHotspotMatcher.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public SecurityHotspotMatcher(Project project) {
+    this.project = project;
+    issueMatcher = new IssueMatcher(project);
+  }
+
+  public LocalHotspot match(RemoteHotspot remoteHotspot) throws IssueMatcher.NoMatchException {
+    for (VirtualFile contentRoot : ProjectRootManager.getInstance(project)
+      .getContentRoots()) {
+      VirtualFile matchedFile = contentRoot.findFileByRelativePath(remoteHotspot.filePath);
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: We look for a perfect match within each content root, that will usually be the project root directory and modules directories. Do you see situations where it won't be able to match ?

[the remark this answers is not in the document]
henryju: Unfortunately, yes :( Let's discuss that.

PR #290, opened 2020-11-04
MMF-2092 Show hotspot in IDE

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/issue/hotspot/SecurityHotspotCategory.java
```diff
@@ -0,0 +1,66 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package org.sonarlint.intellij.issue.hotspot;
+
+import java.util.Arrays;
+import java.util.Optional;
+
+public enum SecurityHotspotCategory {
```
henryju, line 25: Would be great to get rid of this hardcoded mapping before the final release.

henryju (changes requested):
I tried to call `http://127.0.0.1:64120/sonarlint/api/hotspots/show?project=foo&hotspot=bar&server=https://next.sonarqube.com/sonarqube` from my web browser and nothing occurs. Maybe add some messages in the SonarLint console when something goes wrong.

Also I don't like the "Summary" tab very much. This is not really a summary IMO, we could maybe rename it to "Properties" or "Details", and I would rename the "Author" field to to "Assignee" like in SQ UI.

src/main/java/org/sonarlint/intellij/issue/hotspot/SecurityHotspotCategory.java
damien-urruty-sonarsource: As discussed, it would mean getting the long description instead of the short one from SQ, but they don't plan to do it soon, so we have to keep it

damien-urruty-sonarsource:
> I would rename the "Author" field to to "Assignee" like in SQ UI.

Author and Assignee are 2 different things on SQ. The Author is the one who contributed the code (and so introduced the hotspot), the assignee is the person currently responsible of the hotspot (e.g. the one who will fix it). Alex thought it would be valuable to know who introduced the hotspot if you want to contact him/her

src/main/java/org/sonarlint/intellij/server/RequestProcessor.kt
```diff
@@ -0,0 +1,88 @@
...
+        }
+        if (request.path == SHOW_HOTSPOT_ENDPOINT && request.method == HttpMethod.GET) {
+            return processOpenInIdeRequest(request)
+        }
+        return BadRequest("Invalid path or method.")
+    }
+
+    private fun getStatusData(): Response {
+        val description = ProjectManager.getInstance().openProjects.joinToString(", ") { it.name }
+        val status = Status("${appInfo.versionName} ${appInfo.fullVersion}", description)
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: I agree about the version, that could be moved to description because it is unlikely that users will have several instances running at the same time.

I would use the same rationale for the edition name, it is also unlikely that people will run one CE and one Ultimate instance at the same time. I am OK to have it but maybe also in the description ?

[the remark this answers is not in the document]
henryju: Fine by me!

src/main/java/org/sonarlint/intellij/server/RequestProcessor.kt
```diff
@@ -49,8 +53,13 @@ class RequestProcessor(private val appInfo: ApplicationInfo = ApplicationInfo.ge
...
 
     private fun getStatusData(): Response {
-        val description = ProjectManager.getInstance().openProjects.joinToString(", ") { it.name }
-        val status = Status("${appInfo.versionName} ${appInfo.fullVersion}", description)
+        var description = appInfo.fullVersion
+        val edition = ApplicationNamesInfo.getInstance().editionName
+        if (edition != null) {
+            description += " ($edition)"
+        }
+        description += " - " + ProjectManager.getInstance().openProjects.joinToString(", ") { it.name }
```
henryju, line 61: You can have zero projects opened (on the welcome screen).

henryju (approved):
Minor feedback then LGTM

src/main/java/org/sonarlint/intellij/server/RequestProcessor.kt
damien-urruty-sonarsource: Fixed

jblievremont (commented):
Looks good from a technical point of view.

After some end-to-end testing, I have seen that the error cases (connected mode not configured, not the right server, not the right project, etc) currently result in an error message in the "Log" tab of the "SonarLint" tool window.

It's a good first baby step _and_ I expect future versions to have a more visible UI feedback :slightly_smiling_face: e.g a full-fledged notification in the "Event Log".

PR #292, opened 2020-11-12
Dev notifications refresh

henryju:
I suggest you review commits independently since some of them are big refactoring with lot of renames.

### DIFF
--- a/src/main/java/org/sonarlint/intellij/config/global/wizard/AuthStep.java
+++ b/src/main/java/org/sonarlint/intellij/config/global/wizard/AuthStep.java
@@ -39,7 +39,7 @@ import javax.swing.event.DocumentEvent;
 import javax.swing.event.DocumentListener;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
-import org.sonarlint.intellij.config.global.SonarQubeServer;
+import org.sonarlint.intellij.config.global.ServerConnection;
 import org.sonarlint.intellij.tasks.ConnectionTestTask;
 import org.sonarlint.intellij.tasks.GetOrganizationTask;
 import org.sonarlint.intellij.tasks.InformationFetchTask;
@@ -176,7 +176,7 @@ public class AuthStep extends AbstractWizardStepEx {
   }
 
   private void fetchInformation() throws CommitStepException {
-    SonarQubeServer tmpServer = model.createServerWithoutOrganization();
+    ServerConnection tmpServer = model.createServerWithoutOrganization();
     InformationFetchTask task = new InformationFetchTask(tmpServer);
     ProgressManager.getInstance().run(task);
     if (task.getException() != null) {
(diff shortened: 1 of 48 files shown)


src/main/java/org/sonarlint/intellij/core/ProjectServerNotifications.java
```diff
@@ -136,27 +142,39 @@ private NotificationConfiguration createConfiguration(SonarLintProjectSettings s
...
       SonarLintTelemetry telemetry = SonarLintUtils.getService(SonarLintTelemetry.class);
+      boolean isSonarCloud;
+      try {
+        ProjectBindingManager bindingManager = SonarLintUtils.getService(myProject, ProjectBindingManager.class);
+        ServerConnection server = bindingManager.getServerConnection();
+        isSonarCloud = server.isSonarCloud();
+      } catch (InvalidBindingException e) {
+        // do nothing
+        return;
+      }
```
damien-urruty-sonarsource, line 156: It seems a bit strange to me that we have to retrieve the connection from the notification, because we had it when we registered. Would be much simpler if we had notification.getServer() or notification.getConnection(). Or maybe keep the connection as a field of EventListener when we register

src/main/java/org/sonarlint/intellij/core/ProjectServerNotifications.java
```diff
@@ -46,7 +48,8 @@
 import static org.sonarlint.intellij.config.Settings.getSettingsFor;
 
 public class ProjectServerNotifications {
-  private static final NotificationGroup SONARQUBE_GROUP = NotificationGroup.balloonGroup("SonarLint: SonarQube Events");
+  private static final NotificationGroup SERVER_NOTIFICATIONS_GROUP =
+    new NotificationGroup("SonarLint: Server Notifications", NotificationDisplayType.STICKY_BALLOON, true, "SonarLint");
```
damien-urruty-sonarsource, line 52: Probably not a big deal but by changing this text we might lose some user preferences. Users can change settings for specific notifications.
Screenshot from 2020-11-13 14-57-21

damien-urruty-sonarsource (approved):
Minor comments but looks good!

src/main/java/org/sonarlint/intellij/tasks/InformationFetchTask.java
```diff
@@ -32,8 +32,7 @@
 import org.sonarsource.sonarlint.core.client.api.connected.RemoteOrganization;
 import org.sonarsource.sonarlint.core.client.api.connected.ServerConfiguration;
 import org.sonarsource.sonarlint.core.client.api.connected.WsHelper;
-import org.sonarsource.sonarlint.core.client.api.exceptions.UnsupportedServerException;
-import org.sonarsource.sonarlint.core.notifications.SonarQubeNotifications;
+import org.sonarsource.sonarlint.core.notifications.ServerNotifications;
 
 public class InformationFetchTask extends Task.Modal {
```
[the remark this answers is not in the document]
henryju: I agree :)

src/main/java/org/sonarlint/intellij/config/global/wizard/AuthStep.java
```diff
@@ -176,7 +176,7 @@ public JComponent getComponent() {
   }
 
   private void fetchInformation() throws CommitStepException {
-    SonarQubeServer tmpServer = model.createServerWithoutOrganization();
+    ServerConnection tmpServer = model.createServerWithoutOrganization();
```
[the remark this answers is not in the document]
henryju: Yep, I started to rename quite a few, but this will be a continuous effort.

src/main/java/org/sonarlint/intellij/config/project/SearchProjectKeyDialog.java
```diff
@@ -48,17 +48,19 @@
...
   private JBList<RemoteProject> projectList;
 
-  public SearchProjectKeyDialog(Component parent, String selectedProjectKey, Map<String, RemoteProject> projectsByKey) {
+  public SearchProjectKeyDialog(Component parent, String selectedProjectKey, Map<String, RemoteProject> projectsByKey, boolean isSonarCloud) {
     super(parent, false);
     this.lastSelectedProjectKey = selectedProjectKey;
     this.projectsByKey = projectsByKey;
-    setTitle("Search Project in SonarQube Server");
+    this.isSonarCloud = isSonarCloud;
+    setTitle("Search Project in " + (isSonarCloud ? "SonarCloud" : "SonarQube"));
```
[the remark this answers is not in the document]
henryju: I was wondering the same, but there are situations where we don't yet have a ServerConnection object (for example in the wizard). Still it might be interesting.

src/main/java/org/sonarlint/intellij/core/ProjectServerNotifications.java
henryju: I will have a look.

src/main/java/org/sonarlint/intellij/core/ProjectServerNotifications.java
henryju: I know, but I don't think it is worth trying to migrate old settings. Anyway with the refresh of the feature, it is maybe good to "reset" settings.

PR #295, opened 2020-11-18
Update analyzers

henryju:


jblievremont (approved):
Tested in IU 2020.2 and PS 2020.2, LGTM!

PR #298, opened 2020-11-19
Fix binding (restore settings serialization format)

henryju:


jblievremont (approved):
Tested on IU 2020.2, LGTM :+1:

PR #301, opened 2020-12-01
SLI-476 Fix missing text in older versions

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/ui/RecentProjectsPanel.kt
```diff
@@ -131,7 +131,7 @@ open class SonarLintRecentProjectPanel(private val onProjectSelected: (Project)
 
         init {
             setExpandableItemsEnabled(false)
-            setEmptyText(IdeBundle.message("empty.text.no.project.open.yet"))
+            setEmptyText("No project opened yet")
```
henryju, line 134:
```suggestion
            setEmptyText("No projects opened yet")
```

henryju (changes requested):

PR #302, opened 2020-12-07
Add a UI test to check open hotspot feature

damien-urruty-sonarsource:


kirill-knize-sonarsource (approved):

azure-pipelines.yml
```diff
@@ -130,31 +130,81 @@ stages:
...
-          Idea2018:
-            IDEA_VERSION: '2018.3'
-            JDKVersion: '1.11'
+          # remote-robot crashes on 2018.3
           Idea2019:
-            IDEA_VERSION: '2019.2'
-            JDKVersion: '1.11'
+            IDEA_VERSION: 'IC-2019.3'
+          Idea2020:
+            IDEA_VERSION: 'IC-2020.1.3'
```
henryju, line 137: Can we add LATEST EAP?

azure-pipelines.yml
```diff
@@ -130,31 +130,81 @@ stages:
...
+          dir=~/.java/.userPrefs/jetbrains/_\!\(\!\!cg\"p\!\(\}\!\}@\"j\!\(k\!\|w\"w\!\'8\!b\!\"p\!\':\!e@==
+          mkdir -p "$dir"
+          cp .azure-pipelines/prefs.xml $dir
+
+          # copy data sharing prefs
+          mkdir -p ~/.local/share/JetBrains/consentOptions
+          cp .azure-pipelines/accepted ~/.local/share/JetBrains/consentOptions
+
+          echo 'Recording tests on video'
+          ffmpeg -loglevel warning -f x11grab -video_size 1024x768 -framerate 10 -i :10 -q:v 0 /tmp/recording_$(IDEA_VERSION).mp4 &
```
henryju, line 190: Can't wait to do the same on SLE ;)

henryju (approved):

azure-pipelines.yml
damien-urruty-sonarsource: We could use `LATEST-EAP-SNAPSHOT` yes. I was a bit reluctant because I don't know if it would break our tests too often. Maybe best way is to try and see! I'll make the change

henryju (approved):
LGTM!

PR #303, opened 2020-12-11
SLI-475 Use the plugin verifier during the build

damien-urruty-sonarsource:


damien-urruty-sonarsource:
FYI @henryju @jblievremont we now check binary compatibility in our CI with the plugin verifier

kirill-knize-sonarsource (approved):

azure-pipelines.yml
```diff
@@ -38,7 +38,7 @@ stages:
       - task: Gradle@2
         inputs:
           options: --build-cache --stacktrace -i
-          tasks: :buildPlugin :artifactoryPublish
+          tasks: :buildPlugin :artifactoryPublish :runPluginVerifier
```
henryju, line 41: IMO that would have been better in the validate stage. The goal of the build stage is to produce an artifact as fast as possible, and then run the rest of the pipeline in parallel, consuming the same artifact.

damien-urruty-sonarsource: You're probably right. I will make the change

PR #304, opened 2020-12-24
SLI-478 SLI-489 Improve error handling for server notifications

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/BootstrapStartupActivity.java
```diff
@@ -33,9 +34,13 @@
 
   @Override
   public void runActivity(@NotNull Project project) {
+    if (ApplicationManager.getApplication().isUnitTestMode()) {
```
henryju, line 37: Is it really necessary? Won't it remove a lot of coverage?

src/test/java/org/sonarlint/intellij/core/ProjectServerNotificationsSubscriberTest.kt
```diff
@@ -0,0 +1,165 @@
...
+
+  @Captor
+  private lateinit var notificationConfigurationCaptor: ArgumentCaptor<NotificationConfiguration>
+
+  @Captor
+  private lateinit var balloonNotificationCaptor: ArgumentCaptor<Notification>
+
+  @Before
+  fun setup() {
+    replaceProjectService(BalloonNotifier::class.java, balloonNotifier)
```
henryju, line 58: Instead of mocking notification, another option is to let the real component, and add a listener: https://github.com/jansorg/intellij-notifications/blob/master/src/test/java/dev/ja/samples/ConfigurableNotificationTest.java

henryju (approved):
LGTM!

src/main/java/org/sonarlint/intellij/BootstrapStartupActivity.java
damien-urruty-sonarsource: It won't. All the services loaded by this class already have dedicated tests.

The problem is that this method was called automatically in light tests when the project is created, performing undesired side effects.

src/test/java/org/sonarlint/intellij/core/ProjectServerNotificationsSubscriberTest.kt
damien-urruty-sonarsource: Good to know! I added this BalloonNotifier some time ago to be able to mock and check the call to notify. I will try to replace with this mechanism

henryju (approved):

PR #305, opened 2021-01-04
MMF-2126 Display taint vulnerabilities in IDE

damien-urruty-sonarsource:


henryju (changes requested):
I found some small issues during my test:
* the default message when the panel is empty is not correctly vertically centered when the panel is resized
* clicking on "learn more" link doesn't work on KDE
* clicking on the "dismiss" link seems to take too long to refresh the panel

henryju (approved):

PR #308, opened 2021-01-07
Use toolchains to compile with JDK 8

henryju:


damien-urruty-sonarsource (approved):
I didn't know this option! It will help have more controlled build config. :+1:

PR #315, opened 2021-02-11
SLI-486 Match taint vulnerabilities locations source code

damien-urruty-sonarsource:


henryju:
Not related to this PR, but I find the logs in the SonarLint too verbose when "verbose" mode is enabled. There are all logs from Apache httpclient, that are not very useful IMO.

henryju:
I think I found an issue:
* open a file with some taint vulnerabilities
* change the content of the file (like add new lines at the top)
* locations are marked as "not matching" in the treeview, but are still correctly highlighted in the file (thanks to IJ live ranges I guess)
* now close and reopen the file: the locations in the code editor are misplaced:
image

PR #317, opened 2021-02-16
SLI-501 Replace platform logger usage by our own to log an error

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/issue/vulnerabilities/TaintVulnerabilitiesLoader.kt
```diff
@@ -60,7 +58,7 @@ object TaintVulnerabilitiesLoader {
       ?: throw InvalidBindingException("Module ${module.name} is not bound")
     val filePath = project.getRelativePathOf(file)
     if (filePath == null) {
-      LOG.error("Filepath for file ${file.canonicalPath} was not resolved.")
+      SonarLintConsole.get(project).error("Filepath for file ${file.canonicalPath} was not resolved.")
```
henryju, line 61: I wonder if this is not going to be too verbose as well in the SL console? Maybe it should be a debug log?

damien-urruty-sonarsource: Changed

PR #324, opened 2021-03-19
SLI-509 Generate build-wrapper-content

mpaladin:


build.gradle.kts
```diff
@@ -174,6 +174,13 @@ project(":clion") {
     }
     dependencies {
         compile(project(":common"))
+        testImplementation(platform("org.junit:junit-bom:5.7.1"))
+        testImplementation("org.junit.jupiter:junit-jupiter")
+        testImplementation("org.mockito:mockito-core:2.19.0")
+    }
+    tasks.test {
+        useJUnitPlatform()
+        systemProperty("sonarlint.telemetry.disabled", "true")
```
henryju, line 183: Doesn't hurt, but I'm pretty sure this is not needed here, since you are only unit testing CLion specific code

henryju (approved):

build.gradle.kts
mpaladin: ok, thanks @henryju !

PR #328, opened 2021-03-25
SLI-509 Add CFamilyFileExclusionContributor, populate CFamilyAnalysisConfigurator forced languages, use runReadAction

mpaladin:


clion/src/main/java/org/sonarlint/intellij/clion/CFamilyFileExclusionContributor.java
```diff
@@ -0,0 +1,20 @@
...
+
+public class CFamilyFileExclusionContributor implements FileExclusionContributor {
+
+  @Override
+  public ExcludeResult shouldExclude(Module module, VirtualFile fileToAnalyze) {
+    AnalyzerConfiguration.ConfigurationResult configurationResult = new AnalyzerConfiguration(module.getProject()).getConfiguration(fileToAnalyze);
+    if (configurationResult.hasConfiguration()) {
+      return ExcludeResult.notExcluded();
+    }
+    SonarLintConsole.get(module.getProject()).debug(configurationResult.getSkipReason());
```
henryju, line 36: If I'm not mistaken, there is already a debug log in the calling code

henryju (approved):
LGTM, but I have not manually tested. Let me know if you think I should.

clion/src/main/java/org/sonarlint/intellij/clion/CFamilyFileExclusionContributor.java
mpaladin: @henryju ideed, I checked and it is the case, nice catch.

Godin (approved):

PR #341, opened 2021-04-01
Rework LocalFileExclusions

henryju:


src/test/java/org/sonarlint/intellij/AbstractSonarLintLightTests.java
```diff
@@ -135,12 +137,27 @@ public void loadToolWindow() {
...
+    ProjectConfigurationListener projectListener = getProject().getMessageBus().syncPublisher(ProjectConfigurationListener.TOPIC);
+    projectListener.changed(projectSettings);
+  }
+
+  protected void setGlobalLevelExclusions(List<String> exclusions) {
+    SonarLintGlobalSettings globalSettings = getGlobalSettings();
+    globalSettings.setFileExclusions(exclusions);
+    GlobalConfigurationListener globalConfigurationListener = ApplicationManager.getApplication()
+      .getMessageBus().syncPublisher(GlobalConfigurationListener.TOPIC);
+    globalConfigurationListener.applied(globalSettings);
```
damien-urruty-sonarsource, line 161: I find it strange that we have to take care of this in the test code, we probably do this in production code at some point after the settings are changed ?

src/main/java/org/sonarlint/intellij/analysis/LocalFileExclusions.java
```diff
@@ -145,26 +143,124 @@ private ExcludeResult checkFileInSourceFolders(VirtualFile file, Module module)
...
+    throws InvalidBindingException {
+    Map<Module, Collection<VirtualFile>> filesByModule = new LinkedHashMap<>();
+
+    for (VirtualFile file : files) {
+      checkExclusionsFileByFile(forcedAnalysis, excludedFileHandler, filesByModule, file);
+    }
+    if (!forcedAnalysis && !filesByModule.isEmpty()) {
+      // Apply server file exclusions. This is an expensive operation, so we call the core only once per module.
+      filterWithServerExclusions(excludedFileHandler, filesByModule);
     }
```
damien-urruty-sonarsource, line 157: If I'm not wrong, here we could check server exclusions even if the file was already excluded in `checkExclusionsFileByFile`. It sounds like a useless check

damien-urruty-sonarsource (approved):
A few comments but LGTM

src/test/java/org/sonarlint/intellij/AbstractSonarLintLightTests.java
henryju: This is what we do in the production code. We could extract that in a helper, but for now I think it's ok to keep it like this.

src/main/java/org/sonarlint/intellij/analysis/LocalFileExclusions.java
henryju: `filesByModule` contains only non excluded files after `checkExclusionsFileByFile`.

PR #346, opened 2021-04-07
SLI-516 Stream processing of issues

henryju:


src/main/java/org/sonarlint/intellij/issue/IssueManager.java
```diff
@@ -252,7 +255,8 @@ public void insertNewIssue(VirtualFile file, LiveIssue liveIssue) {
   }
 
   private void notifyAboutIssueChangeForFile(VirtualFile file) {
-    myProject.getMessageBus().syncPublisher(IssueStoreListener.SONARLINT_ISSUE_STORE_TOPIC).fileChanged(file, liveIssueCache.getLive(file));
+    Collection<LiveIssue> issues = liveIssueCache.getLive(file) == null ? new ArrayList<>() : liveIssueCache.getLive(file);
```
henryju, line 258: This is not correct. `liveIssueCache.getLive(file)` can return non null the first time, and `null` the second time...

src/main/java/org/sonarlint/intellij/issue/persistence/LiveIssueCache.java
```diff
@@ -87,7 +85,10 @@ protected boolean removeEldestEntry(Map.Entry<VirtualFile, Collection<LiveIssue>
    */
   @CheckForNull
   public synchronized Collection<LiveIssue> getLive(VirtualFile virtualFile) {
-    return cache.get(virtualFile);
+    if(cache.get(virtualFile) == null) {
```
henryju, line 88: Same here, even if it is synchronized, I would change the code to be more concurrent friendly. Please also ensure you follow code formatting conventions

src/main/java/org/sonarlint/intellij/issue/IssueManager.java
henryju, line 258:
```suggestion
    Collection<LiveIssue> issues = liveIssueCache.getLive(file);
    if (issues == null) {
      issues = Collections.emptyList();
    }
```

src/main/java/org/sonarlint/intellij/issue/IssueManager.java
```diff
@@ -252,7 +255,8 @@ public void insertNewIssue(VirtualFile file, LiveIssue liveIssue) {
...
+    myProject.getMessageBus().syncPublisher(IssueStoreListener.SONARLINT_ISSUE_STORE_TOPIC).fileChanged(file, Objects.requireNonNull(issues));
```
henryju, line 259: `requireNonNull` is not needed, we don't use this in our code base

PR #365, opened 2021-04-21
SLI-526 Fix an error when annotating a file that is no longer valid

damien-urruty-sonarsource:


henryju (approved):
LGTM, but please don't merge until release

PR #374, opened 2021-05-03
SLI-531 Recover from issue index store corruption

henryju:


src/main/java/org/sonarlint/intellij/issue/persistence/StringStoreIndex.java
```diff
@@ -76,7 +79,8 @@ private void save(Sonarlint.StorageIndex index) {
     try (OutputStream stream = Files.newOutputStream(indexFilePath)) {
       index.writeTo(stream);
     } catch (IOException e) {
-      throw new IllegalStateException("Failed to write local issue store index", e);
+      // Don't log in the SonarLint console as the problem can occurs when stopping the IDE
+      LOG.warn("Unable to write SonarLint issue store.", e);
```
damien-urruty-sonarsource, line 83: What if it happens not when closing the IDE ? Shouldn't we throw here and log the error from the stop method ? As the log output can be hard to find I'm a bit concerned to "just" log the error

henryju: I think the behavior is that issues are not persisted, and state will be lost after restarting the IDE. Not a big deal IMO, and much better than crashing the analysis. I don't think there is anything we can do here...

PR #384, opened 2021-05-25
Update robot framework

henryju:


its/src/test/kotlin/org/sonarlint/intellij/its/BaseUiTest.kt
```diff
@@ -140,8 +140,8 @@ open class BaseUiTest {
         }
         searchField().text = fileName
         val fileList = jList(JListFixture.byType(), Duration.ofSeconds(5))
-        waitFor(Duration.ofSeconds(5)) { fileList.items.isNotEmpty() }
-        fileList.selectItem(fileList.items[0], false)
+        waitFor(Duration.ofSeconds(5)) { fileList.collectItems().isNotEmpty() }
+        fileList.clickItemAtIndex(0)
         waitFor(Duration.ofSeconds(10)) { editor(fileName).isShowing }
         waitBackgroundTasksFinished()
```
damien-urruty-sonarsource, line 146: A bit of duplication here

damien-urruty-sonarsource (approved):
Minor comments, else :+1:

its/src/main/resources/META-INF/plugin.xml
```diff
@@ -27,4 +27,6 @@
   <description>Fake Plugin To Run SonarLint UI tests with multiple flavors</description>
 
   <idea-version since-build="181.3007.14"/>
+
+  <depends>com.intellij.modules.lang</depends>
```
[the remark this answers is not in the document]
henryju: There was an error in CLion logs, saying that plugins without dependencies are only supported in IDEA. Not sure it was a blocker, but this make logs cleaner.

its/src/test/kotlin/org/sonarlint/intellij/its/BaseUiTest.kt
henryju: Indeed, but I think we can improve that later

PR #387, opened 2021-05-31
SLI-550 Map module lifecycle management and file system

jblievremont:


src/main/java/org/sonarlint/intellij/module/ModuleFileSystem.kt
```diff
@@ -0,0 +1,52 @@
...
+      .filter { f -> f.relativePath().endsWith(language) }
+      .filter { f -> f.isTest == (type == InputFile.Type.TEST) }
+  }
+
+  override fun files(): Stream<ClientInputFile> {
+    val files: MutableList<ClientInputFile> = ArrayList()
+    val sonarLintAnalyzer = project.getService(SonarLintAnalyzer::class.java)
+    ModuleRootManager.getInstance(module).fileIndex.iterateContent { fileOrDir: VirtualFile ->
+        if (!fileOrDir.isDirectory && !ProjectCoreUtil.isProjectOrWorkspaceFile(fileOrDir, fileOrDir.fileType)) {
+          files.add(sonarLintAnalyzer.createClientInputFile(module, fileOrDir, null)!!)
```
damien-urruty-sonarsource, line 46: I have a doubt about this `!!`. Down createClientInputFile we try to get the relative path using the project base dir and I'm not sure it will always return something. By the way, I think in some cases this relative path could be wrong (and I also use the same logic)

jblievremont: To be honest, the `!!` was generated by the Java to KT conversion tool. I'm not sure about what would be the appropriate unsurprising behavior here, maybe try to create the client input file and only add it if it is not `null`.

damien-urruty-sonarsource: I think so yes, else it will throw when trying to dereference the `null`.

PR #388, opened 2021-05-31
SLI-551 Notify analyzers of file system events

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/core/ConnectedSonarLintFacade.java
```diff
@@ -93,6 +93,11 @@ protected AnalysisResults analyze(Path baseDir, Path workDir, Collection<ClientI
     return engine.getPluginDetails();
   }
 
+  @Override
+  public ConnectedSonarLintEngine getEngine() {
```
jblievremont, line 97: Ah, I might have added this one also on my side 😉 making a note to be vigilant during rebase/merge.

jblievremont (approved):
Looks good to me 👍🏻 let's merge and test with the full stack.

PR #398, opened 2021-06-10
SLI-551 Use document listener to notify analyzers of changes in open files

jblievremont:


src/main/java/org/sonarlint/intellij/fs/EditorFileChangeListener.kt
```diff
@@ -19,26 +19,69 @@
...
-        val engine = getService(project, ProjectBindingManager::class.java).engineIfStarted ?: return
-        val file = FileDocumentManager.getInstance().getFile(event.document) ?: return
+const val DEBOUNCE_DELAY_MS = 1000L
+
+class EditorFileChangeListener : BulkAwareDocumentListener.Simple, StartupActivity {
+
+    private var fileEventsNotifier: ModuleFileEventsNotifier? = null
+    private lateinit var project: Project
+    private var eventsToSend = LinkedHashMap<Module, MutableMap<String, ClientModuleFileEvent>>()
+    private var triggerTimer = initTimer()
```
damien-urruty-sonarsource, line 45: Not too sure but as you create a timer instance, it might be better to dispose it when the project closes. Maybe it's sufficient to implement Disposable here and stop the timer, but not sure it would be called. Another option is to call `Disposer.register` with the project as the parent

src/main/java/org/sonarlint/intellij/fs/VirtualFileSystemListener.kt
```diff
@@ -96,6 +97,7 @@ class VirtualFileSystemListener(
         for (event in events) {
             // call event.file only once as it can be hurting performance
             val file = event.file ?: continue
+            if(ProjectCoreUtil.isProjectOrWorkspaceFile(file, file.fileType)) continue;
```
damien-urruty-sonarsource, line 100: :+1:

damien-urruty-sonarsource (approved):

PR #399, opened 2021-06-11
Hardening

henryju:


src/main/java/org/sonarlint/intellij/analysis/DefaultClientInputFile.java
```diff
@@ -93,7 +96,15 @@ private URI createURI() {
...
-      return new String(vFile.contentsToByteArray(), charset);
+      ByteArrayOutputStream result = new ByteArrayOutputStream();
+      try (InputStream inputStream = inputStream()) {
+        byte[] buffer = new byte[DEFAULT_BUFFER_SIZE];
+        int length;
+        while ((length = inputStream.read(buffer)) != -1) {
+          result.write(buffer, 0, length);
+        }
+      }
+      return result.toString(charset.name());
```
damien-urruty-sonarsource, line 107: We could also use `IOUtils.toString(InputStream, Charset)`. It's from commons-io but provided through IJ dependency

damien-urruty-sonarsource (approved):
:+1:

src/main/java/org/sonarlint/intellij/analysis/DefaultClientInputFile.java
henryju: :+1:

PR #402, opened 2021-06-17
SLI-561 Check exclusions in analysis task

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/analysis/ForcedAnalysisTask.java
```diff
@@ -26,18 +26,18 @@
 /**
  * A modal task (blocking) initiated explicitly by the user.
  */
-class SonarLintUserTask extends SonarLintTask {
+class ForcedAnalysisTask extends AnalysisTask {
```
henryju, line 29: Other suggestion: `UserTriggeredAnalysisTask

henryju (approved):
Nice!

src/main/java/org/sonarlint/intellij/analysis/ForcedAnalysisTask.java
damien-urruty-sonarsource: :+1:

PR #403, opened 2021-06-21
SLI-564 Fix an exception when creating ModuleChangeListener

damien-urruty-sonarsource:


jblievremont (approved):
Nice! Tested in IC 2019.3.5 with Python Community (merged with #404)

PR #406, opened 2021-06-23
Pass PID to engines

henryju:


src/main/java/org/sonarlint/intellij/core/SonarLintEngineFactory.java
```diff
@@ -61,34 +65,25 @@
...
     URL cFamilyPluginUrl = findEmbeddedCFamilyPlugin(getPluginsDir());
-    ConnectedGlobalConfiguration.Builder config = ConnectedGlobalConfiguration.builder()
-      .setLogOutput(globalLogOutput)
-      .setSonarLintUserHome(getSonarLintHome())
+    ConnectedGlobalConfiguration.Builder configBuilder = ConnectedGlobalConfiguration.builder()
       .addEnabledLanguages(enabledLanguages.toArray(new Language[0]))
-      .setExtraProperties(prepareExtraProps())
-      .setNodeJs(nodeJsManager.getNodeJsPath(), nodeJsManager.getNodeJsVersion())
-      .setWorkDir(getWorkDir())
-      .setModulesProvider(() -> modulesRegistry.getModulesForEngine(connectionId))
```
jblievremont, line 84: This got lost in the refactoring to `configureCommonEngine`, the management of modules is not the same for connected engines vs the standalone engine.

jblievremont (changes requested):
A small difference on module management between connected and standalone engines got lost in the refactoring, else LGTM.

src/main/java/org/sonarlint/intellij/core/SonarLintEngineFactory.java
henryju: Oups, good catch indeed!

jblievremont (approved):
LGTM! 👍🏻

PR #407, opened 2021-06-24
SLI-567 Update embedded Kotlin analyzer to 2.0

Godin:


henryju (approved):
Tested, works fine.
Feedback:
* I noticed that the rule `kotlin:ParsingError` is visible in SonarLint. I think it might be better to simply drop this rule in the SonarLint context.
* Shouldn't Gradle *.kts files be considered out of the box?

PR #409, opened 2021-06-28
add rules data to telemetry

kirill-knize-sonarsource:


src/main/java/org/sonarlint/intellij/analysis/AnalysisTask.java
```diff
@@ -284,6 +285,9 @@ private void processRawIssue(LiveIssueBuilder issueBuilder, IssueManager manager
       LiveIssue locallyTrackedIssue = manager.trackSingleIssue(vFile, previousIssues, liveIssue);
       issuesPerFile.computeIfAbsent(vFile, f -> new ArrayList<>()).add(locallyTrackedIssue);
     }
+
+    SonarLintTelemetry telemetry = SonarLintUtils.getService(SonarLintTelemetry.class);
+    telemetry.addReportedRule(liveIssue.getRuleKey());
```
henryju, line 290: I wonder if there is not a risk of performance decrease with such call. On telemetry side, we take a lock on the telemetry file every time we write to it.
I think we should change the API to allow a single call per analysis, where we report all rules in one call.

damien-urruty-sonarsource:
I was having a look at what you did for rules telemetry, to integrate it to our eclipse sprint. I have some feedback:

* ~~looks like at least for raisedIssuesRules you accumulate rule keys, this means you could have something like "raised_issues": ["java:S1234", "java:S1234", "java:S1234"]. Is it what we expect to collect ?~~ In fact you use a Set to store those values : https://github.com/SonarSource/sonarlint-core/blob/master/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryLocalStorage.java#L55. It is a bit hidden, maybe we could make it more explicit by having a Set<String> instead of a Collection<String> everywhere this value is mentioned
* the ticket https://jira.sonarsource.com/browse/SLCORE-311 does not reflect the names you finally chose for the payload fields
* in IntelliJ when we change the telemetry payload there is also the sample to update: https://github.com/SonarSource/sonarlint-intellij/blob/master/src/main/java/org/sonarlint/intellij/config/global/SonarLintAboutPanel.java#L132

henryju (approved):

PR #410, opened 2021-07-01
Rider hardening

henryju:


src/main/java/org/sonarlint/intellij/core/SecurityHotspotMatcher.java
```diff
@@ -48,11 +48,17 @@ public LocalHotspot match(ServerHotspot serverHotspot) {
...
-        return matchTextRange(matchedFile, serverHotspot.textRange, serverHotspot.message);
+    for (VirtualFile contentRoot : ProjectRootManager.getInstance(project).getContentRoots()) {
+      if (contentRoot.isDirectory()) {
+        VirtualFile matchedFile = contentRoot.findFileByRelativePath(serverHotspot.filePath);
+        if (matchedFile != null) {
+          return matchTextRange(matchedFile, serverHotspot.textRange, serverHotspot.message);
+        }
+      } else {
+        // On Rider, all source files are returned as individual content roots, so simply check for equality
+        if (contentRoot.getPath().endsWith(serverHotspot.filePath)) {
```
damien-urruty-sonarsource, line 59: Couldn't we match completely unrelated paths using `endsWith` ?

src/main/java/org/sonarlint/intellij/issue/vulnerabilities/TaintVulnerabilityMatcher.kt
```diff
@@ -40,9 +40,16 @@ class TaintVulnerabilityMatcher(private val project: Project) {
   private fun matchLocation(location: ServerIssueLocation): Location {
     for (contentRoot in ProjectRootManager.getInstance(project)
       .contentRoots) {
-      val matchedFile = contentRoot.findFileByRelativePath(location.filePath!!)
-      if (matchedFile != null) {
-        return matchTextRange(matchedFile, location)
+      if (contentRoot.isDirectory()) {
+        val matchedFile = contentRoot.findFileByRelativePath(location.filePath!!)
```
damien-urruty-sonarsource, line 44: I suppose here `!!` is safe because we can be sure a taint will always have a file path

damien-urruty-sonarsource (approved):

src/main/java/org/sonarlint/intellij/core/SecurityHotspotMatcher.java
henryju: Yep, it is possible, but quite unlikely, because this is the relative path from Solution baseDir, so having a mismatch would be really strange.

src/main/java/org/sonarlint/intellij/issue/vulnerabilities/TaintVulnerabilityMatcher.kt
henryju: It was already used few lines before, so I did not question that ;)

PR #414, opened 2021-07-07
SLI-582 fix for rules telemetry

kirill-knize-sonarsource:


src/test/java/org/sonarlint/intellij/telemetry/TelemetryClientAttributeProviderImplTest.kt
```diff
@@ -0,0 +1,56 @@
...
+    fun nonDefaultEnabledRules() {
+        assertThat(underTest.nonDefaultEnabledRules).isEmpty()
+
+        Settings.getGlobalSettings().enableRule("ruleKey1")
+        Settings.getGlobalSettings().enableRule("ruleKey1")
+        Settings.getGlobalSettings().enableRule("ruleKey2")
+        Settings.getGlobalSettings().setRuleParam("java:S107", "max", "10")
+
+        assertThat(underTest.nonDefaultEnabledRules).hasSize(2)
+        assertThat(underTest.nonDefaultEnabledRules).containsOnly("ruleKey1", "ruleKey2")
```
henryju, line 53: `containsOnly` will already assert the size, so no need for the `hasSize` assertion

src/test/java/org/sonarlint/intellij/telemetry/TelemetryClientAttributeProviderImplTest.kt
```diff
@@ -0,0 +1,56 @@
...
+    @Test
+    fun defaultDisabledRules() {
+        assertThat(underTest.defaultDisabledRules).isEmpty()
+
+        Settings.getGlobalSettings().disableRule("ruleKey1")
+        Settings.getGlobalSettings().disableRule("ruleKey1")
+        Settings.getGlobalSettings().disableRule("ruleKey2")
+
+        assertThat(underTest.defaultDisabledRules).hasSize(2)
+        assertThat(underTest.defaultDisabledRules).containsOnly("ruleKey1", "ruleKey2")
```
henryju, line 40: `containsOnly` will already assert the size, so no need for the `hasSize` assertion

src/test/java/org/sonarlint/intellij/telemetry/TelemetryClientAttributeProviderImplTest.kt
kirill-knize-sonarsource: Changed to `containsExactly`.

src/test/java/org/sonarlint/intellij/telemetry/TelemetryClientAttributeProviderImplTest.kt
kirill-knize-sonarsource: Same.

henryju (approved):

src/test/java/org/sonarlint/intellij/telemetry/TelemetryClientAttributeProviderImplTest.kt
damien-urruty-sonarsource: Sorry for the late feedback here. The difference between `containsOnly` and `containsExactly` is that the second will check that the elements order is the same in actual and expected.
1) I am not sure the order really matters in this case, we just want to check that the elements are present
2) it could lead to subtle random test failures in some cases, depending on how the elements will be returned by the underlying Set.

So i would follow Julien's advice which is to keep containsOnly but remove hasSize that is redundant

henryju: Or use `containsExactlyInAnyOrder` :)

PR #415, opened 2021-07-15
Hardening day 1

henryju:


src/main/java/org/sonarlint/intellij/actions/SonarAnalyzeAllFilesAction.java
```diff
@@ -43,6 +43,7 @@
 import org.sonarlint.intellij.util.SonarLintUtils;
 
 public class SonarAnalyzeAllFilesAction extends AbstractSonarAction {
+
```
jblievremont, line 46: Nitpick: unnecessary diff

src/main/java/org/sonarlint/intellij/notifications/OpenProjectSettingsAction.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.options.ShowSettingsUtil;
+import com.intellij.openapi.project.Project;
+import org.jetbrains.annotations.NotNull;
+import org.sonarlint.intellij.config.project.SonarLintProjectConfigurable;
+
+class OpenProjectSettingsAction extends NotificationAction {
```
jblievremont, line 30: I wonder if it would be worth refactoring this with `OpenGlobalSettingsAction`, the code is very similar and the only differences are the title string and the `Configurable`.

jblievremont (approved):
LGTM 👍🏻 !

src/main/java/org/sonarlint/intellij/actions/SonarAnalyzeAllFilesAction.java
henryju: Fixed

src/main/java/org/sonarlint/intellij/notifications/OpenProjectSettingsAction.java
henryju: Fixed

PR #419, opened 2021-07-22
Hardening (day 5)

henryju:


src/main/java/org/sonarlint/intellij/ui/RecentProjectsPanel.kt
```diff
@@ -183,7 +183,7 @@ open class SonarLintRecentProjectPanel(private val onProjectSelected: (Project)
     }
 
     init {
-        var recentProjectActions = RecentProjectsManager.getInstance().getRecentProjectsActions(false, false).toList()
+        var recentProjectActions = RecentProjectListActionProvider.getInstance().getActions(false, false).toList()
```
jblievremont, line 186: Looks like this class was only [introduced in 2020.1](https://github.com/JetBrains/intellij-community/commit/8c4e82f42915f98af74b72fba32ceb8f0c2e012c) 😕

jblievremont (changes requested):

src/main/java/org/sonarlint/intellij/ui/RecentProjectsPanel.kt
henryju: Indeed, the QA is red on 2019.3 axis. I will revert this change.

src/main/java/org/sonarlint/intellij/trigger/EditorChangeTrigger.java
```diff
@@ -68,7 +68,7 @@ public void started(AnalysisRequest analysisRequest) {
         }
       });
     watcher.start();
-    EditorFactory.getInstance().getEventMulticaster().addDocumentListener(this);
+    EditorFactory.getInstance().getEventMulticaster().addDocumentListener(this, this);
```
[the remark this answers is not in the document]
henryju: This is a parentDisposable. When the parent object (=this) is disposed, then the listener (=this) will be automatically removed from the list of document listeners. I think this code is only sensitive for dynamic unloading, so we will have to retest it anyway when reworking on the topic.

src/main/java/org/sonarlint/intellij/server/SonarLintHttpServer.kt
```diff
@@ -119,25 +121,29 @@ class ServerHandler : SimpleChannelInboundHandler<Any?>() {
...
-            response = RequestProcessor().processRequest(Request(msg.uri(), msg.method()))
+            response = RequestProcessor().processRequest(Request(msg.uri(), msg.method(), isTrustedOrigin(origin)))
         }
         if (msg is LastHttpContent) {
             ctx.writeAndFlush(createResponse(response, origin))
         }
     }
 
+    private fun isTrustedOrigin(origin: String?): Boolean {
+        return origin != null && (SonarLintUtils.isSonarCloudAlias(origin) || !Settings.getGlobalSettings().getConnectionsTo(origin).isEmpty())
```
jblievremont, line 132: Spotted an issue while testing: the condition in `getConnectionsTo(String)` is not the right one for this need.

If the server is hosted under a directory - e.g http://some.host/dir - then the `Origin` will be http://some.host (without `/dir`). According to the `Origin` header specification, the correct check in our case would be to find a server connection whose `hostUrl` starts with the passed `Origin`.

jblievremont (changes requested):
Manually tested:
* Secrets in CLion and Rider 👍🏻  
* Check on trusted servers 👎🏻 - see my comment

jblievremont (approved):
LGTM! 👍🏻

PR #426, opened 2021-08-03
SLI-596 Fix check-in analysis error when a file appears in different changelists

damien-urruty-sonarsource:


jblievremont (approved):
LGTM 👍🏻

PR #442, opened 2021-10-05
MMF-2471 Module binding base feature branch

kirill-knize-sonarsource:


src/main/java/org/sonarlint/intellij/config/module/SonarLintModuleSettings.java
```diff
@@ -22,6 +22,23 @@
...
+  public String getProjectKey() {
+    return projectKey;
+  }
+
+  public void overrideProjectBinding(String projectKey) {
+    this.projectKey = projectKey;
+  }
+
+  public boolean isProjectBindingOverridden() {
+    return !projectKey.equals("");
```
henryju, line 36:
```suggestion
    return !StringUtils.isBlank(projectKey);
```
will be safer

henryju (changes requested):

PR #461, opened 2021-10-15
SLI-621 Add support for CLion 2021.3

mpaladin:


henryju:
@mpaladin I added a commit to enable CLion EAP axis again

PR #468, opened 2021-11-03
SLI-629 Fix issue with HTTP protocol version

jblievremont:


src/main/java/org/sonarlint/intellij/http/ApacheHttpResponse.kt
```diff
@@ -32,7 +32,7 @@ internal class ApacheHttpResponse(
 
   override fun bodyAsString(): String = response.bodyText
 
-  override fun bodyAsStream() = ByteArrayInputStream(response.bodyBytes)
+  override fun bodyAsStream() = ByteArrayInputStream(if(response.bodyBytes != null) response.bodyBytes else ByteArray(0))
```
damien-urruty-sonarsource, line 35: I _think_ the getBodyBytes() method does some computation behind the scenes so maybe we should call it only once ?

damien-urruty-sonarsource (approved):

PR #481, opened 2022-01-11
Bump omnisharp version + plugin version

rita-sonarsource:


henryju (changes requested):
I guess there are some significant improvements/new rules. So the changelog should be updated. See the file `src/main/resources/META-INF/plugin.xml`

PR #485, opened 2022-01-21
SLI-659 Subscribe for server events

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/core/server/events/BindingChangeListener.kt
```diff
@@ -0,0 +1,32 @@
...
+package org.sonarlint.intellij.core.server.events
+
+import com.intellij.openapi.project.Project
+import org.sonarlint.intellij.common.util.SonarLintUtils.getService
+import org.sonarlint.intellij.core.ProjectBinding
+import org.sonarlint.intellij.messages.ProjectBindingListener
+
+class BindingChangeListener(project: Project) : ProjectBindingListener {
+    override fun bindingChanged(previousBinding: ProjectBinding, newBinding: ProjectBinding) {
+        getService(ServerEventsService::class.java).autoSubscribe(previousBinding)
```
henryju, line 29: What is it for? Why subscribing to previous binding?

damien-urruty-sonarsource: This naming of `ProjectBinding` and `previousBinding `is really misleading, I plan to change it

This `bindingChanged` callback can be called when you switch from one connection to another, so you need to re-register for the previous connection with up-to-date project keys

PR #490, opened 2022-02-16
SLI-664 Replace the use of deprecated AnnotationHolder.createAnnotation

damien-urruty-sonarsource:


src/main/resources/META-INF/plugin.xml
```diff
@@ -136,7 +136,7 @@
 
     <!-- please see http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/build_number_ranges.html for description -->
     // we need listeners registration
```
henryju, line 138: Shouldn't we remove this comment?

henryju (approved):

PR #493, opened 2022-02-18
SLI-646 Escape multi-values properties

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/java/JavaAnalysisConfigurator.java
```diff
@@ -88,11 +89,16 @@ public AnalysisConfiguration configure(@NotNull Module ijModule, Collection<Virt
 
   private static void setMultiValuePropertyIfNonEmpty(Map<String, String> properties, String propKey, Set<String> values) {
     if (!values.isEmpty()) {
-      var joinedLibs = StringUtils.join(values, SEPARATOR);
+      var joinedLibs = StringUtils.join(values.stream().map(JavaAnalysisConfigurator::csvEscape).collect(Collectors.toList()), SEPARATOR);
```
henryju, line 92: We should only escape when there is a comma in a path, because old versions of SonarJava will not support the CSV syntax

henryju (changes requested):

henryju (approved):

PR #496, opened 2022-02-21
SLI-680 Run smart notifications support check outside of UI-thread

damien-urruty-sonarsource:


damien-urruty-sonarsource:
@henryju as we will probably migrate the notifications to SSE I did the bare minimum

henryju (approved):

PR #497, opened 2022-02-21
SLI-676 Reduce the number of IO reactor dispatch threads

damien-urruty-sonarsource:


damien-urruty-sonarsource:
Closing as won't be merged. Was related to https://community.sonarsource.com/t/cpu-reaches-too-100-on-intellij-randomly/57710 and the problem was finally with the JDK

PR #500, opened 2022-03-01
Replace deprecated PermGen option with Metaspace (Java 8+)

jblievremont:


gradle.properties
```diff
@@ -1,5 +1,5 @@
 version=6.6-SNAPSHOT
-org.gradle.jvmargs=-XX:MaxPermSize=256M
+org.gradle.jvmargs=-XX:MaxMetaspaceSize=512M
```
jblievremont, line 2: NB: I tried both without this flag and with the 256M that was set previously, both led to out of memory exceptions on metaspace.

damien-urruty-sonarsource (approved):
:+1:

PR #510, opened 2022-03-18
SLI-693 Update embedded CFamily analyzer to 6.32.0.44918

mpaladin:


damien-urruty-sonarsource (approved):

mpaladin:
@henryju this version removes the use of legacy keys in CFamily plugin, is there any logic in `sonarlint-intellij` that should be modified?

damien-urruty-sonarsource:
@mpaladin I missed this change, it was maybe a bit prematurate to merge, my bad

damien-urruty-sonarsource:
I double checked in SLCORE and we should be safe :+1:

henryju:
You mean that thanks to the deprecatedRuleKeys attribute of rules definition, there will be a mapping of old -> new rule keys? Does it work also in connected mode?

PR #511, opened 2022-03-29
SLI-701 Subscribe for server events when the engine starts

damien-urruty-sonarsource:


henryju (approved):
I guess it is difficult, but have you considered adding a test to avoid future regressions?

damien-urruty-sonarsource:
I added tests. They should verify that listeners are correctly declared and do their job

damien-urruty-sonarsource:
@henryju merging this one so that it goes to dogfood, feel free to comment anyway and I can still make adjustments later

PR #518, opened 2022-05-27
Use unreleased SLCORE with debug traces for active rules

damien-urruty-sonarsource:


damien-urruty-sonarsource:
Closing as it was used for debug purposes

PR #522, opened 2022-06-20
SLI-736 Update embedded Ruby analyzer to 1.10.0.3710

alban-auzeill:


damien-urruty-sonarsource:
Change integrated in #527

PR #523, opened 2022-06-20
Automatic issue sync

henryju:


src/main/java/org/sonarlint/intellij/tasks/BindingStorageUpdateTask.java
```diff
@@ -149,40 +150,49 @@ public void doRun() {
...
+  private static void updateAllProjectIssuesForCurrentBranch(ConnectedSonarLintEngine engine, ServerConnection connection, TaskProgressMonitor monitor,
+    Collection<Project> projectsToUpdate) {
+    var allProjectAndBranchesToSync = new HashSet<ProjectBindingManager.ProjectKeyAndBranch>();
+    for (Project project : projectsToUpdate) {
+      ProjectBindingManager bindingManager = getService(project, ProjectBindingManager.class);
+      var projectAndBranchesToSync = bindingManager.getUniqueProjectKeysAndBranchesPairs();
+      allProjectAndBranchesToSync.addAll(projectAndBranchesToSync);
+    }
+    allProjectAndBranchesToSync
+      .forEach(pb -> engine.downloadAllServerIssues(connection.getEndpointParams(), connection.getHttpClient(), pb.getProjectKey(), pb.getBranchName(), monitor));
```
damien-urruty-sonarsource, line 190: Why not use `syncServerIssues` here ? You don't want to use the incremental update but reset the storage ?

damien-urruty-sonarsource (approved):
LGTM

build.gradle.kts
```diff
@@ -63,6 +63,7 @@ allprojects {
     }
 
     repositories {
+        mavenLocal()
```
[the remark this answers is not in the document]
henryju: Indeed, I will remove

src/main/java/org/sonarlint/intellij/tasks/BindingStorageUpdateTask.java
henryju: This is bit misleading, but `syncServerIssues` is a noop for older SQ versions and SC. Here we want to refresh issues for all SQ versions.

PR #524, opened 2022-06-23
Hardening Rider #1

henryju:
Depends on https://github.com/SonarSource/sonarlint-omnisharp/pull/118

### DIFF
--- a/src/main/java/org/sonarlint/intellij/issue/persistence/IssuePersistence.java
+++ b/src/main/java/org/sonarlint/intellij/issue/persistence/IssuePersistence.java
@@ -24,9 +24,13 @@ import com.intellij.openapi.project.Project;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Collection;
+import java.util.Objects;
 import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
+
+import com.intellij.project.ProjectKt;
 import org.sonarlint.intellij.issue.LiveIssue;
 import org.sonarlint.intellij.issue.LocalIssueTrackable;
 import org.sonarlint.intellij.proto.Sonarlint;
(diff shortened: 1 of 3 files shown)


src/main/java/org/sonarlint/intellij/issue/persistence/IssuePersistence.java
```diff
@@ -86,8 +90,8 @@ public synchronized Collection<LocalIssueTrackable> read(String key) throws IOEx
   }
 
   private Path getBasePath() {
-    var ideaDir = new File(myProject.getBasePath(), Project.DIRECTORY_STORE_FOLDER).toPath();
-    return ideaDir.resolve("sonarlint").resolve("issuestore");
+    var ideaDir = Objects.requireNonNull(ProjectKt.getStateStore(myProject).getDirectoryStorePath(true));
```
damien-urruty-sonarsource, line 93: I wonder about this `requireNonNull`, the `getDirectoryStorePath` returns a Nullable so it could happen, not sure when. Should we fallback on the previous behaviour if it's the case ?

damien-urruty-sonarsource (approved):

src/main/java/org/sonarlint/intellij/issue/persistence/IssuePersistence.java
damien-urruty-sonarsource: Browsing over the deprecated methods in the project I noticed several things:

1. the `getBasePath()` method we were using previously was deprecated so in case you want to fallback on old behaviour it might be good to find a replacement for the method.
2. this new `getDirectoryStorePath(boolean)` is also deprecated. See [here](https://github.com/JetBrains/intellij-community/blob/master/platform/projectModel-impl/src/com/intellij/openapi/components/impl/stores/IProjectStore.java#L45). The suggested replacement is the overload with no parameter.
3. The [javadoc](https://github.com/JetBrains/intellij-community/blob/master/platform/projectModel-impl/src/com/intellij/openapi/components/impl/stores/IProjectStore.java#L48) for this method says "The directory of project configuration files for directory-based project or null for file-based.". So I guess it returns null in the case of old .ipr project format. Where should we store the issues in this case ? I guess previously we were creating a `sonarlint` folder directly inside the project folder

henryju: I don't see `getBasePath` as deprecated... but it can return null in certain cases. We were already using it, so my plan is to fallback on it, since it was the old behavior.

damien-urruty-sonarsource: you're right, i confused with `getBaseDir`. It's not deprecated, although they recently updated the javadoc:
> It's strongly recommended to use other methods instead of this one (see {@link #getBaseDir()} for alternatives. Most probably any use of this method in production code may lead to unexpected results for some projects (e.g. if {@code .idea} directory is stored not near the project files).

PR #525, opened 2022-06-24
Hardening for 6.8

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/vcs/DefaultVcsService.kt
```diff
@@ -58,14 +58,20 @@ class DefaultVcsService @NonInjectable constructor(private val project: Project,
...
         }
         if (repositories.size > 1) {
             logger.warn("Several candidate Vcs repositories detected for module $module, cannot choose one")
             return null
         }
         val repo = repositories.first()
-        return repo.electBestMatchingServerBranchForCurrentHead(validConnectedEngine.getServerBranches(projectKey))
+        val resolvedBranch =
+            repo.electBestMatchingServerBranchForCurrentHead(validConnectedEngine.getServerBranches(projectKey))
+        if (resolvedBranch == null) {
```
henryju, line 71: I guess this change is not strictly related to the change of IJ api

henryju (approved):

src/main/java/org/sonarlint/intellij/vcs/DefaultVcsService.kt
damien-urruty-sonarsource: No it was added for debug purposes when a test was failing. I will remove it

PR #526, opened 2022-06-29
Hardening 6.8 - round 2

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/editor/SonarExternalAnnotator.java
```diff
@@ -196,8 +181,16 @@ private static HighlightSeverity getSeverity(@Nullable String severity) {
     return SonarLintSeverity.byName(severity).highlightSeverity();
   }
 
-  private static TextRange createTextRange(RangeMarker rangeMarker) {
-    return new TextRange(rangeMarker.getStartOffset(), rangeMarker.getEndOffset());
+  @CheckForNull
+  private static TextRange toValidTextRange(@Nullable RangeMarker rangeMarker) {
```
henryju, line 185: Is it used somewhere?

henryju (approved):
Minor comment then LGTM

src/main/java/org/sonarlint/intellij/editor/SonarExternalAnnotator.java
damien-urruty-sonarsource: Good catch! Forgot to remove it after a move

PR #542, opened 2022-09-20
Update test dependencies

henryju:


build.gradle.kts
```diff
@@ -14,26 +14,26 @@ import java.util.zip.ZipEntry
...
+    id("org.cyclonedx.bom") version "1.7.1"
 }
 
 buildscript {
     repositories {
         mavenCentral()
     }
     dependencies {
-        "classpath"(group = "org.jetbrains.intellij", name = "blockmap", version = "1.0.5")
+        "classpath"(group = "org.jetbrains.intellij", name = "blockmap", version = "1.0.6")
```
damien-urruty-sonarsource, line 36: I noticed the other day that we are building these blockmap files. I don't think we push them to the marketplace, do we still need to keep that in the build?

damien-urruty-sonarsource (approved):

build.gradle.kts
henryju: They are normally automatically computed by the marketplace. Here the goal was an attempt to have incremental update for our dogfood channel from repox. I have never been convinced it is actually working. We would have to spend some time on it.

PR #543, opened 2022-09-20
SLI-762 Update embedded C# analyzer to 8.45

henryju:


andrei-epure-sonarsource:
should this be dropped (not merged) based on the discussion in Jira?
https://community.sonarsource.com/t/8-45-0-54064-many-false-positives-with-xx-is-null-on-at-least-one-execution-path/71593

henryju:
Yes, we will likely wait for the next version of SonarC#.

PR #544, opened 2022-09-21
Update analyzers for 7.1

damien-urruty-sonarsource:


henryju (approved):
As discussed, let's disable YAML, and then LGTM

PR #546, opened 2022-09-22
SLI-763 Pull issues only when branch can be resolved

damien-urruty-sonarsource:


henryju (commented):
I have a doubt regarding `Stop synchronizing taint issues on file open` commit. Is the method only called when opening file?
Else LGTM

src/main/java/org/sonarlint/intellij/core/ConnectedModeStorageSynchronizer.java
```diff
@@ -132,6 +133,7 @@ public void run(@NotNull ProgressIndicator progressIndicator) {
             var serverConnection = projectBindingManager.getServerConnection();
             progressIndicator.setIndeterminate(false);
             syncIssuesForBranch(engine, serverConnection, moduleProjectKey, branchName, progressIndicator);
+            getService(myProject, TaintVulnerabilitiesPresenter.class).presentTaintVulnerabilitiesForOpenFiles();
```
henryju, line 136: Shouldn't you test if taint feature is enabled before calling this?

henryju (approved):
A question but the rest LGTM

PR #547, opened 2022-09-22
SLI-759 Rename 'Commit anyway' option to 'Continue' in pre-commit dialog

damien-urruty-sonarsource:
Also rename the 'Close' option to 'Cancel' as it better describes the behaviour


henryju (approved):
LGTM

PR #549, opened 2022-09-22
SLI-727 Collect architecture in telemetry

henryju:


src/main/java/org/sonarlint/intellij/telemetry/TelemetryManagerProvider.java
```diff
@@ -23,6 +23,8 @@
 import com.intellij.openapi.application.PathManager;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+
+import org.apache.commons.lang3.SystemUtils;
```
damien-urruty-sonarsource, line 27: Unused import ?

damien-urruty-sonarsource (approved):
There is another issue on Next. Else LGTM

PR #551, opened 2022-09-27
SLI-777 Fix SSF-317

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/issue/hotspot/SecurityHotspotShowRequestHandler.kt
```diff
@@ -50,6 +53,12 @@ open class SecurityHotspotShowRequestHandler(
 ) {
 
   open fun open(projectKey: String, hotspotKey: String, serverUrl: String) {
+      try {
+          URL(serverUrl)
+      } catch (e: Exception) {
+          GlobalLogOutput.get().log("Malicious open hotspot request received, the server URL is $serverUrl", ClientLogOutput.Level.INFO)
```
henryju, line 59: I would not log the `serverUrl` at all

henryju: I would replace "Malicious" by "Invalid"

henryju (changes requested):

PR #553, opened 2022-10-12
SLI-310 Highlight info issues as weak warnings

lbenedetto:
In 2018 @dbmeneses made the change "Highlight info issues as weak warnings" for SLI-310.
However, the change was incomplete as info issues are still highlighted identical to regular warnings. This is because the ProblemHighlightType must also be changed as well. 
image

This is what the highlight would look like with these new settings. I actually couldn't get sonarlint to build (failed to resolve dependencies), so this screenshot is from a different plugin that I wrote using the same highlighting settings.
image

This change is important as without it, there are lint issues that we will be forced to disable to prevent them from drowning out more important issues.

In SLI-309 it says "Users can always configure the way issues are shown in the settings." but I do not see how to do this, nor can I find any code implementing that feature.


henryju:
Hi @lbenedetto 

Thanks for your contribution. We have started to discuss your proposal with the team. We will update this PR once we know more.

> In [SLI-309](https://sonarsource.atlassian.net/browse/SLI-309) it says "Users can always configure the way issues are shown in the settings." but I do not see how to do this, nor can I find any code implementing that feature.

You can customize highlighting for each issue severity in IntelliJ settings:

image

nquinquenel:
This PR is quite outdated, and we changed the SonarLint severities some time ago. I'm closing the PR, but I invite you to create another one or reach out to us on our Community forum if there is still an issue.

PR #554, opened 2022-10-13
 SLI-789 Help users to generate a token when creating a new connection

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/server/RequestProcessor.kt
```diff
@@ -84,6 +91,26 @@ class RequestProcessor(
         }
         return Success()
     }
+
+    private fun processUserToken(request: Request): Response {
+        val token = try {
+            parseToken(request)
```
henryju, line 97: Is there any validation we should add here? Like a maximum size, to avoid vandalism?

henryju (approved):

src/main/java/org/sonarlint/intellij/server/RequestProcessor.kt
damien-urruty-sonarsource: As discussed, I did not add any limit, but I restrict the processing of the token to only when the user initiated a token creation request. This limits the window

henryju (changes requested):
If I understand the code correctly, the token listener is never unregistered, so it will never be garbage collected until another listener is registered to replace it.

damien-urruty-sonarsource:
@henryju it is unregistered in the [dispose method](https://github.com/SonarSource/sonarlint-intellij/pull/554/files#diff-55d960b950f07b674ae4deeb9372094e994172792529b09f6449914762e665dfR316) of the AuthStep wizard step

henryju (approved):

PR #557, opened 2022-10-17
Hardening

damien-urruty-sonarsource:


azure-pipelines.yml
```diff
@@ -124,7 +124,7 @@ stages:
       - task: Gradle@2
         inputs:
           options: --build-cache --stacktrace -i
-          tasks: check jacocoTestReport -x :its:check
+          tasks: jacocoTestReport -x :its:check
```
henryju, line 127: Is it because `jacocoTestReport` depends on `check` already?

henryju (approved):

azure-pipelines.yml
damien-urruty-sonarsource: I made the change in the build.gradle file to make `jacocoTestReport` depend on `check` [here](https://github.com/SonarSource/sonarlint-intellij/pull/557/files#diff-c0dfa6bc7a8685217f70a860145fbdf416d449eaff052fa28352c5cec1a98c06R305) in the same PR. I think it makes more sense

PR #559, opened 2022-10-27
DOC-274 Update Node.js requirements

damien-urruty-sonarsource:


damien-urruty-sonarsource:
When the dedicated docs site is ready, I would prefer to remove references to Node version in those files, and replace them with a link to the Requirements page on the docs site

henryju (approved):

PR #560, opened 2022-10-28
SLI-793 Rework manual binding UI

henryju:


src/main/java/org/sonarlint/intellij/config/project/SearchProjectKeyDialog.kt
```diff
@@ -0,0 +1,179 @@
...
+        val projectComparator = java.util.Comparator { o1: ServerProject, o2: ServerProject ->
+            val c1 = o1.name.compareTo(o2.name, ignoreCase = true)
+            if (c1 != 0) {
+                return@Comparator c1
+            }
+            o1.key.compareTo(o2.key, ignoreCase = true)
+        }
+        val sortedProjects = projectsByKey.values.stream()
+            .sorted(projectComparator)
+            .collect(Collectors.toList())
```
damien-urruty-sonarsource, line 121: I think you can replace those lines with:
`val sortedProjects = projectsByKey.values.sortedWith(compareBy({ it.name.toLowerCase() }, { it.key.toLowerCase() }))`

damien-urruty-sonarsource (approved):

PR #567, opened 2022-11-22
SLI-821 Rework HTML rule description display

henryju:
* convert some classes to Kotlin
* extract rule "header" (name, type, severity, key) as native components
* share as much code as possible between toolwindow and settings


src/main/java/org/sonarlint/intellij/ui/ruledescription/RuleHeaderPanel.kt
```diff
@@ -0,0 +1,82 @@
...
+        ruleTypeIcon.icon = SonarLintIcons.type(type.toString())
+        ruleTypeLabel.text = clean(type.toString())
+        ruleSeverityIcon.icon = SonarLintIcons.severity(severity.toString())
+        ruleSeverityLabel.text = clean(severity.toString())
+        ruleKeyLabel.text = ruleKey
+        revalidate()
+    }
+
+    fun showMessage(msg: String) {
+        clear();
```
damien-urruty-sonarsource, line 73: Nitpick: semicolon

damien-urruty-sonarsource (approved):
Only really minor comments, else LGTM

PR #570, opened 2022-11-28
SLI-823 Declare modules as configuration scopes in the backend

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -204,7 +267,14 @@ class BackendService @NonInjectable constructor(private val backend: SonarLintBa
     }
 
     companion object {
+        private var moduleCount = 1
+        internal val uniqueIdentifierForModules = hashMapOf<Module, String>()
         fun projectId(project: Project) = project.projectFilePath ?: "DEFAULT_PROJECT"
+
+        fun moduleId(module: Module): String {
+            // there is no reliable unique identifier for modules, but a module is represented by a single object
+            return uniqueIdentifierForModules.computeIfAbsent(module) { m -> m.name + "-" + moduleCount++ }
```
henryju, line 276: To me it seems a module name should be unique for a given project. Wouldn't it be better to have unique id of the form "projectId/moduleName"?

henryju (approved):

src/main/java/org/sonarlint/intellij/core/BackendService.kt
damien-urruty-sonarsource: As discussed, a module can be renamed by the user, so it means we would need a way to update the ID of a scope in the backend. The IJ platform does not remove/re-add a module when renaming but there is a `ModuleListener.modulesRenamed` method. We could then decide to remove/re-add the scope in the backend. I fear if those scopes contain heavy data like a symbol table we would need to recompute it just for a rename. I prefer to keep it that way.
Maybe at some point it should be the backend that generates a technical ID (e.g. a UUID) at creation time, a bit like what can be found in web app backends

PR #572, opened 2022-12-05
SLI-812 Return PSI elements only from background data providers

damien-urruty-sonarsource:


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=7.3-SNAPSHOT
+version=7.2.1-SNAPSHOT
```
henryju, line 1: Should be in a separate commit to ease cherry-pick

henryju (approved):
LGTM, but the PR should target a branch of the 7.2 tag, and then we should cherry-pick the commit (without the version change) in master

henryju (approved):

PR #580, opened 2023-01-17
Use released version of SLCORE 8.13

henryju:


src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -78,6 +79,7 @@ class BackendService @NonInjectable constructor(private val backend: SonarLintBa
             serverConnections.filter { !it.isSonarCloud }.map { toSonarQubeBackendConnection(it) }
         backend.initialize(
             InitializeParams(
+                HostInfoDto("FIXME"),
```
henryju, line 82: Should not be used because we set `manageLocalServer` to `false`

damien-urruty-sonarsource (approved):

PR #582, opened 2023-01-24
SLI-851 Add support for clang-cl

alejandro-alvarez-sonarsource:
This is blocked by CPP-2455. Once the CFamily plugin is released, as far as I can see this is the only required change to enable clang-cl in CLion.


henryju:
What is the status of this PR? Are you waiting for something on our side?

alejandro-alvarez-sonarsource:
It needs the CFamily analyzer to be updated to 6.42.

PR #583, opened 2023-01-31
SLI-861 Include security hotspots in analysis

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/actions/SonarLintToolWindow.java
```diff
@@ -59,19 +62,32 @@ public SonarLintToolWindow(Project project) {
   /**
    * Must run in EDT
    */
-  public void openAnalysisResults() {
+  public void openReportTab(AnalysisResult analysisResult) {
+    this.<ReportPanel>withTab(SonarLintToolWindowFactory.REPORT_TAB_TITLE, panel -> panel.updateIssues(analysisResult));
+  }
+  public void clearReportTab() {
```
nquinquenel, line 68:
```suggestion
  }
  
  public void clearReportTab() {
```

src/main/java/org/sonarlint/intellij/actions/ClearReportAction.java
```diff
@@ -22,12 +22,11 @@
...
-import org.jetbrains.annotations.Nullable;
-import org.sonarlint.intellij.common.util.SonarLintUtils;
-import org.sonarlint.intellij.issue.IssueStore;
 
-public class SonarClearAnalysisResultsAction extends AnAction {
-  public SonarClearAnalysisResultsAction(@Nullable String text, @Nullable String description, @Nullable Icon icon) {
+import static org.sonarlint.intellij.common.util.SonarLintUtils.getService;
+
+public class ClearReportAction extends AnAction {
+  public ClearReportAction(String text, String description, Icon icon) {
```
nquinquenel, line 29: We lost information of `@Nullable` on constructor parameters

src/main/java/org/sonarlint/intellij/core/SonarLintEngineFactory.java
```diff
@@ -53,6 +53,7 @@ ConnectedSonarLintEngine createEngine(String connectionId, boolean isSonarCloud)
     var configBuilder = isSonarCloud ? ConnectedGlobalConfiguration.sonarCloudBuilder() : ConnectedGlobalConfiguration.sonarQubeBuilder();
     configBuilder
       .addEnabledLanguages(EmbeddedPlugins.getEnabledLanguagesInConnectedMode().toArray(new Language[0]))
+      .enableHotspots()
```
nquinquenel, line 56: Should be enabled only in the case of `sonarQubeBuilder`?

src/main/java/org/sonarlint/intellij/finding/LiveFinding.java
```diff
@@ -173,7 +167,7 @@ public IssueSeverity getUserSeverity() {
 
   @Override
   public RuleType getType() {
-    return type;
+    return RuleType.SECURITY_HOTSPOT;
```
nquinquenel, line 170: Why do we force `RuleType` to be a `SECURITY_HOTSPOT`? `getType()` is also called for issues

src/main/java/org/sonarlint/intellij/finding/RawIssueAdapter.java
```diff
@@ -0,0 +1,154 @@
...
+          var locInputFile = loc.getInputFile();
+          if (textRange != null && locInputFile != null) {
+            var locPsiFile = toPsiFile(project, locInputFile.getClientObject());
+            var range = matcher.match(locPsiFile, textRange);
+            matchedLocations.add(resolvedLocation(locPsiFile.getVirtualFile(), range, loc.getMessage(), null));
+          }
+        } catch (TextRangeMatcher.NoMatchException e) {
+          // File content is likely to have changed during the analysis, should be fixed in next analysis
+          SonarLintConsole.get(project)
+            .debug("Failed to find secondary location of issue for file: '" + psiFile.getName() + "'. The location won't be displayed - " + e.getMessage());
```
nquinquenel, line 95:
```suggestion
            .debug("Failed to find secondary location of finding for file: '" + psiFile.getName() + "'. The location won't be displayed - " + e.getMessage());
```

src/main/java/org/sonarlint/intellij/core/SonarLintEngineFactory.java
nquinquenel, line 56: In the case of SonarCloud we should not enable hotspots for now

src/main/java/org/sonarlint/intellij/finding/persistence/FindingsManager.java
```diff
@@ -0,0 +1,247 @@
...
+        } catch (Exception e) {
+          SonarLintConsole.get(myProject).error("Cannot flush issues", e);
+        }
+      }
+    });
+  }
+
+  public void clearAllIssuesForAllFiles() {
+    liveIssueCache.clear();
+    myProject.getMessageBus().syncPublisher(IssueStoreListener.SONARLINT_ISSUE_STORE_TOPIC).allChanged();
```
nquinquenel, line 89: What about renaming IssueStoreListener to FindingStoreListener?

src/main/java/org/sonarlint/intellij/finding/persistence/FindingsManager.java
```diff
@@ -0,0 +1,247 @@
...
+  }
+
+  public void removeResolvedFindings(VirtualFile file, List<LiveIssue> issues, List<LiveSecurityHotspot> securityHotspots) {
+    liveIssueCache.removeFindings(file, issues);
+    liveSecurityHotspotCache.removeFindings(file, securityHotspots);
+    notifyFindingsChangedForFile(file);
+  }
+
+  public boolean neverAnalyzed(VirtualFile file) {
+    return liveIssueCache.getLive(file) == null;
```
nquinquenel, line 111: I think we should also check `&& liveSecurityHotspotCache.getLive(file) == null`

src/main/java/org/sonarlint/intellij/finding/tracking/Tracker.java
```diff
@@ -373,7 +373,7 @@ private static class ServerIssueSearchKey implements SearchKey {
     private final String serverIssueKey;
 
     ServerIssueSearchKey(Trackable trackable) {
-      serverIssueKey = trackable.getServerIssueKey();
+      serverIssueKey = trackable.getServerFindingKey();
```
nquinquenel, line 376: I don't think `serverIssueKey` makes sense if we're using Finding now, what about renaming inner class `ServerIssueSearchKey`  and enum `ServerIssueSearchKeyFactory`

src/main/java/org/sonarlint/intellij/core/SonarLintEngineFactory.java
damien-urruty-sonarsource: This `enableHotspots` flag means "enable if the server permits it", so anyway we check if it's SC [a bit later](https://github.com/SonarSource/sonarlint-core/blob/master/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/hotspot/HotspotApi.java#L59) down the calls. I'm tempted to leave it like this, the day we can get hotspots from SC we don't have to change SLI

src/main/java/org/sonarlint/intellij/finding/persistence/FindingsManager.java
damien-urruty-sonarsource: Yes. And I hope to kill this interface in a next refactoring :)

src/main/java/org/sonarlint/intellij/finding/persistence/FindingsManager.java
damien-urruty-sonarsource: Good catch! And I see we have a `neverAnalyzed` and `wasAnalyzed`, I will get rid of one of them

damien-urruty-sonarsource: In fact the meaning of both methods is slightly different so I will keep them both but fix the issue and rename one

src/main/java/org/sonarlint/intellij/finding/tracking/Tracker.java
damien-urruty-sonarsource: Yes

henryju (approved):
Didn't look in detail, but overall LGTM. I suggest merging in the common branch to not block the rest of the sprint

nquinquenel (approved):

PR #584, opened 2023-02-01
SLI-584 Add Security Hotspots in report tab list

nquinquenel:


src/main/java/org/sonarlint/intellij/actions/OpenSecurityHotspotInBrowserAction.kt
```diff
@@ -0,0 +1,75 @@
...
+
+  override fun actionPerformed(e: AnActionEvent) {
+    val project = e.project ?: return
+    val securityHotspot = e.getData(SECURITY_HOTSPOT_DATA_KEY)
+    val key = securityHotspot?.serverFindingKey ?: return
+    val localFile = securityHotspot.psiFile().virtualFile ?: return
+    val localFileModule = ModuleUtil.findModuleForFile(localFile, project) ?: return
+    val serverUrl = serverConnection(project)?.hostUrl ?: return
+    val projectKey = getService(localFileModule, ModuleBindingManager::class.java).resolveProjectKey() ?: return
+    BrowserUtil.browse(buildLink(serverUrl, projectKey, key))
```
damien-urruty-sonarsource, line 60: We could try to use the new backend here, I can help you do that

src/main/java/org/sonarlint/intellij/ui/AbstractFindingsPanel.java
```diff
@@ -0,0 +1,200 @@
...
+  protected final Project project;
+  protected SonarLintRulePanel rulePanel;
+  protected JBTabbedPane detailsTab;
+  protected Tree tree;
+  protected IssueTreeModelBuilder treeBuilder;
+  protected FlowsTree flowsTree;
+  protected FlowsTreeModelBuilder flowsTreeBuilder;
+  private ActionToolbar mainToolbar;
+  protected SecurityHotspotTreeModelBuilder shTreeBuilder;
+  protected Tree shTree;
```
damien-urruty-sonarsource, line 60: I don't think we use such a notation in our code bases, usually I try to avoid abbreviations for clarity

src/main/java/org/sonarlint/intellij/ui/AbstractFindingsPanel.java
```diff
@@ -0,0 +1,200 @@
...
+    // Rule panel
+    rulePanel = new SonarLintRulePanel(project);
+
+    detailsTab = new JBTabbedPane();
+    detailsTab.insertTab("Rule", null, rulePanel, "Details about the rule", RULE_TAB_INDEX);
+    detailsTab.insertTab("Locations", null, flowsPanel, "All locations involved in the finding", LOCATIONS_TAB_INDEX);
+  }
+
+  protected void findingTreeSelectionChanged(TreeSelectionEvent e) {
+    if (e.getSource() instanceof IssueTree) {
```
damien-urruty-sonarsource, line 116: You could register 2 different listeners instead of using instanceof. Or as the 2 branches share code in common maybe we could factorize using LiveFinding? Not sure about that

src/main/java/org/sonarlint/intellij/editor/EditorDecorator.kt
```diff
@@ -107,6 +108,10 @@ open class EditorDecorator(private val project: Project) {
     createHighlight(hotspot.primaryLocation.range, hotspot.message)?.let { updateHighlights(listOf(it)) }
   }
 
+  open fun highlight(securityHotspot: LiveSecurityHotspot) {
```
damien-urruty-sonarsource, line 111: as there is nothing specific to security hotspots regarding highligthing, we could probably have a single method that takes a LiveFinding

src/main/java/org/sonarlint/intellij/ui/AbstractFindingsPanel.java
nquinquenel: I factorized some code and split in 2 listeners

src/main/java/org/sonarlint/intellij/trigger/SonarLintCheckinHandler.java
```diff
@@ -33,6 +33,7 @@
 import com.intellij.util.PairConsumer;
 import com.intellij.util.ui.UIUtil;
 import java.awt.BorderLayout;
+import java.time.Instant;
```
henryju, line 36: What for?

nquinquenel: Leftover from rebase operations!

PR #586, opened 2023-02-06
Migrate tests to JUnit 5

damien-urruty-sonarsource:


src/test/java/org/sonarlint/intellij/AbstractSonarLintLightTests.java
```diff
@@ -60,8 +62,11 @@ protected final String getTestDataPath() {
     return Paths.get("src/test/testData/" + this.getClass().getSimpleName()).toAbsolutePath().toString();
   }
 
-  @Before
-  public final void init() {
+  @BeforeEach
+  final void beforeEachLightTest() throws Exception {
+    // dummy test name, not used
+    setName("test");
```
nquinquenel, line 68: What's the purpose of this?

src/test/java/org/sonarlint/intellij/core/ServerIssueUpdaterTests.java
```diff
@@ -52,10 +52,10 @@
 import static org.mockito.Mockito.verifyNoInteractions;
 import static org.mockito.Mockito.when;
 
-public class ServerIssueUpdaterTest extends AbstractSonarLintLightTests {
+class ServerIssueUpdaterTests extends AbstractSonarLintLightTests {
   private static final String SERVER_ID = "myServer";
-  public static final String PROJECT_KEY = "foo";
-  public static final ProjectBinding PROJECT_BINDING = new ProjectBinding(PROJECT_KEY, "", "");
+  static final String PROJECT_KEY = "foo";
+  static final ProjectBinding PROJECT_BINDING = new ProjectBinding(PROJECT_KEY, "", "");
```
nquinquenel, line 58: Better private?

henryju (approved):

src/test/java/org/sonarlint/intellij/AbstractSonarLintLightTests.java
damien-urruty-sonarsource: IntelliJ test framework is built around JUnit 3 & 4, to make it work with JUnit 5 I have to explicitly call `TestCase.setName(String)`.
Thanks for your comment, I changed this to provide the real test name instead of dummy, it will be cleaner.

src/test/java/org/sonarlint/intellij/core/ServerIssueUpdaterTests.java
damien-urruty-sonarsource: Yes, good catch!

nquinquenel (approved):

PR #588, opened 2023-02-07
SLI-863 Change minimal version of IntelliJ Platform to 2021.3 

henryju:


clion/build.gradle.kts
```diff
@@ -9,11 +9,16 @@ intellij {
...
+    testRuntimeOnly(libs.junit.engine)
+    // Needed for https://github.com/gradle/gradle/issues/22333
+    testRuntimeOnly(libs.junit.launcher)
+    testImplementation(libs.mockito.core)
 }
 
 tasks.test {
     useJUnitPlatform()
+    testLogging {
+        events("passed", "skipped", "failed")
```
damien-urruty-sonarsource, line 22: I didn't know about this property. Does it print a summary at the end of tests execution ? I looked in the CI logs and I didn't find any additional logs.

I would be in favor of activating such traces on all projects, I find the Gradle test task too silent about what was run

damien-urruty-sonarsource (approved):
LGTM

clion/build.gradle.kts
henryju: I tried to add this when investigating JUnit runtime error, but it was not useful, so I suggest to remove it for now. I agree that Gradle test task is too silent, but this is another topic ;)

PR #592, opened 2023-02-08
SLI-861 Fix tracking

damien-urruty-sonarsource:


nquinquenel (approved):
Issue not reproduced with this fix, LGTM :+1:

PR #593, opened 2023-02-09
SLI-846 Replace embedded server by the one from SLCORE

henryju:


src/main/java/org/sonarlint/intellij/util/ProgressUtils.kt
```diff
@@ -0,0 +1,55 @@
...
+                return future.get(100, TimeUnit.MILLISECONDS)
+            } catch (t: TimeoutException) {
+                continue
+            } catch (e: InterruptedException) {
+                throw ProcessCanceledException()
+            } catch (e: CancellationException) {
+                throw ProcessCanceledException()
+            } catch (e: ExecutionException) {
+                throw e.cause ?: e
+            }
```
nquinquenel, line 51:
```suggestion
            try {
                return future.get(100, TimeUnit.MILLISECONDS)
            } catch (e: Exception) {
                when (e) {
                    is TimeoutException -> continue
                    is InterruptedException, is CancellationException -> throw ProcessCanceledException()
                    is ExecutionException -> throw e.cause ?: e
                    else -> throw e
                }
            }
```
WDYT?

henryju: Not exactly equivalent, since all exceptions will be caught and rethrown:
https://discuss.kotlinlang.org/t/does-kotlin-have-multi-catch/486/6

nquinquenel: Good to know!

nquinquenel (approved):
LGTM

PR #594, opened 2023-02-09
 SLI-861 Refactor the streaming of issues to benefit security hotspots

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/analysis/FindingStreamer.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  private void triggerStreaming(AnalysisCallback analysisCallback) {
+    analysisCallback.onIntermediateResult(new AnalysisIntermediateResult(new LiveFindings(issuesPerFile, securityHotspotsPerFile)));
+  }
+
+  private static class Alarm {
+    private final Duration duration;
+    private final Runnable endRunnable;
+    private final ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
```
henryju, line 67: Does it mean each analysis will create a new execute service? Can't we share a common executor service for the entire application lifetime?

henryju (approved):
LGTM, with a minor doubt regarding the ExecutotService creation in `Alarm`

src/main/java/org/sonarlint/intellij/analysis/FindingStreamer.java
damien-urruty-sonarsource: Maybe yes. I will merge the PR as is if you don't have other feedback, and I will open another one

PR #595, opened 2023-02-13
SLI-856 Rework open Hotspot in IDE to select detected one

henryju:


nquinquenel (approved):
LGTM

src/main/java/org/sonarlint/intellij/actions/SonarLintToolWindow.java
```diff
@@ -217,22 +216,19 @@ public void showIssueLocations(LiveIssue liveIssue) {
...
 
-  public void show(LocalHotspot remoteHotspot) {
-    //TODO plug in when OPEN in IDE feature modified
-//    activeHotspot = liveHotspot;
-    if (getToolWindow() == null) {
-      // can happen if we try to show while the project is opening
-      // we will show the hotspot when the tool window is built
-      return;
+  public boolean trySelectSecurityHotspot(String securityHotspotKey) {
+    var foundHotspot = liveSecurityHotspots.stream().filter(sh -> securityHotspotKey.equals(sh.getServerFindingKey())).findFirst();
```
damien-urruty-sonarsource, line 220: I noticed this from Mithil's branch, I'm not sure we should keep a list of live findings in the tool window (there is a logic to merge etc, I think it's too much for a UI component). In your case you could rely on the `OnTheFlyFindingsHolder` from `ShowSecurityHotspotCallable`.

src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.java
```diff
@@ -117,6 +118,12 @@ public void analyzeFilesOnUserAction(Collection<VirtualFile> files, AnActionEven
     }
   }
 
+  public void analyzeFileAndTrySelectHotspot(VirtualFile file, String securityHotspotKey) {
+    AnalysisCallback callback = new ShowSecurityHotspotCallable(project, onTheFlyFindingsHolder, securityHotspotKey);
+    var task = new Analysis(project, List.of(file), TriggerType.OPEN_SECURITY_HOTSPOT, true, callback);
+    TaskRunnerKt.startBackgroundTask(project, ANALYSIS_TASK_TITLE, task::run);
```
damien-urruty-sonarsource, line 124: I would have expected to have a modal task. As the user explicitly request to open the hotspot in the IDE, I think it's better to show him that something is going on

damien-urruty-sonarsource (approved):
2 comments else LGTM

src/main/java/org/sonarlint/intellij/actions/SonarLintToolWindow.java
nquinquenel: This is something I removed on my refactoring, maybe I can take care of that on my rebase later?

henryju: OK, so I let you handle the change @nquinquenel

src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.java
henryju: I will change that in a later PR, I don't think it is blocker for the merge.

PR #596, opened 2023-02-13
SLI-853 Display security hotspots of opened files in "Security Hostpots" tab

nquinquenel:


src/main/java/org/sonarlint/intellij/actions/SonarLintToolWindow.java
```diff
@@ -105,7 +100,10 @@ public void openCurrentFileTab() {
   }
 
   public void openSecurityHotspotsTab() {
-    openTab(getSecurityHotspotContent());
+    var content = getSecurityHotspotContent();
+    if (content != null) {
```
henryju, line 104: In which case would this happen to be `null`? I am a bit worried about not being able to open the tab when requested.

src/main/java/org/sonarlint/intellij/ui/ReportPanel.java
```diff
@@ -171,7 +171,7 @@ private void updateOnSelect(LiveFinding liveFinding) {
     var moduleForFile = ProjectRootManager.getInstance(project).getFileIndex().getModuleForFile(liveFinding.psiFile().getVirtualFile());
     rulePanel.setRuleKey(moduleForFile, liveFinding.getRuleKey(), null);
     SonarLintUtils.getService(project, EditorDecorator.class).highlightFinding(liveFinding);
-    flowsTree.getEmptyText().setText("No finding selected");
+    flowsTree.getEmptyText().setText("Selected finding doesn't have flows");
```
henryju, line 174: What if no findings are selected?

src/main/java/org/sonarlint/intellij/ui/SonarLintHotspotsListPanel.java
```diff
@@ -19,99 +19,201 @@
...
+import static org.sonarlint.intellij.ui.SonarLintToolWindowFactory.createSplitter;
 
-public class SonarLintHotspotsListPanel {
-  private final JPanel mainPanel;
-  private final JBList<LiveSecurityHotspot> liveHotspotJBList;
-  private final Project project;
-  private final CardPanel cardPanel;
-  private LiveSecurityHotspot selectedHotspot;
-  private final SonarLintRulePanel sonarLintRulePanel;
+public class SonarLintHotspotsListPanel extends SimpleToolWindowPanel implements Disposable {
```
henryju, line 69: I guess we should rename the class since it won't be a list anymore?

src/main/java/org/sonarlint/intellij/ui/nodes/SummaryNode.java
```diff
@@ -81,6 +81,24 @@ public int insertFileNode(FileNode newChild, Comparator<FileNode> comparator) {
     return insertIdx;
   }
 
+  public int insertLiveSecurityHotspotNode(LiveSecurityHotspotNode newChild, Comparator<LiveSecurityHotspotNode> comparator) {
+    if (children == null) {
+      insert(newChild, 0);
+      return 0;
+    }
+
+    // keep the cast for Java 8 compat
```
henryju, line 90: We don't need Java 8 compat anymore

henryju (approved):
Some minor comments, but overall LGTM

src/main/java/org/sonarlint/intellij/ui/SonarLintHotspotsListPanel.java
damien-urruty-sonarsource: Also we could get rid of the `SonarLint` prefix ?

src/main/java/org/sonarlint/intellij/util/SonarLintActions.java
```diff
@@ -24,6 +24,7 @@
 import com.intellij.openapi.actionSystem.DefaultActionGroup;
 import com.intellij.serviceContainer.NonInjectable;
 import icons.SonarLintIcons;
+import org.sonarlint.intellij.actions.ClearReportAction;
```
damien-urruty-sonarsource, line 27: Unused import ?

src/main/java/org/sonarlint/intellij/ui/ReportPanel.java
nquinquenel: By default, it is initialized as `No finding selected`
If the user clicks on a non `SecurityHotspotNode` or `IssueNode` (meaning root node or file node) it will be cleared as `No finding selected` as well

src/main/java/org/sonarlint/intellij/actions/SonarLintToolWindow.java
nquinquenel: Right, this shouldn't be the case here, I will remove it

It is needed on the method called during the on the fly analysis, it seems it can be executed before the hotspot tab is rendered (at least it makes the tests fail)

src/main/java/org/sonarlint/intellij/ui/SonarLintHotspotsListPanel.java
nquinquenel: I removed SonarLint prefix

PR #597, opened 2023-02-14
Fix local tracking

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/finding/LiveFinding.java
```diff
@@ -64,6 +64,7 @@ public LiveFinding(Issue issue, PsiFile psiFile, @Nullable RangeMarker range, @N
     this.uid = UID_GEN.getAndIncrement();
     this.context = context;
     this.quickFixes = quickFixes;
+    this.creationDate = System.currentTimeMillis();
```
henryju, line 67: For an issue that is detected on the first analysis of a file, we keep the creation date to null. Is it still supported?

damien-urruty-sonarsource: No. Why do we need to keep it null ? The only difference today is when there is no creation date on an issue, we don't display "a few seconds ago" after the issue message. If we need to know that the issue is new, I suggest we use another field

src/test/java/org/sonarlint/intellij/mediumtests/SecurityHotspotsMediumTest.kt
```diff
@@ -22,81 +22,188 @@ package org.sonarlint.intellij.mediumtests
...
+        branchName: String = "master",
+    ) {
+        createStorage(activeRuleKey = serverSecurityHotspot.ruleKey, projectKey = serverSecurityHotspot.projectKey, branchName = branchName)
+        mockServer.addProtobufResponse(
+            "/api/hotspots/search.protobuf?projectKey=${serverSecurityHotspot.projectKey}&files=${serverSecurityHotspot.filePath}&branch=$branchName&ps=500&p=1",
+            Hotspots.SearchWsResponse.newBuilder().setPaging(Common.Paging.newBuilder().setTotal(1).build()).addHotspots(
+                Hotspots.SearchWsResponse.Hotspot.newBuilder().setKey(serverSecurityHotspot.key).setMessage(serverSecurityHotspot.message)
+                    .setRuleKey(serverSecurityHotspot.ruleKey).setStatus(serverSecurityHotspot.status)
+                    .setVulnerabilityProbability(serverSecurityHotspot.vulnerabilityProbability)
+                    .setCreationDate(serverSecurityHotspot.introductionDate).setComponent(serverSecurityHotspot.componentKey).setTextRange(
```
nquinquenel, line 156: should be `setIntroductionDate`

src/test/java/org/sonarlint/intellij/mediumtests/SecurityHotspotsMediumTest.kt
```diff
@@ -22,81 +22,188 @@ package org.sonarlint.intellij.mediumtests
...
+            ).addComponents(
+                Hotspots.Component.newBuilder().setKey(serverSecurityHotspot.componentKey).setPath(serverSecurityHotspot.filePath).build()
+            ).build()
+        )
     }
 
     private fun createStorage(
-        serverVersion: String? = "9.7",
-        activeRuleKey: String
+        serverVersion: String? = "99.9",
```
nquinquenel, line 169: 99.9 on purpose? or typo error for 9.9?

src/test/java/org/sonarlint/intellij/mediumtests/SecurityHotspotsMediumTest.kt
damien-urruty-sonarsource: Yes we could rename it on the core side.

src/test/java/org/sonarlint/intellij/mediumtests/SecurityHotspotsMediumTest.kt
damien-urruty-sonarsource: it's on purpose, I want the default value to be something high. When not providing the argument in the call from the test, it means you want to simulate a recent server version

PR #598, opened 2023-02-14
Rework rule description

henryju:


damien-urruty-sonarsource (approved):
Not fan of the name `**Smart**RuleDetailsLoader`, appart from that LGTM!

henryju:
@damien-urruty-sonarsource I am open to suggestions for the name of the SmartRuleDetailsLoader class. I was considering to use "LazyRuleDetailsLoader", but it is misleading IMO, as "lazy" is usually used for things that delay the computation as long as possible. What about `RuleDetailsMemoizer`, as it is closer to something like https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/concurrent/Memoizer.html?

src/main/java/org/sonarlint/intellij/ui/HotspotCellRenderer.java
```diff
@@ -36,16 +36,14 @@ class HotspotCellRenderer implements ListCellRenderer {
...
-
   @Override
   public Component getListCellRendererComponent(JList list, Object value, int index, boolean isSelected, boolean cellHasFocus) {
     var hotspot = (LiveSecurityHotspot) value;
     var layout = new FlowLayout(FlowLayout.LEFT);
     var panel = new JPanel(layout);
 
     var shIconLabel = new JBLabel();
-    var icon = SonarLintIcons.type(ICON_KEY + hotspot.getVulnerabilityProbability().name());
+    var icon = SonarLintIcons.hotspotTypeWithProbability(hotspot.getVulnerabilityProbability());
```
nquinquenel, line 46: I think this file is not used anymore and can be deleted

src/main/java/org/sonarlint/intellij/ui/SonarLintRulePanel.kt
```diff
@@ -215,29 +226,30 @@ class SonarLintRulePanel(private val project: Project) : JBLoadingPanel(BorderLa
...
+                <br>
+                At the moment, the status of a Security Hotspot can only be updated in SonarQube. 
+                """.trimIndent())
+            if (finding.serverFindingKey != null) {
+                val serverConnection =
+                    SonarLintUtils.getService(project, ProjectBindingManager::class.java).serverConnection
+                val projectKey = Settings.getSettingsFor(project).projectKey
+
+                htmlStringBuilder.append("""
+                Click <a href="${serverConnection.hostUrl}/security_hotspots?id=${projectKey}&hotspots=${finding.serverFindingKey}">here</a>
```
nquinquenel, line 244: Parameters are not encoded with UTF-8, not sure if that's really an issue nowadays

src/main/java/org/sonarlint/intellij/ui/ruledescription/RuleHeaderPanel.kt
```diff
@@ -24,28 +24,36 @@ import com.intellij.ui.JBColor
...
     private val ruleTypeLabel = JBLabel()
     private val ruleSeverityIcon = JBLabel()
     private val ruleSeverityLabel = JBLabel()
+    private val hotspotVulnerabilityLabel = JBLabel("Review Priority: ")
+    private val hotspotVulnerabilityValueLabel = JBLabel()
     private val ruleKeyLabel = JBLabel()
-    private val colorsByProbability: MutableMap<VulnerabilityProbability, JBColor> = EnumMap(VulnerabilityProbability::class.java)
+
+    // Same colors and opacity as in the icons SVG
+    private val highColor = Color(0x99e05555.toInt(), true)
```
nquinquenel, line 49: What about putting color hex values in the `SonarLintIcons` class since this is based on standard UI color from IntelliJ guidelines

src/main/java/org/sonarlint/intellij/ui/HotspotCellRenderer.java
henryju: Good catch!

src/main/java/org/sonarlint/intellij/ui/SonarLintRulePanel.kt
henryju: You mean escaped for URL compliance. You are right!

src/main/java/org/sonarlint/intellij/ui/ruledescription/RuleHeaderPanel.kt
henryju: Good idea!

src/main/java/org/sonarlint/intellij/ui/SonarLintRulePanel.kt
```diff
@@ -102,69 +99,83 @@ class SonarLintRulePanel(private val project: Project) : JBLoadingPanel(BorderLa
...
-        if (currentModule == module && currentRuleKey == ruleKey && currentContextKey == contextKey) {
-            return
-        }
-        currentRuleKey = ruleKey
-        currentModule = module
-        currentContextKey = contextKey
-        if (module == null || ruleKey == null) {
-            nothingToDisplay(false)
-            return
+    inner class SmartRuleDetailsLoader {
```
damien-urruty-sonarsource, line 106: Just removing Smart would work for me. But no strong pushback from me, I think it's not that important.
You could also group the 3 `last` fields in one and use a Kotlin data class, that would ease comparison.

henryju: Good idea to use a data class!

PR #604, opened 2023-02-16
Analyzers update for 8.0

damien-urruty-sonarsource:


src/main/resources/META-INF/plugin.xml
```diff
@@ -57,6 +57,7 @@
 
     <change-notes><![CDATA[
       <ul>
+        <li>8.0 - Support of JavaScript analysis in HTML files. Support clang-cl and Microchip compilers in CLion. New quick fix for a C++ rule. New quick fixes for 4 Python rules. 1 new rule for VB.NET in Rider. Bug fixes, fewer FPs and improvements for many languages.</li>
```
henryju, line 60: For clang-cl support, shouldn't this PR be merged first? https://github.com/SonarSource/sonarlint-intellij/pull/582

henryju: We don't support VB.NET in Rider...

henryju (changes requested):

src/main/resources/META-INF/plugin.xml
damien-urruty-sonarsource: For 1. yes, I was about to merge it after this one. But I will move the release notes to the other PR
For 2. OK I learned something :smile:

henryju (approved):
I would change the last commit message (and JIRA ticket) to not mention VB.NET, else LGTM

PR #605, opened 2023-02-16
Minor fixes

henryju:


src/main/java/org/sonarlint/intellij/ui/SonarLintRulePanel.kt
```diff
@@ -77,28 +75,33 @@ class SonarLintRulePanel(private val project: Project) : JBLoadingPanel(BorderLa
...
                 editorKit = UIUtil.getHTMLEditorKit()
-                border = JBUI.Borders.empty(10)
+                border = JBUI.Borders.empty()
                 isEditable = false
                 isOpaque = false
+                isFocusable = false
 
                 addHyperlinkListener(BrowserHyperlinkListener.INSTANCE)
-            }, BorderLayout.PAGE_END)
+            }, BorderLayout.SOUTH)
```
nquinquenel, line 97: Based on [Oracle documentation](https://docs.oracle.com/javase/tutorial/uiswing/layout/border.html), we should keep usage of `PAGE_END` IMO

`Before JDK release 1.4, the preferred names for the various areas were different, ranging from points of the compass (for example, BorderLayout.NORTH for the top area) to wordier versions of the constants we use in our examples. The constants our examples use are preferred because they are standard and enable programs to adjust to languages that have different orientations.`

src/main/java/org/sonarlint/intellij/ui/SonarLintRulePanel.kt
```diff
@@ -77,28 +75,33 @@ class SonarLintRulePanel(private val project: Project) : JBLoadingPanel(BorderLa
...
+    private var finding: Finding? = null
+    private var ruleDetails: ActiveRuleDetailsDto? = null
 
     init {
         add(JBPanel<JBPanel<*>>(BorderLayout()).apply {
-            ruleNameLabel.font = UIUtil.getLabelFont().deriveFont((UIUtil.getLabelFont().size2D + JBUIScale.scale(3))).deriveFont(Font.BOLD)
-            add(ruleNameLabel, BorderLayout.NORTH)
+            add(ruleNameLabel.apply {
+                font = UIUtil.getLabelFont().deriveFont((UIUtil.getLabelFont().size2D + JBUIScale.scale(3))).deriveFont(Font.BOLD)
+            }, BorderLayout.NORTH)
```
nquinquenel, line 85: Same here, we should change `NORTH` to `PAGE_START` IMO (cf below comment)

src/main/java/org/sonarlint/intellij/ui/SonarLintRulePanel.kt
```diff
@@ -77,28 +75,33 @@ class SonarLintRulePanel(private val project: Project) : JBLoadingPanel(BorderLa
...
+            }, BorderLayout.SOUTH)
         }, BorderLayout.NORTH)
 
         add(descriptionPanel, BorderLayout.CENTER)
 
-        paramsPanel.border = JBUI.Borders.emptyLeft(12)
-        add(paramsPanel, BorderLayout.SOUTH)
+        add(paramsPanel.apply {
+            border = IdeBorderFactory.createTitledBorder("Parameters")
+        }, BorderLayout.SOUTH)
```
nquinquenel, line 104: Switch to `PAGE_END` here as well IMO

src/main/java/org/sonarlint/intellij/finding/LiveFinding.java
```diff
@@ -64,6 +67,7 @@ protected LiveFinding(Issue issue, PsiFile psiFile, @Nullable RangeMarker range,
     this.uid = UID_GEN.getAndIncrement();
     this.context = context;
     this.quickFixes = quickFixes;
+    this.ruleDescriptionContextKey = issue.getRuleDescriptionContextKey().orElse(null);
```
damien-urruty-sonarsource, line 70: Is this something that I removed by mistake at some point ? Or it was not supported ?

src/main/java/org/sonarlint/intellij/finding/LiveFinding.java
```diff
@@ -203,4 +208,10 @@ public Optional<FindingContext> context() {
   public List<QuickFix> quickFixes() {
     return quickFixes;
   }
+
+  @org.jetbrains.annotations.Nullable
```
damien-urruty-sonarsource, line 212: Don't we use javax annotations usually ?

damien-urruty-sonarsource (approved):

src/main/java/org/sonarlint/intellij/ui/SonarLintRulePanel.kt
henryju: Interesting, but I am not sure this is what we want. In the case of languages with inverted orientation, I am not sure we want to have our layout inverted.
See https://stackoverflow.com/a/62894110/534773
I suggest keeping using NORTH/SOUTH EAST/WEST at the moment.

src/main/java/org/sonarlint/intellij/finding/LiveFinding.java
henryju: I don't think it was supported. We were passing null as a contextKey for non taint issues. But since I was trying to make the rule description panel simpler, I did this change to avoid having two cases.

src/main/java/org/sonarlint/intellij/finding/LiveFinding.java
henryju: This is not very important I guess, since annotations are not loaded at runtime. I did this change to silence an InteliiJ inspection, because of the Kotlin -> Java conversion. As we will continue to migrate the codebase to Kotlin, this will disappear.

PR #607, opened 2023-02-17
SLI-857 Add filtering on security hotspots

nquinquenel:


src/main/java/org/sonarlint/intellij/ui/tree/SecurityHotspotTreeModelBuilder.java
```diff
@@ -213,6 +217,57 @@ private void setRootSecurityHotspots(VirtualFile file, Iterable<LiveSecurityHots
...
+    for (var securityHotspotNode : nonFilteredNodes) {
+      var idx = summary.insertLiveSecurityHotspotNode(securityHotspotNode, new LiveSecurityHotspotNodeComparator());
+      var newIdx = new int[]{idx};
+      model.nodesWereInserted(summary, newIdx);
+      model.nodeChanged(summary);
+    }
+
+    model.reload();
+  }
+
```
damien-urruty-sonarsource, line 270: There is a bunch of duplication in those 3 show methods. What about using the `SecurityHotspotFilters` enum and introduce a `shouldInclude` method ? This way you just pass the filter to this class and you have a single show method

src/main/java/org/sonarlint/intellij/actions/filters/SecurityHotspotFilters.java
```diff
@@ -0,0 +1,19 @@
+package org.sonarlint.intellij.actions.filters;
+
+public enum SecurityHotspotFilters {
+
+  SHOW_ALL("Show all"),
+  LOCAL_ONLY("Local only"),
+  MATCHED_ON_SONARQUBE("Matched on SonarQube");
```
damien-urruty-sonarsource, line 7: Matched with SQ instead of on ? Not sure

src/main/java/org/sonarlint/intellij/ui/tree/SecurityHotspotTreeModelBuilder.java
nquinquenel: You're right, way better :+1:

src/main/java/org/sonarlint/intellij/actions/filters/SecurityHotspotFilters.java
nquinquenel: I've changed to `Existing on SonarQube` as we discussed on MMF specification

damien-urruty-sonarsource (approved):

PR #614, opened 2023-02-28
SLI-886 Fix application of quick fixes in injected languages

henryju:
I would like to add a test, but the Python analyzer with quick fixes on regexp is not yet released.


nquinquenel (approved):
LGTM

PR #616, opened 2023-03-02
SLI-888 Check if project not disposed when updating view with findings

nquinquenel:


damien-urruty-sonarsource (approved):
LGTM! I think we should generalize this as a utility method that we could call everywhere when invokeLater needs to be called in a project's context

PR #618, opened 2023-03-13
Hardening for 8.1 #1

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/ui/SonarLintRulePanel.kt
```diff
@@ -233,6 +232,7 @@ class SonarLintRulePanel(private val project: Project) : JBLoadingPanel(BorderLa
 
     private fun updateHeader(finding: Finding, ruleDescription: ActiveRuleDetailsDto) {
         ruleNameLabel.text = ruleDescription.name
+        ruleNameLabel.setCopyable(true)
```
nquinquenel, line 236: Not only the rule name but also the severity, finding type, rule key, etc., were copyable (from the header).

nquinquenel (approved):
LGTM!

PR #629, opened 2023-03-31
SLI-901 Let core register smart notifications

nquinquenel:


src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -177,6 +174,15 @@ object SonarLintIntelliJClient : SonarLintClient {
         showBalloon(null, params.text, convert(params.type))
     }
 
+    override fun showSmartNotification(params: ShowSmartNotificationParams) {
+        val projects = params.scopeIds.stream()
+            .map { BackendService.findModule(it)?.project ?: BackendService.findProject(it) }
+            .filter { it != null}
```
damien-urruty-sonarsource, line 180: You can combine map and filter with `mapNotNull`

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -177,6 +174,15 @@ object SonarLintIntelliJClient : SonarLintClient {
...
+            .collect(Collectors.toSet())
```
damien-urruty-sonarsource, line 181: I think you can replace that by `toSet()`

src/main/java/org/sonarlint/intellij/notifications/SonarLintProjectNotifications.java
```diff
@@ -133,4 +140,34 @@ public void expireCurrentHotspotNotificationIfNeeded() {
       currentOpenHotspotNotification = null;
     }
   }
+
+  public void handle(ShowSmartNotificationParams smartNotificationParams) {
+    var telemetry = getService(SonarLintTelemetry.class);
+    var category = smartNotificationParams.getCategory();
+    telemetry.devNotificationsReceived(category);
```
damien-urruty-sonarsource, line 147: Isn't this managed by the backend already ?

damien-urruty-sonarsource (approved):
A few comments, else LGTM

src/main/java/org/sonarlint/intellij/notifications/SonarLintProjectNotifications.java
nquinquenel: Yes you're right, only dev notifications clicked should be counted on client side

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
nquinquenel: Thanks, I didn't know!

damien-urruty-sonarsource (approved):
LGTM

PR #634, opened 2023-04-19
SLI-910 Follow the current theme for code examples

damien-urruty-sonarsource:


nquinquenel (approved):
Nice, I forgot to fix that! When changing the theme, for example, it was not updating the editor correctly, with your PR it works well :+1:

PR #638, opened 2023-04-26
 SLI-839 Run actions update in a background thread 

damien-urruty-sonarsource:


damien-urruty-sonarsource:
@henryju you can maybe review commit by commit. My idea was to make sure all of our actions have a common ancestor, but I had to keep 3 different ones: one for classic actions, one for ToggleActions and one for Checkbox Actions. But then the real fix was much easier, because I only had to modify those 3 places

src/main/java/org/sonarlint/intellij/actions/AbstractSonarAction.java
```diff
@@ -21,40 +21,43 @@
...
+import com.intellij.openapi.project.DumbAware;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.vfs.VirtualFile;
 import java.util.stream.Stream;
 import javax.swing.Icon;
 import org.jetbrains.annotations.Nullable;
 import org.sonarlint.intellij.analysis.AnalysisStatus;
 
-public abstract class AbstractSonarAction extends AnAction {
+public abstract class AbstractSonarAction extends AnAction implements DumbAware {
```
henryju: Have you anticipated the impacts of implementing `DumbAware`? Are we sure all or actions can work properly in Dumb mode?

henryju (approved):
Minor doubt about the impact of DumAware on all actions, but overall this is a very good move, well done.

src/main/java/org/sonarlint/intellij/actions/AbstractSonarAction.java
damien-urruty-sonarsource: I tried the actions that are available with Shift-Shift when the project is in dumb mode. For all other actions they are available through the tool window and during indexing it is not shown. So I think it should be safe

PR #640, opened 2023-04-28
Update Analyzers

eray-felek-sonarsource:


henryju (approved):
LGTM, you can merge when QA is green

PR #644, opened 2023-05-05
SLI-889 Fix security hotspots not loading on startup

nquinquenel:


src/main/java/org/sonarlint/intellij/StartServicesOnProjectOpened.java
```diff
@@ -42,14 +42,12 @@ public void runActivity(@NotNull Project project) {
       return;
     }
     getService(project, EditorChangeTrigger.class).onProjectOpened();
+    getService(BackendService.class).projectOpened(project);
```
nquinquenel, line 45: By moving this before `SecurityHotspotsRefreshTrigger` gets subscribed, we are able to determine that SH are supported - otherwise, during the first SH refresh, we cannot find the effective binding and the status returned is `NOT_SUPPORTED`

damien-urruty-sonarsource (approved):
Good job :+1:

PR #647, opened 2023-05-10
SLI-934 Fix hotspot filter detection and remove hotspot after review

nquinquenel:


src/main/java/org/sonarlint/intellij/actions/filters/SecurityHotspotFilters.java
```diff
@@ -37,6 +37,10 @@ public String getTitle() {
     return title;
   }
 
+  public static SecurityHotspotFilters getDefaultFilter() {
+    return SHOW_ALL;
```
damien-urruty-sonarsource, line 41: Couldn't it be a static final field instead of a method ?

damien-urruty-sonarsource (approved):
Small comment, else LGTM!

PR #649, opened 2023-05-12
SLI-946 Honor POST requests after redirects

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/http/ApacheHttpClient.kt
```diff
@@ -240,3 +248,19 @@ class ApacheHttpClient private constructor(
...
+            // trick the response code so that POST is kept after redirect
+            response.code = HttpStatus.SC_PERMANENT_REDIRECT
+        }
+    }
+
+    private fun willPostBeChangedToGetOnRedirect(context: HttpContext, response: HttpResponse): Boolean {
+        val request = context.getAttribute(HttpCoreContext.HTTP_REQUEST) as HttpRequest?
+        val isPost = request != null && Method.POST.isSame(request.method)
+        val code = response.code
+        return isPost && (code == HttpStatus.SC_MOVED_PERMANENTLY || code == HttpStatus.SC_MOVED_TEMPORARILY || code == HttpStatus.SC_SEE_OTHER);
```
nquinquenel, line 264: I'm trying to understand and from what I read, shouldn't we map 301 to 308 and 302 to 307?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307
`The only difference between 307 and 302is that 307 guarantees that the method and the body will not be changed when the redirected request is made.`

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308
`The request method and the body will not be altered, whereas 301 may incorrectly sometimes be changed to a GET method.`

damien-urruty-sonarsource: yes, it's probably more precise

nquinquenel (approved):
LGTM!

PR #652, opened 2023-05-16
SLI-950 Hide resolved hotspots from report, fix selection, fix highlight

nquinquenel:
- Hide resolved security hotspots from the report tab
- Fix the status not being selected rightfully when updating from the report tab
- Only highlight security hotspot that are currently displayed in the security hotspot tab list
- Fix filtering when status is not supported
- Trigger highlighting on filter change

### DIFF
--- a/src/main/java/org/sonarlint/intellij/editor/SonarExternalAnnotator.java
+++ b/src/main/java/org/sonarlint/intellij/editor/SonarExternalAnnotator.java
@@ -77,17 +77,15 @@ public class SonarExternalAnnotator extends ExternalAnnotator<SonarExternalAnnot
       });
 
     var toolWindowService = getService(project, SonarLintToolWindow.class);
-    if (toolWindowService.isSecurityHotspotsTabActive()) {
-      var securityHotspots = issueManager.getSecurityHotspotsForFile(file.getVirtualFile());
-      securityHotspots
-        .forEach(securityHotspot -> {
-          // reject ranges that are no longer valid. It probably means that they were deleted from the file, or the file was deleted
-          var validTextRange = securityHotspot.getValidTextRange();
-          if (validTextRange != null) {
-            addAnnotation(project, securityHotspot, validTextRange, holder);
-          }
-        });
-    }
+    var securityHotspots = issueManager.getSecurityHotspotsForFile(file.getVirtualFile());
+    toolWindowService.getDisplayedSecurityHotspots(securityHotspots)
+      .forEach(securityHotspot -> {
+        // reject ranges that are no longer valid. It probably means that they were deleted from the file, or the file was deleted
+        var validTextRange = securityHotspot.getValidTextRange();
+        if (validTextRange != null) {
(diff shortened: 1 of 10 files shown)


src/main/java/org/sonarlint/intellij/editor/SonarExternalAnnotator.java
```diff
@@ -77,17 +77,16 @@ public void apply(@NotNull PsiFile file, AnnotationContext annotationResult, @No
...
-        .forEach(securityHotspot -> {
-          // reject ranges that are no longer valid. It probably means that they were deleted from the file, or the file was deleted
-          var validTextRange = securityHotspot.getValidTextRange();
-          if (validTextRange != null) {
-            addAnnotation(project, securityHotspot, validTextRange, holder);
-          }
-        });
-    }
+    var securityHotspots = issueManager.getSecurityHotspotsForFile(file.getVirtualFile());
+    var displayedSecurityHotspots = toolWindowService.getDisplayedSecurityHotspots(securityHotspots);
```
damien-urruty-sonarsource, line 81: In fact I'm not sure we should annotate security hotspots that come from the Report tab. The report is a snapshot, and it could happen that the code changes after the report has been generated, so in this case we could end up annotating the wrong code. Have you checked this with Marco ? I would be in favor of not annotating hotspots that come from the report tab at all

src/main/java/org/sonarlint/intellij/ui/nodes/SummaryNode.java
```diff
@@ -97,6 +99,28 @@ public int insertLiveSecurityHotspotNode(LiveSecurityHotspotNode newChild, Compa
...
+      while(children.hasMoreElements()) {
+        var hotspotNode = (LiveSecurityHotspotNode) children.nextElement();
+        var hotspot = hotspotNode.getHotspot();
+        if (securityHotspotKey.equals(hotspot.getServerFindingKey())) {
+          hotspot.setStatus(status);
+          if (hotspot.isResolved()) {
+            fileNode.remove(hotspotNode);
+            if (fileNode.getFindingCount() == 0) {
+              index.remove(fileNode.file());
+              remove(fileNode);
```
damien-urruty-sonarsource, line 115: I am surprised that we need to look for the node by going down in the hierarchy and remove it ourselves, I would have expected other ways to achieve that using the model, but maybe I'm wrong

src/main/java/org/sonarlint/intellij/editor/SonarExternalAnnotator.java
nquinquenel: Removed after further discussions :+1:

src/main/java/org/sonarlint/intellij/ui/nodes/SummaryNode.java
nquinquenel: Improved

damien-urruty-sonarsource (approved):
This looks simpler!

PR #654, opened 2023-05-17
SLI-948 Support security hotspots when connected to SonarCloud

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/ui/SonarLintRulePanel.kt
```diff
@@ -297,12 +292,39 @@ class SonarLintRulePanel(private val project: Project, private val parent: Dispo
...
+            urlEncode(
+                serverFindingKey
+            )
+        }"
+    }
+
+    private fun securityHotspotsDocLink(serverConnection: ServerConnection) = externalLink(
+        "Security Hotspot",
+        if (serverConnection.isSonarCloud) "https://docs.sonarcloud.io/digging-deeper/security-hotspots/"
+        else "https://docs.sonarqube.org/latest/user-guide/security-hotspots/"
```
nquinquenel, line 325: Random idea but would it make sense to centralize all these static URLs into a specific class? In case we might want to review them in the future to see which one or deprecated or not

nquinquenel (approved):
LGTM

src/main/java/org/sonarlint/intellij/ui/nodes/LiveSecurityHotspotNode.java
```diff
@@ -68,6 +78,19 @@ public void render(TreeCellRenderer renderer) {
     }
   }
 
+  @NotNull
+  private static Icon[] getIcons(Icon typeIcon, Optional<ServerConnection> serverConnection) {
```
nquinquenel, line 82: Code smell: avoid usage of optional as paremeter

src/main/java/org/sonarlint/intellij/ui/SonarLintRulePanel.kt
damien-urruty-sonarsource: good point. I will treat it separately

src/main/java/org/sonarlint/intellij/ui/nodes/LiveSecurityHotspotNode.java
damien-urruty-sonarsource: Fixed

nquinquenel (approved):

nquinquenel (approved):

nquinquenel (approved):
:rocket:

PR #655, opened 2023-05-17
SLI-921: Theme changes in rule description

thahnen:
## Summary

When changing the theme (Look And Feel) the rule description window refreshes to show the content correctly. This will also work with every other change to the UI (e.g. accessability).

## Testing

Testing was done manually, on macOS there is a slight "lag", so please test it on Linux to check if it's only a macOS thing ^^

### DIFF
--- a/src/main/java/org/sonarlint/intellij/ui/AbstractIssuesPanel.java
+++ b/src/main/java/org/sonarlint/intellij/ui/AbstractIssuesPanel.java
@@ -20,11 +20,14 @@
 package org.sonarlint.intellij.ui;
 
 import com.intellij.ide.OccurenceNavigator;
+import com.intellij.ide.ui.LafManager;
+import com.intellij.ide.ui.LafManagerListener;
 import com.intellij.openapi.Disposable;
 import com.intellij.openapi.actionSystem.ActionGroup;
 import com.intellij.openapi.actionSystem.ActionManager;
 import com.intellij.openapi.actionSystem.ActionToolbar;
 import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.fileEditor.OpenFileDescriptor;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.roots.ProjectRootManager;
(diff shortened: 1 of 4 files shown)


src/main/java/org/sonarlint/intellij/ui/AbstractIssuesPanel.java
```diff
@@ -94,7 +96,17 @@ private void createTabs() {
     detailsTab.insertTab("Locations", null, flowsPanel, "All locations involved in the issue", LOCATIONS_TAB_INDEX);
   }
 
+  /**
+   *  To handle UI theme changes the rule windows must be reloaded immediately, otherwise the accessibility isn't great
+   *  (bad contrast, ...). We want to subscribe to the message bus everytime the rule panel content changes to reload
+   *  the panel with the correct rule content.
+   */
   protected void issueTreeSelectionChanged() {
+    ApplicationManager.getApplication().getMessageBus().connect().subscribe(LafManagerListener.TOPIC, ignored -> actualIssueTreeSelectionChanged());
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: I just tested with the following steps:

1. analyze a project
2. go to the Report tab and navigate between files and issues in the tree
3. change the theme in the settings

The LaF callbacks were called as many times as I changed the selection in step 2. And in fact it is cumulated across the different tabs. I think this is not good for several reasons:

* resource waste. We might end up doing the same thing several times
* some of those callbacks pass the event that was received on the selection change. I don't think there is a guarantee on the order on which the messages will be delivered from the bus, so you could end up applying an outdated event
* to my knowledge the only part needing a refresh when the theme changes is the rule description, so refreshing the selection on the trees and lists does not seem very effective.

What about listening to the theme changes directly inside the component responsible for the rule description ? I would try with subscribing to the bus from `SonarLintRulePanel` and call `updateUiComponents` when the theme changes.

src/main/java/org/sonarlint/intellij/ui/SonarLintRulePanel.kt
```diff
@@ -126,6 +127,8 @@ class SonarLintRulePanel(private val project: Project, private val parent: Dispo
 
         setLoadingText("Loading rule description...")
         clear()
+
+        ApplicationManager.getApplication().messageBus.connect().subscribe(LafManagerListener.TOPIC, LafManagerListener { updateUiComponents() })
```
damien-urruty-sonarsource, line 131: The last small improvement that you can do here is to connect to the bus by passing the `parent` disposable existing in this class. The listener will have a shorter lifecycle

thahnen: Done, there ya go mate 😄

damien-urruty-sonarsource (approved):
Thanks for the rework! It looks very simple now

PR #656, opened 2023-05-17
SLI-833: Remove use of internal API

thahnen:
Cleanup / changes in the codebase based on the [Internal API Migration](https://plugins.jetbrains.com/docs/intellij/api-internal.html#plugins).

### DIFF
--- a/src/main/java/org/sonarlint/intellij/fs/VirtualFileSystemListener.kt
+++ b/src/main/java/org/sonarlint/intellij/fs/VirtualFileSystemListener.kt
@@ -33,8 +33,7 @@ import com.intellij.openapi.vfs.newvfs.events.VFileDeleteEvent
 import com.intellij.openapi.vfs.newvfs.events.VFileEvent
 import com.intellij.openapi.vfs.newvfs.events.VFileMoveEvent
 import com.intellij.openapi.vfs.newvfs.events.VFilePropertyChangeEvent
-import com.intellij.util.PlatformUtils
-import org.sonarlint.intellij.common.util.SonarLintUtils.getService
+import org.sonarlint.intellij.common.util.SonarLintUtils
 import org.sonarlint.intellij.core.ProjectBindingManager
 import org.sonarlint.intellij.util.GlobalLogOutput
 import org.sonarsource.sonarlint.core.analysis.api.ClientModuleFileEvent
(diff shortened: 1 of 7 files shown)


nquinquenel (approved):
LGTM!

src/main/java/org/sonarlint/intellij/fs/VirtualFileSystemListener.kt
```diff
@@ -33,8 +33,7 @@ import com.intellij.openapi.vfs.newvfs.events.VFileDeleteEvent
 import com.intellij.openapi.vfs.newvfs.events.VFileEvent
 import com.intellij.openapi.vfs.newvfs.events.VFileMoveEvent
 import com.intellij.openapi.vfs.newvfs.events.VFilePropertyChangeEvent
-import com.intellij.util.PlatformUtils
-import org.sonarlint.intellij.common.util.SonarLintUtils.getService
+import org.sonarlint.intellij.common.util.SonarLintUtils
```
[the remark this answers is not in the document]
thahnen: Done 👍

src/main/java/org/sonarlint/intellij/util/SonarLintAppUtils.java
```diff
@@ -128,16 +129,18 @@ private static String getPathRelativeToProjectBaseDir(Project project, VirtualFi
...
-      return null;
-    }
-    var baseDir = Paths.get(moduleFilePath).getParent();
+    var moduleContentRoots = Arrays.stream(ModuleRootManager.getInstance(module).getContentRoots())
+            .filter(contentRoot -> contentRoot.getPath().trim().length() > 0)
+            .toArray(VirtualFile[]::new);
+
+    // If there are multiple content roots (this can be possible based on the SDK API documentation), just take the
+    // first one, assuming this one is the path of the ".iml" file.
+    var baseDir = Paths.get(moduleContentRoots[0].getPath()).getParent();
```
damien-urruty-sonarsource, line 143: I have a doubt about this:
* why stopping at the first content root ? The `file` parameter could be located in another root
* I don't think we need to still call `getParent()` now that we get the exact content root

This part is not covered with tests today, so we could take the opportunity to cover the code a bit better, e.g. with light/heavy tests to verify the different scenarios

thahnen: Very good point, I'll reopen the ticket. I was assuming a bit too much, that it would be handled by tests and actually didn't check if there even were corresponding tests.

PR #658, opened 2023-05-19
SLI-920: Move to new light service annotations

thahnen:
## Summary

The Jetbrains IntelliJ SDK provides a new method to mark light services other than configuring them in the `plugin.xml`. This way it's all done programmatically via the `@Service(...)` annotation.

## Testing

Testing was done locally but connected mode wasn't fully tested yet.


src/main/resources/META-INF/plugin.xml
```diff
@@ -188,67 +188,68 @@
...
+
+        <!-- Cannot be moved to service annotation because of 'serviceInterface' / 'testServiceImplementation' -->
         <applicationService serviceInterface="org.sonarlint.intellij.util.GlobalLogOutput"
                             serviceImplementation="org.sonarlint.intellij.util.GlobalLogOutputImpl"
                             testServiceImplementation="org.sonarlint.intellij.util.GlobalLogOutputTestImpl"/>
+
         <applicationService serviceImplementation="org.sonarlint.intellij.util.SonarLintActions"/>
+
+        <!-- Cannot be moved to service annotation because of 'serviceInterface' -->
         <applicationService serviceInterface="org.sonarlint.intellij.telemetry.SonarLintTelemetry"
```
damien-urruty-sonarsource, line 220: I might be wrong but for service classes that implement an interface it might still work, you can probably annotate the class that inherits the interface, but still use the interface when getting the service

src/main/resources/META-INF/plugin.xml
```diff
@@ -188,67 +188,68 @@
...
                             serviceImplementation="org.sonarlint.intellij.util.ApplicationThreadPoolExecutor"
                             testServiceImplementation="org.sonarlint.intellij.util.SynchronousThreadPoolExecutor"/>
 
-        <projectService serviceImplementation="org.sonarlint.intellij.core.ProjectBindingManager"/>
+        <!-- Cannot be moved to service annotation because of 'serviceInterface' -->
         <projectService serviceInterface="org.sonarlint.intellij.common.vcs.VcsService"
                         serviceImplementation="org.sonarlint.intellij.vcs.DefaultVcsService"/>
+
+        <!-- Cannot be moved to service annotation because of persisting project state -->
         <projectService serviceImplementation="org.sonarlint.intellij.config.project.SonarLintProjectSettingsStore"/>
```
damien-urruty-sonarsource, line 235: I saw classes in the platform that use both `@Service` and `@State` for persistent storage, so I think it could work

thahnen: I'll try that on Monday.

src/main/resources/META-INF/plugin.xml
thahnen: I'll try that on Monday.

src/main/resources/META-INF/plugin.xml
```diff
@@ -188,67 +188,68 @@
...
+
+        <!-- Cannot be moved to service annotation because of 'serviceInterface' -->
         <applicationService serviceInterface="org.sonarlint.intellij.telemetry.SonarLintTelemetry"
                             serviceImplementation="org.sonarlint.intellij.telemetry.SonarLintTelemetryImpl" />
-        <applicationService serviceImplementation="org.sonarlint.intellij.core.ServerNotificationsService"/>
-        <applicationService serviceImplementation="org.sonarlint.intellij.core.NodeJsManager"/>
-        <applicationService serviceImplementation="org.sonarlint.intellij.fs.ModuleFileEventsNotifier"/>
-        <applicationService serviceImplementation="org.sonarlint.intellij.module.ModulesRegistry"/>
+
         <applicationService serviceImplementation="org.sonarlint.intellij.core.BackendService"/>
```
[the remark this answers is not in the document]
thahnen: I can check once more, but there was an issue for another service because of the `@NonInjectable` (I assume).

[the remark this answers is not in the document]
thahnen: The comment was wrong, the service was migrated.

src/main/resources/META-INF/plugin.xml
thahnen: Done, see the question regarding the the application service in the slack channel for the last state

src/main/resources/META-INF/plugin.xml
thahnen: Investigated that: All the ones accesing the service via the interface cannot be migrated as of today. The docs are quite sparse on that but I found some info nonetheless. Maybe they can be migrated in the future, looks like they try to simplify it even further.

damien-urruty-sonarsource (approved):
Well done!

PR #661, opened 2023-05-22
SLI-945 Improve experience for users missing hotspots status change permissions

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/ui/UiUtils.kt
```diff
@@ -27,11 +27,10 @@ class UiUtils {
     companion object {
         @JvmStatic
         fun runOnUiThread(project: Project, runnable: Runnable) {
-            runOnUiThread(project, runnable, ModalityState.defaultModalityState())
+            runOnUiThread(project, ModalityState.defaultModalityState(), runnable)
         }
 
-        @JvmStatic
-        fun runOnUiThread(project: Project, runnable: Runnable, modality: ModalityState) {
+        fun runOnUiThread(project: Project, modality: ModalityState, runnable: Runnable) {
```
nquinquenel, line 33: Just as a question, what's the reason for this change?

src/main/java/org/sonarlint/intellij/ui/review/ReviewSecurityHotspotDialog.kt
```diff
@@ -85,20 +91,42 @@ class ReviewSecurityHotspotDialog(
...
+                        canChangeStatus = false
+                        setErrorText(checkResponse.notPermittedReason)
+                    }
+                    reviewPanel.display(checkResponse.allowedStatuses)
+                    changeStatusAction.isEnabled = checkResponse.isPermitted
+                    mainPanel.stopLoading()
+                }
+            }.exceptionally { error ->
+                SonarLintConsole.get(project).error(
+                    "Error while retrieving the list of allowed statuses for security hotspots",
```
nquinquenel, line 120:
```suggestion
                    "Error while retrieving the list of allowed statuses for Security Hotspots",
```

src/main/java/org/sonarlint/intellij/ui/review/ReviewSecurityHotspotDialog.kt
```diff
@@ -85,20 +91,42 @@ class ReviewSecurityHotspotDialog(
...
+                    error
+                )
+                displayErrorNotification(project, "The statuses for this security hotspot could not be retrieved.")
```
nquinquenel, line 123:
```suggestion
                displayErrorNotification(project, "The statuses for this Security Hotspot could not be retrieved.")
```

src/main/java/org/sonarlint/intellij/ui/review/ReviewSecurityHotspotDialog.kt
```diff
@@ -85,20 +91,42 @@ class ReviewSecurityHotspotDialog(
...
         init()
+        SonarLintUtils.getService(BackendService::class.java).checkStatusChangePermitted(connection.name, securityHotspotKey)
+            .thenAccept { checkResponse ->
+                runOnUiThread(project, ModalityState.stateForComponent(this@ReviewSecurityHotspotDialog.contentPane)) {
+                    if (!checkResponse.isPermitted) {
+                        canChangeStatus = false
+                        setErrorText(checkResponse.notPermittedReason)
+                    }
+                    reviewPanel.display(checkResponse.allowedStatuses)
+                    changeStatusAction.isEnabled = checkResponse.isPermitted
```
nquinquenel, line 115: By doing this, you override the callback function above saying that it should be enabled if the current status is different than the one currently applied on SQ.

I would simply change `canChangeStatus` to false or true depending on `if (!checkResponse.isPermitted)`, and remove this line, it should make it works

nquinquenel (commented):
Few missing bits but good overall!

src/main/java/org/sonarlint/intellij/ui/UiUtils.kt
damien-urruty-sonarsource: I prefer to have the Runnable as the last parameter this way I can use Kotlin [trailing lambdas](https://kotlinlang.org/docs/lambdas.html#passing-trailing-lambdas)

nquinquenel (approved):
LGTM, splendid :rocket:

src/main/java/org/sonarlint/intellij/tasks/FutureAwaitingTask.kt
```diff
@@ -0,0 +1,56 @@
...
+                future.cancel(true)
+                return null
+            }
+            return try {
+                future.get(100, TimeUnit.MILLISECONDS)
+            } catch (t: TimeoutException) {
+                continue
+            } catch (e: InterruptedException) {
+                throw InterruptedException("Interrupted")
+            } catch (e: CancellationException) {
```
nquinquenel, line 49: Do you know in which case this exception is reached? It is not when calling `future.cancel(true)`

damien-urruty-sonarsource: I believe the future can be canceled downstream (e.g. by the HTTP stack under certain situations)

PR #662, opened 2023-05-23
SLI-767 Disable the 'Update local storage' button when binding for the first time

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -146,11 +147,12 @@ public String getSelectedProjectKey() {
...
-    projectKeyTextField.setEnabled(connectionSelected);
-    projectKeyTextField.setEditable(connectionSelected);
-    searchProjectButton.setEnabled(connectionSelected);
-    updateStorageButton.setEnabled(connectionSelected);
+    var selectedConnection = getSelectedConnection();
+    var isAnyConnectionSelected = selectedConnection != null;
+    projectKeyTextField.setEnabled(isAnyConnectionSelected);
+    projectKeyTextField.setEditable(isAnyConnectionSelected);
+    searchProjectButton.setEnabled(isAnyConnectionSelected);
+    updateStorageButton.setEnabled(selectedConnection != null && getGlobalSettings().exists(selectedConnection.getName()));
```
eray-felek-sonarsource, line 155: we can use isAnyConnectionSelected instead of selectedConnection != null here

damien-urruty-sonarsource: Good point, I was worried that the IDE would complain about selectedConnection being nullable but it's fine

eray-felek-sonarsource (approved):

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -182,6 +182,10 @@ public Optional<ServerConnection> getServerConnectionByName(String name) {
       .findFirst();
   }
 
+  public boolean exists(String connectionName) {
```
eray-felek-sonarsource, line 185: We can use the name serverConnectionExists or connectionExist instead of exists

damien-urruty-sonarsource: :+1:

PR #663, opened 2023-05-23
SLI-936: Update Java analyzer

thahnen:
Java analyzer update and logical change regarding the `sonar.java.enablePreview` property.


damien-urruty-sonarsource (approved):
LGTM! You can probably close SLI-924 as implemented with SLI-936

PR #664, opened 2023-05-23
SLI-943: Update Kotlin analyzer

thahnen:
Tests done with local project and new rule [S6529](https://sonarsource.atlassian.net/browse/SONARKT-297).


nquinquenel (approved):
Could you add in plugin.xml the fact that there is a new rule? (or anything else new for Kotlin)

Otherwise, LGTM!

PR #665, opened 2023-05-23
SLI-833: Reworked the former solution

thahnen:
## Summary

Minor changes to the former solution to check for all module content roots and not the first one. Checking every content root is necessary when for example a source folder is outside of the main project, this can be achieved with Gradle very easily. The other option is to have a project span multiple drives in Windows (I have no idea why anybody should do that, but okey).

## Testing

To test the easiest case (file inside module with one content root) and the second easiest case (file outside module with one content root) are implemented as light tests. The *corner case* test (file inside module with multiple content roots not connected to one another) was implemented as a heavy test, as `com.intellij.testFramework.HeavyPlatformTestCase` provides the best way to create a standalone module with multiple content roots.


damien-urruty-sonarsource (approved):
Good job and kudos for the tests!

PR #667, opened 2023-05-23
SLI-958 Wait for server findings

nquinquenel:
Remove parameter `waitForServerFindings` and put true by default. As a result, we will always wait for the server matching to be finished before ending the analysis.

The drawback is that the analysis might stay blocked for up to 60 seconds (timeout duration for the tasks) when waiting for the server findings.

As a result, I improved the analysis cancellation process (currently, we check if the status is `CANCELLING` only during the analysis of the modules, and before waiting for the server findings, so pressing the cancel button during the fetching had no effect).

### DIFF
--- a/src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.java
+++ b/src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.java
@@ -49,17 +49,17 @@ public final class AnalysisSubmitter {
   public static final String ANALYSIS_TASK_TITLE = "SonarLint Analysis";
   private final Project project;
   private final OnTheFlyFindingsHolder onTheFlyFindingsHolder;
-  private Analysis currentAutoAnalysis;
+  private Cancelable currentManualAnalysis;
 
   public AnalysisSubmitter(Project project) {
     this.project = project;
     this.onTheFlyFindingsHolder = new OnTheFlyFindingsHolder(project);
   }
 
-  public void cancelCurrentAutoAnalysis() {
-    if (currentAutoAnalysis != null) {
-      currentAutoAnalysis.cancel();
-      currentAutoAnalysis = null;
+  public void cancelCurrentManualAnalysis() {
+    if (currentManualAnalysis != null) {
+      currentManualAnalysis.cancel();
+      currentManualAnalysis = null;
     }
   }
 
@@ -124,7 +124,7 @@ public final class AnalysisSubmitter {
     if (shouldExecuteInBackground(actionEvent)) {
       analyzeInBackground(files, TriggerType.ACTION, callback);
     } else {
-      analyzeInBackgroundableModal(files, TriggerType.ACTION, callback);
+      currentManualAnalysis = analyzeInBackgroundableModal(files, TriggerType.ACTION, callback);
     }
   }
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.java
```diff
@@ -139,17 +147,20 @@ private static boolean shouldExecuteInBackground(AnActionEvent e) {
   }
 
   private Cancelable analyzeInBackground(Collection<VirtualFile> files, TriggerType trigger, AnalysisCallback callback) {
-    var task = new Analysis(project, files, trigger, false, callback);
-    TaskRunnerKt.startBackgroundTask(project, ANALYSIS_TASK_TITLE, task::run);
-    return task;
+    if (shouldSkipAnalysis()) {
```
damien-urruty-sonarsource, line 150: I am not sure about this check here. It means if there is a long analysis running and users change some code, it won't be analyzed

src/main/java/org/sonarlint/intellij/util/FutureUtils.java
```diff
@@ -41,12 +44,34 @@ public static void waitForTask(Project project, Future<?> task, String taskName,
...
+  private static void waitForFuture(ProgressIndicator indicator, Future<?> future, Duration durationTimeout) throws InterruptedException, ExecutionException, TimeoutException {
+    long counter = 0;
+    while (counter < durationTimeout.toMillis()) {
+      counter += 100;
+      if (indicator.isCanceled()) {
+        future.cancel(true);
+        return;
+      }
+      try {
+        future.get(100, TimeUnit.MILLISECONDS);
```
damien-urruty-sonarsource, line 62: `100` is present twice here, maybe use a constant because they should change together

src/main/java/org/sonarlint/intellij/util/FutureUtils.java
```diff
@@ -41,12 +44,34 @@ public static void waitForTask(Project project, Future<?> task, String taskName,
...
 
-  public static void waitForTasks(Project project, List<Future<?>> updateTasks, String taskName) {
+  public static void waitForTasks(Project project, ProgressIndicator indicator, List<Future<?>> updateTasks, String taskName) {
     for (var f : updateTasks) {
-      waitForTask(project, f, taskName, Duration.ofSeconds(20));
+      waitForTask(project, indicator, f, taskName, Duration.ofSeconds(20));
     }
   }
 
+  private static void waitForFuture(ProgressIndicator indicator, Future<?> future, Duration durationTimeout) throws InterruptedException, ExecutionException, TimeoutException {
```
damien-urruty-sonarsource, line 53: You can probably make `FutureAwaitingTask` use this method

damien-urruty-sonarsource (commented):
It looks mostly good except one thing I have a doubt about, let's discuss

src/main/java/org/sonarlint/intellij/util/FutureUtils.java
nquinquenel: Not sure, yours is running indefinitely

damien-urruty-sonarsource: true

PR #672, opened 2023-05-24
Rework filters for Security Hotspots

nquinquenel:


src/main/java/org/sonarlint/intellij/ui/tree/SecurityHotspotTreeModelBuilder.java
```diff
@@ -63,8 +63,9 @@ public class SecurityHotspotTreeModelBuilder implements FindingTreeModelBuilder
   private static final Comparator<LiveSecurityHotspot> SECURITY_HOTSPOT_COMPARATOR = new SecurityHotspotComparator();
   private static final Comparator<LiveSecurityHotspotNode> SECURITY_HOTSPOT_WITHOUT_FILE_COMPARATOR = new LiveSecurityHotspotNodeComparator();
 
+  protected SecurityHotspotFilters currentFilter = SecurityHotspotFilters.DEFAULT_FILTER;
+  protected boolean currentResolvedFilter = false;
```
damien-urruty-sonarsource, line 67: I would suggest to improve the name `currentResolvedFilter`,it is misleading as it is not a filter. Something like `shouldShowResolvedHotspots` ?

damien-urruty-sonarsource (approved):
One minor comment, else LGTM

src/main/java/org/sonarlint/intellij/ui/tree/SecurityHotspotTreeModelBuilder.java
nquinquenel: Good point :+1:

PR #678, opened 2023-05-26
SLI-942: Don't analyze binary files

thahnen:
Binary files are explicetly ommited from being analyzed. `VirtualFileUtilsTests` extends `AbstractSonarLintHeavyTests` as using `AbstractSonarLintLightTests` did not work correctly 😞


damien-urruty-sonarsource (commented):
The check you implemented in ModuleFileSystem is not directly related to the analysis. I wonder if we need to add the same check where we check if a file should be included in the analysis. The rest looks good!

damien-urruty-sonarsource (approved):

PR #683, opened 2023-06-05
SLI-994 Implement new backend HTTP client

nquinquenel:


src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -128,7 +140,16 @@ class BackendService @NonInjectable constructor(private val backend: SonarLintBa
                         this@BackendService.projectClosed(project)
                     }
                 })
-        }.thenApply { backend }
+        }.get()
```
damien-urruty-sonarsource, line 143: I thought at some point we tried to avoid that to not risk blocking the EDT, but I might be wrong

src/test/java/org/sonarlint/intellij/core/BackendServiceTests.kt
```diff
@@ -102,6 +116,8 @@ class BackendServiceTests : AbstractSonarLintHeavyTests() {
 
     @Test
     fun test_notify_backend_when_opening_a_non_bound_project() {
+
+
```
damien-urruty-sonarsource, line 120: Maybe not intended ?

damien-urruty-sonarsource (approved):
A few small remarks, else LGTM

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -324,6 +334,49 @@ object SonarLintIntelliJClient : SonarLintClient {
...
+    override fun checkServerTrusted(params: CheckServerTrustedParams): CompletableFuture<CheckServerTrustedResponse>? {
+        val certificateFactory = CertificateFactory.getInstance("X.509")
+        val certificates: Array<X509Certificate> = params.chain.stream()
+            .map { certificateFactory.generateCertificate(ByteArrayInputStream(it.pem.toByteArray())) as X509Certificate }
+            .collect(Collectors.toList()).toTypedArray()
+        return try {
+            CertificateManager.getInstance().trustManager.checkServerTrusted(certificates, params.authType)
+            CompletableFuture.completedFuture(CheckServerTrustedResponse(true))
+        } catch (e: CertificateException) {
+            return CompletableFuture.completedFuture(CheckServerTrustedResponse(false))
```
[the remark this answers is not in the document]
nquinquenel: `CertificateException` is thrown whenever the user does not accept the certificate, this is not an error but we could still write some warn level log

PR #684, opened 2023-06-06
Adapt to recent backend changes

nquinquenel:


src/test/java/org/sonarlint/intellij/mediumtests/SecurityHotspotsMediumTest.kt
```diff
@@ -178,7 +178,7 @@ class SecurityHotspotsMediumTest : AbstractSonarLintLightTests() {
     }
 
     private fun createStorage(
-        serverVersion: String? = "99.9",
+        serverVersion: String? = "9.9",
```
henryju, line 181: Is this change really intended?

henryju (approved):

PR #686, opened 2023-06-07
SLI-975 Remove issues from the trees

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/ui/tree/TaintVulnerabilityTreeModelBuilder.kt
```diff
@@ -130,6 +135,19 @@ class TaintVulnerabilityTreeModelBuilder {
     return file.isValid
   }
 
+  fun remove(taintVulnerability: LocalTaintVulnerability) {
+    val file = taintVulnerability.file() ?: return
+    val fileNode = index.getFileNode(file)
+    fileNode?.findChildren { child ->
```
eray-felek-sonarsource, line 141: We can add a null check here(fileNode?)

eray-felek-sonarsource (changes requested):

src/main/java/org/sonarlint/intellij/ui/tree/TaintVulnerabilityTreeModelBuilder.kt
damien-urruty-sonarsource: If we are not able to find the file, it's maybe because the view has changed and no more shows the file, so there is nothing to remove. I don't think we need to warn the user

eray-felek-sonarsource (approved):

PR #690, opened 2023-06-08
SLI-974 Integrate the dialog with the backend

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -158,11 +161,19 @@ class BackendService @NonInjectable constructor(private val backend: SonarLintBa
     }
 
     private fun toSonarQubeBackendConnection(createdConnection: ServerConnection): SonarQubeConnectionConfigurationDto {
-        return SonarQubeConnectionConfigurationDto(createdConnection.name, createdConnection.hostUrl, createdConnection.isDisableNotifications)
+        return SonarQubeConnectionConfigurationDto(
+            createdConnection.name,
+            createdConnection.hostUrl,
+            createdConnection.isDisableNotifications
+        )
```
damien-urruty-sonarsource, line 168: There are some changes like this one that are purely related to formatting. Are you following the style [guidelines](https://github.com/SonarSource/sonar-developer-toolset)? If so, I would suggest to move that to a separate commit

src/main/java/org/sonarlint/intellij/ui/resolve/MarkAsResolvedDialog.kt
```diff
@@ -43,12 +63,42 @@ class MarkAsResolvedDialog (
                 val status = getStatus()
 
                 if(status != null){
-                    println("go to server")
-                    println("status is:" + status.name)
+                    SonarLintUtils.getService(BackendService::class.java)
+                        .markAsResolved(BackendService.moduleId(module), issueKey, status, isTaintVulnerability)
```
damien-urruty-sonarsource, line 67: Usually we pass the module to the backend service, and we make the call to `moduleId` there, I think it's better in terms of responsibility. I know we did the same for hotspots but we should change it too.

src/main/java/org/sonarlint/intellij/ui/resolve/MarkAsResolvedDialog.kt
```diff
@@ -43,12 +63,42 @@ class MarkAsResolvedDialog (
...
+                        .thenAccept {
+                            UiUtils.runOnUiThread(
```
damien-urruty-sonarsource, line 69: You could use a static import to shorten this

src/main/java/org/sonarlint/intellij/actions/MarkAsResolvedAction.kt
```diff
@@ -96,10 +111,43 @@ class MarkAsResolvedAction :
...
     override fun getIcon(flags: Int) = AllIcons.Actions.BuildLoadChanges
+
+    private class CheckIssueStatusChangePermission(
+        project: Project,
+        connection: ServerConnection,
+        issueKey: String,
+    ) :
+        FutureAwaitingTask<CheckStatusChangePermittedResponse>(
+            project,
+            "Checking Mark as Resolve Permission",
```
damien-urruty-sonarsource, line 150: Resolve -> Resolved

PR #692, opened 2023-06-09
SLI-976 Check permissions to mark as resolved

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/ui/ruledescription/RuleHeaderPanel.kt
```diff
@@ -101,36 +102,47 @@ class RuleHeaderPanel : JBPanel<RuleHeaderPanel>(FlowLayout(FlowLayout.LEFT)) {
...
 
-    private fun listenIssueButton(isValid: Boolean) {
+    private fun listenIssueButton(
+        isValid: Boolean, project: Project, file: VirtualFile, issueKey: String, isTaintVulnerability: Boolean,
+        liveIssue: LiveIssue?, taintVulnerability: LocalTaintVulnerability?,
+    ) {
         changeStatusButton.action = object : AbstractAction(MARK_AS_RESOLVED) {
             override fun actionPerformed(e: ActionEvent?) {
-                System.out.println("Open Dialog")
+                MarkAsResolvedAction().openMarkAsResolvedDialog(project, file, issueKey, isTaintVulnerability, liveIssue, taintVulnerability)
```
damien-urruty-sonarsource, line 145: I know that we did the same for hotspots, but I find it a bit strange to create the action from here. Creating the action is not useful because we don't invoke it. But that should work. Maybe we could make the method static or part of a service?

damien-urruty-sonarsource (approved):
Maybe one thing to simplify, we can discuss about it

PR #697, opened 2023-06-12
SLI-985 Show an icon to indicate server matched issues

damien-urruty-sonarsource:


src/main/resources/META-INF/plugin.xml
```diff
@@ -64,6 +64,7 @@
 
     <change-notes><![CDATA[
       <ul>
+        <li>8.3 - Allow marking issues existing on SonarQube/SonarCloud as resolved.</li>
```
nquinquenel, line 67: 8.4 :smiley:

damien-urruty-sonarsource: Good catch!

PR #699, opened 2023-06-12
SLI-984 Create a common panel

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/ui/panelutil/GenericPanel.kt
```diff
@@ -17,33 +17,35 @@
...
+import java.awt.event.MouseAdapter
+import java.awt.event.MouseEvent
 import javax.swing.BorderFactory
+import javax.swing.ButtonGroup
 
-class StatusPanel(
-    status: HotspotStatus,
-    isSelected: Boolean,
-) : JBPanel<StatusPanel>(GridBagLayout()) {
+open class GenericPanel (
```
damien-urruty-sonarsource, line 35: I think having "generic" in a name is too... "generic" :)
The purpose of the class is to show an option that the user can select among others, so maybe `OptionPanel`?

damien-urruty-sonarsource: And in fact I think we could go further with this class, there is still some duplicated logic around it, like creating the radio group. So what about having an `OptionsPanel`, to which you could provide a model (in the AWT spirit) containing the list of options?

eray-felek-sonarsource: Yes I was thinking the same thing for the name, I think OptionPanel is indeed a better... option :)

PR #704, opened 2023-06-16
SLI-983 Handle SSE for Security Hotspots

nquinquenel:


src/main/java/org/sonarlint/intellij/core/server/events/ServerEventsService.kt
```diff
@@ -85,17 +92,30 @@ class ServerEventsProductionService : ServerEventsService {
...
+            }
+            getService(project, AnalysisSubmitter::class.java).autoAnalyzeFiles(impactedFiles, TriggerType.SERVER_SENT_EVENT)
         }
     }
 
     private fun identifyProjectsImpactedByTaintEvent(event: ServerEvent): Set<Project> {
         val projectKey = when (event) {
             is TaintVulnerabilityRaisedEvent -> event.projectKey
             is TaintVulnerabilityClosedEvent -> event.projectKey
-            is IssueChangedEvent -> event.projectKey
```
nquinquenel, line 98: @damien-urruty-sonarsource what was the reason to put `IssueChangedEvent` along with the taint vulnerability events?

damien-urruty-sonarsource (approved):
Tested locally, LGTM!

PR #707, opened 2023-06-19
8.4 Hardening

nquinquenel:


src/main/java/org/sonarlint/intellij/ui/ruledescription/RuleParsingUtils.kt
```diff
@@ -0,0 +1,130 @@
...
+                when (it) {
+                    is HtmlFragment -> RuleHtmlViewer(false).apply { updateHtml(it.html) }
+                    is CodeExampleFragment -> RuleCodeSnippet(project, fileType, it).apply {
+                        Disposer.register(parent, this)
+                    }
+                }
+            }
+                .forEach { mainPanel.add(it) }
+
+            val scrollPane = ScrollPaneFactory.createScrollPane(mainPanel)
```
eray-felek-sonarsource, line 101: We can extract the scrollPane operations as a function (101 - 107)

src/main/java/org/sonarlint/intellij/ui/ruledescription/RuleParsingUtils.kt
```diff
@@ -0,0 +1,130 @@
...
+            var computedRuleDescription = ""
+            var matcherStart: Matcher = Pattern.compile("<pre[^>]*>").matcher(remainingRuleDescription)
+            var matcherEnd: Matcher = Pattern.compile(PRE_TAG_ENDING).matcher(remainingRuleDescription)
+
+            val section = Section()
+            val xmlElementFactory = XmlElementFactory.getInstance(project)
+            while (matcherStart.find() && matcherEnd.find()) {
+                val front: String = remainingRuleDescription.substring(0, matcherStart.start()).trim()
+
+                if (front.isNotBlank()) {
```
eray-felek-sonarsource, line 60: We are using similar logic on line 87. Maybe we can extract this as a function too.

eray-felek-sonarsource (approved):
Small comments, else LGTM!

src/main/java/org/sonarlint/intellij/ui/ruledescription/RuleParsingUtils.kt
nquinquenel: I think the benefits to extract this as a function are not great enough, it would make the code maybe less readable for minimal factorizing.

PR #711, opened 2023-06-21
SLI-987 Add null checks for caret

eray-felek-sonarsource:


nquinquenel (commented):
Few questions:
- What if the caret is null?
- Do we actually need a caret in the rule description? As of today, I don't remember seeing any caret, and do we really want a caret? Maybe we should put it null by default

eray-felek-sonarsource:
> Few questions:
> 
> * What if the caret is null?
> * Do we actually need a caret in the rule description? As of today, I don't remember seeing any caret, and do we really want a caret? Maybe we should put it null by default

If the caret is null, we get a null pointer exception on JTextComponent while clicking on the rule description. But, we avoid the caret exception while casting and we can still show the rule description(If we don't do a null check and get an exception while casting it is not possible to show anything).

Setting the caret to null by default hence is not an option.

I will be creating a separate ticket to handle the JTextComponent exception case.

nquinquenel (approved):

PR #712, opened 2023-06-21
SLI-1000: Python analyzer update

thahnen:
Tested the same way as for SonarLint for Eclipse


eray-felek-sonarsource (changes requested):
We could change to commit message to: "SLI-1000 Update embedded Python analyzer to 4.4.0.11907" to include version

eray-felek-sonarsource (approved):

PR #722, opened 2023-07-04
SLI-1006 Support PL/SQL analysis in Intellij Ultimate

eray-felek-sonarsource:


build.gradle.kts
```diff
@@ -217,6 +217,7 @@ dependencies {
     "sqplugins"("org.sonarsource.iac:sonar-iac-plugin:1.16.0.3845")
     if (artifactoryUsername.isNotEmpty() && artifactoryPassword.isNotEmpty()) {
         "sqplugins"("com.sonarsource.cpp:sonar-cfamily-plugin:6.46.0.62229")
+        "sqplugins"("com.sonarsource.plsql:sonar-plsql-plugin:3.9.0.5181")
```
henryju, line 220: I thought the goal was to only support PL/SQL in connected mode? If yes, then we should not embed it.

src/main/java/org/sonarlint/intellij/ui/ruledescription/RuleLanguages.kt
```diff
@@ -56,7 +56,8 @@ enum class RuleLanguages(private val language: Language, private vararg val file
     TERRAFORM(Language.TERRAFORM, "Terraform"),
     TS(Language.TS, *typescriptFileTypesByPriority),
     XML(Language.XML, "XML"),
-    YAML(Language.YAML, "YAML");
+    YAML(Language.YAML, "YAML"),
+    PLSQL(Language.PLSQL, "PL/SQL");
```
nquinquenel, line 60: I don't have `PL/SQL` as a file type in IntelliJ Ultimate, is this from a plugin?

PR #723, opened 2023-07-05
Feature/ef/enable plsql

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/ui/ruledescription/RuleLanguages.kt
```diff
@@ -56,7 +56,8 @@ enum class RuleLanguages(private val language: Language, private vararg val file
     TERRAFORM(Language.TERRAFORM, "Terraform"),
     TS(Language.TS, *typescriptFileTypesByPriority),
     XML(Language.XML, "XML"),
-    YAML(Language.YAML, "YAML");
+    YAML(Language.YAML, "YAML"),
+    PLSQL(Language.PLSQL, "PL/SQL");
```
nquinquenel, line 60: Testing with IntelliJ Ultimate, I don't have syntax highlighting for `PL/SQL`. Does it come from a plugin? We can probably check for primary SQL language from the IntelliJ file type and add it to the list of possible file types for PL/SQL, as we did for TypeScript

nquinquenel (approved):
LGTM

PR #728, opened 2023-07-12
Update to latest SLCORE

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -103,24 +112,22 @@ class BackendService @NonInjectable constructor(private val backend: SonarLintBa
...
         backend.initialize(
             InitializeParams(
-                HostInfoDto(ApplicationInfo.getInstance().versionName),
-                TelemetryManagerProvider.TELEMETRY_PRODUCT_KEY,
+                ClientInfoDto(
+                    ApplicationInfo.getInstance().versionName,
+                    TelemetryManagerProvider.TELEMETRY_PRODUCT_KEY,
+                    "SonarLint IntelliJ " + getService(SonarLintPlugin::class.java).version
+                ),
+                FeatureFlagsDto(true, true, true, true, true),
```
henryju, line 120: Aren't TaintVulnerabilities disabled in Rider or CLion?

henryju (approved):

src/main/java/org/sonarlint/intellij/core/BackendService.kt
damien-urruty-sonarsource: The corresponding flag is deprecated and should be removed. The CORE has the knowledge of when to download taint vulnerabilities or not

damien-urruty-sonarsource: *deprecated and not used

PR #731, opened 2023-07-17
SLI-1026 Support PL/SQL analysis in CLion

eray-felek-sonarsource:


clion/src/main/java/org/sonarlint/intellij/clion/CFamilyFileExclusionContributor.java
```diff
@@ -21,13 +21,19 @@
...
 import org.sonarlint.intellij.common.analysis.FileExclusionContributor;
 
 public class CFamilyFileExclusionContributor implements FileExclusionContributor {
 
   @Override
   public ExcludeResult shouldExclude(Module module, VirtualFile fileToAnalyze) {
+    var psiFile = PsiManager.getInstance(module.getProject()).findFile(fileToAnalyze);
+    if(!(psiFile instanceof OCPsiFile)){
+      return  ExcludeResult.notExcluded();
+    }
```
damien-urruty-sonarsource, line 36: The formatting looks strange, could you make sure it's ok?

src/main/java/org/sonarlint/intellij/plsql/PLSQLLanguageActivator.java
```diff
@@ -26,8 +26,11 @@
 public class PLSQLLanguageActivator implements LanguageActivator {
 
   @Override
-  public void amendLanguages(Set<Language> enabledLanguages) {
-    enabledLanguages.add(Language.PLSQL);
+  public void amendLanguages(Set<Language> enabledLanguages, boolean isConnected) {
+    boolean isAlreadyEnabled = enabledLanguages.contains(Language.PLSQL);
```
damien-urruty-sonarsource, line 30: I don't see a situation where it would already be enabled. Do you really need to check that?
Also `enabledLanguages` is a Set, so there won't be any duplication if you add it twice.
I suggest removing the `!isAlreadyEnabled` part in the if below

damien-urruty-sonarsource (approved):
2 small comments, else LGTM

PR #735, opened 2023-07-20
SLI-1027 Provide a stable ID to the backend for module scopes

damien-urruty-sonarsource:


henryju (approved):
LGTM, let's see if this is a technique we can use on other IDEs

PR #742, opened 2023-07-31
SLCORE update

eray-felek-sonarsource:


thahnen (approved):
LGTM! Do not forget to update SLCORE once again after releasing!

PR #746, opened 2023-08-03
SLI-1043 Selecting proxy should return right hostname and port

nquinquenel:


src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -374,8 +376,14 @@ object SonarLintIntelliJClient : SonarLintClient {
...
         val uri = URI.create(params.uri)
         val proxiesResponse =
-            SelectProxiesResponse(CommonProxy.getInstance().select(uri).stream().map { ProxyDto(it.type(), uri.host, uri.port) }
-                .collect(Collectors.toList()))
+            SelectProxiesResponse(CommonProxy.getInstance().select(uri).stream().map {
+                if (it.type() != Proxy.Type.DIRECT) {
+                    val socketAddress = it.address() as InetSocketAddress
+                    ProxyDto(it.type(), socketAddress.hostString, socketAddress.port)
+                } else {
+                    ProxyDto.NO_PROXY
```
thahnen, line 384: On which occasion can a proxy not be a proxy? I have to say I find this quite irritating that it actually happened. The rest of the changes make sense 👍

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -374,8 +376,14 @@ object SonarLintIntelliJClient : SonarLintClient {
     override fun selectProxies(params: SelectProxiesParams): CompletableFuture<SelectProxiesResponse> {
         val uri = URI.create(params.uri)
         val proxiesResponse =
-            SelectProxiesResponse(CommonProxy.getInstance().select(uri).stream().map { ProxyDto(it.type(), uri.host, uri.port) }
-                .collect(Collectors.toList()))
+            SelectProxiesResponse(CommonProxy.getInstance().select(uri).stream().map {
+                if (it.type() != Proxy.Type.DIRECT) {
+                    val socketAddress = it.address() as InetSocketAddress
```
thahnen, line 381: You're sure this can't fail?

thahnen (approved):
Looks good, on a normal day I would actually suggest creating an integration test for it. But I have never really used a proxy in an IDE and I'm clueless on how to emulate this very behaviour!

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
thahnen: Ah it's already like this on JDK side, so nvm.

PR #747, opened 2023-08-04
SLI-1045 Retrieve SC organizations from backend

nquinquenel:


src/main/java/org/sonarlint/intellij/config/global/wizard/OrganizationStep.java
```diff
@@ -38,16 +38,16 @@
...
 import org.sonarlint.intellij.tasks.GetOrganizationTask;
-import org.sonarsource.sonarlint.core.serverapi.organization.ServerOrganization;
+import org.sonarsource.sonarlint.core.clientapi.backend.connection.org.OrganizationDto;
 
 import static javax.swing.JList.VERTICAL;
 
 public class OrganizationStep extends AbstractWizardStepEx {
   private final WizardModel model;
-  private JList<ServerOrganization> orgList;
+  private JList<OrganizationDto> orgList;
```
thahnen, line 47: I'm a bit irritated but also not too deep into that matter: Why does `OrganicationDto` not implement the interface `ServerOrganization`? We could have used it then anyway ^^

src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -400,6 +404,20 @@ class BackendService @NonInjectable constructor(private val backend: SonarLintBa
         return initializedBackend.connectionService.validateConnection(params)
     }
 
+    fun listUserOrganizations(server: ServerConnection): CompletableFuture<ListUserOrganizationsResponse> {
```
thahnen, line 407: Very good! I have to check if we have something similar in Eclipse or if we might have the same issue but no community report as of yet!
I can check manually but we have no Eclipse-compatible project using SonarCloud if I'm not mistaken, maybe you are aware?

thahnen (approved):
LGTM!

src/main/java/org/sonarlint/intellij/core/BackendService.kt
nquinquenel: It's already implemented on Eclipse side, no worries about that
For testing, you should just connect to your SC account by using GitHub for example, it should do the job

src/main/java/org/sonarlint/intellij/config/global/wizard/OrganizationStep.java
nquinquenel: I guess a DTO should not have such logic 🤷🏻

PR #749, opened 2023-08-07
SLI-1036 Check if project is disposed at beginning of read actions

damien-urruty-sonarsource:


nquinquenel (approved):
LGTM, thanks for the fix!

PR #755, opened 2023-08-09
SLI-1055 Add attributes to Finding

nquinquenel:


eray-felek-sonarsource (commented):
LGTM! We can merge when checks have passed

PR #758, opened 2023-08-10
Update list for current file tab

nquinquenel:


src/main/java/org/sonarlint/intellij/config/global/rules/RulesFilterModel.java
```diff
@@ -101,10 +101,10 @@ public void reset(boolean triggerListener) {
   }
 
   public boolean filter(RulesTreeNode.Rule rule) {
-    if (showOnlyEnabled && !rule.isActivated()) {
+    if (showOnlyEnabled && Boolean.FALSE.equals(rule.isActivated())) {
```
damien-urruty-sonarsource, line 104: I find it strange that `isActivated` returns the wrapper type, but it was already like this. Do you know if it can actually happen that it is null?

src/main/java/org/sonarlint/intellij/config/global/rules/RulesTreeTable.java
```diff
@@ -100,8 +103,15 @@ public void mouseMoved(final MouseEvent e) {
...
-          var severityAndType = rule.severity() + " " + rule.type();
-          label.setText(StringUtil.capitalize(severityAndType.replace('_', ' ').toLowerCase(Locale.US)));
+          // TODO: Remove condition once we have the mapping done on CORE
+          // It should always be new CCT for standalone rules
+          if (rule.attribute() != null && !rule.impacts().isEmpty()) {
+            var highestQualityImpact = Collections.max(rule.impacts().entrySet(), Map.Entry.comparingByValue(Comparator.comparing(Enum::ordinal)));
+            var impactText = StringUtil.capitalize(highestQualityImpact.getValue().toString().toLowerCase(Locale.ENGLISH));
+            var qualityText = highestQualityImpact.getKey().toString().toLowerCase(Locale.ENGLISH);
+            var text = impactText + " " + qualityText;
+            label.setText(StringUtil.capitalize(text.replace('_', ' ').toLowerCase(Locale.US)));
```
damien-urruty-sonarsource, line 113: Is there a reason to use sometimes ENGLISH sometimes US as the locale?

src/main/java/org/sonarlint/intellij/ui/nodes/LocalTaintVulnerabilityNode.kt
```diff
@@ -27,28 +27,53 @@ import org.sonarlint.intellij.finding.issue.vulnerabilities.LocalTaintVulnerabil
...
-                SonarLintIcons.severity(issue.severity())
+
+        if (issue.getCleanCodeAttribute() != null && issue.getImpacts().isNotEmpty()) {
+            val (quality, impact) = Collections.max(
+                issue.getImpacts().entries,
+                Map.Entry.comparingByValue(Comparator.comparing { obj: ImpactSeverity -> obj.ordinal })
+            )
+            val impactText = StringUtil.capitalize(impact.toString().lowercase())
+            val qualityText = quality.toString().lowercase()
+            toolTipText = "$impactText $qualityText"
```
damien-urruty-sonarsource, line 51: I have the impression we are doing 3 times the same thing: compute a displayable text out of impacts and quality. Do you think it would make sense to have this code in a single place?

damien-urruty-sonarsource (approved):
Some questions but LGTM

src/main/java/org/sonarlint/intellij/config/global/rules/RulesFilterModel.java
nquinquenel: For a `RulesTreeNode.Rule` it doesn't happen in the code, but for `RulesTreeNode.Language` is can

So in this case it's 'safe' because it's only Rule

src/main/java/org/sonarlint/intellij/config/global/rules/RulesTreeTable.java
nquinquenel: Indeed... I don't there is any reason, we have a mix of both in the project. I'd safe it's safe to choose any of them and stick with it... let's choose `ENGLISH`?

src/main/java/org/sonarlint/intellij/ui/nodes/LocalTaintVulnerabilityNode.kt
nquinquenel: I don't see a proper way of doing this, the max value is also used for other things (get icon type), and those string attributes come from both LiveFinding and RuleTreeNode so it's difficult to factorize, and a utils class seems overkill, WDYT?

PR #759, opened 2023-08-10
Make SonarCloud URL customizable

damien-urruty-sonarsource:


common/src/main/java/org/sonarlint/intellij/common/util/SonarLintUtils.java
```diff
@@ -47,8 +48,9 @@
 public class SonarLintUtils {
 
   private static final Logger LOG = Logger.getInstance(SonarLintUtils.class);
-  private static final String[] SONARCLOUD_ALIAS = {"https://sonarqube.com", "https://www.sonarqube.com",
-    "https://www.sonarcloud.io", "https://sonarcloud.io"};
+  public static final String SONARCLOUD_URL = System.getProperty("sonarlint.internal.sonarcloud.url", "https://sonarcloud.io");
+  private static final Set<String> SONARCLOUD_ALIAS = Set.copyOf(List.of("https://sonarqube.com", "https://www.sonarqube.com",
```
nquinquenel, line 52: Why not directly create a Set with `Set.of`?

build.gradle.kts
```diff
@@ -171,6 +171,7 @@ val runIdeDirectory: String by project
 
 tasks.runIde {
     systemProperty("sonarlint.telemetry.disabled", "true")
+    //systemProperty("sonarlint.internal.sonarcloud.url", "https://sonarcloud.io/")
```
nquinquenel, line 174: Is there a reason to keep this comment?

nquinquenel (approved):
LGTM, just few questions

common/src/main/java/org/sonarlint/intellij/common/util/SonarLintUtils.java
damien-urruty-sonarsource: `Set.of` throws when there are duplicates in the arguments. What I want to do here is to keep `https://sonarcloud.io` in all cases, plus the custom one that might be configured

build.gradle.kts
damien-urruty-sonarsource: this is how you activate the custom URL. I will add a comment to explain that

nquinquenel (approved):

PR #761, opened 2023-08-11
BUILD-3527 Do not override build version on Cirrus `validate_task`

elrzn:
This script is doing the following:

     Replacing version 9.0-SNAPSHOT with 9.0.0.$build_version

Which should not happen in this step.


nquinquenel (approved):
LGTM, this was also the script I identified as being problematic when I reported the issue

Pipeline seems ok 👍🏻

PR #762, opened 2023-08-16
BUILD-3523 Fix version missing from dogfood.json

matemoln:
Tested by interactively executing the commands on CirrusCI


nquinquenel (approved):
LGTM, should fix the ${PROJECT_VERSION} variable not being set

PR #769, opened 2023-08-21
DOC-780 add link to new docs and harden others

kevin-hinz:
An update to the README to point to the new docs site & harden other links formally pointing to sonarlint.org URLs.


nquinquenel:
We have some other links to `sonarlint.org`, should they all be renamed to sonar `sourcesonarsource.com/products/sonarlint`?

kevin-hinz:
> We have some other links in the code to `sonarlint.org`, should they all be renamed to `sonarsource.com/products/sonarlint`?

Most of the time: yes. 
Sometimes the text is more specific to IntelliJ, in that case, `sonarsource.com/products/sonarlint/features/jetbrains/` is more relevant.

damien-urruty-sonarsource:
I think from the README we should point to the docs website and not the marketing website. I expect people reading a README to be technical people. This would be in line with what I did with Eclipse. Would that be ok @kevin-hinz ?

PR #770, opened 2023-08-21
SLI-1044 Check if rule panel is disposed when parsing description

nquinquenel:
Reproducible by setting a thread sleep inside the new thread starting when loading the rules.

This happens when we save the rule panel, on `apply()` we do a `reset()` that load a bunch of stuff, including rules in a background thread.

With some unfortunate timing, or by simulating a thread sleep, we see that the panel is disposed but the background thread keeps going and tries to register the parent (which is now disposed).


eray-felek-sonarsource (approved):
LGTM! We can merge when the pipelines are passing

PR #772, opened 2023-08-23
SLI-1073 Update indicator for power save mode

eray-felek-sonarsource:


thahnen (changes requested):
LGTM! Could you please add an IT for the power save mode, we have one for excluding the file where the information pops up if I'm not mistaken!

thahnen:
@eray-felek-sonarsource please fix both SQ issues and then you're ready to go I assume ^^

PR #773, opened 2023-08-23
SLI-1032 Let users reopen resolved issues

nquinquenel:


src/main/java/org/sonarlint/intellij/editor/SonarExternalAnnotator.java
```diff
@@ -135,6 +137,7 @@ private static void addAnnotation(Project project, LiveFinding finding, TextRang
 
     if (finding instanceof LiveIssue) {
       intentionActions.add(new MarkAsResolvedAction((LiveIssue) finding));
+      intentionActions.add(new ReopenIssueAction((LiveIssue) finding));
```
damien-urruty-sonarsource, line 140: Is this possible? Do we annotate issues that are marked resolved? I'm not sure we want to bring attention on them for users

damien-urruty-sonarsource (approved):
A few comments but LGTM!
I think you can also update the release notes

src/main/java/org/sonarlint/intellij/actions/MarkAsResolvedAction.kt
```diff
@@ -170,9 +174,14 @@ class MarkAsResolvedAction(
     }
 
     override fun isEnabled(e: AnActionEvent, project: Project, status: AnalysisStatus): Boolean {
-        // always disabled for standalone mode
-        // the checks will be done inside the popup for connected mode
-        return serverConnection(project) != null
+        var issue: Issue? = e.getData(ISSUE_DATA_KEY)
```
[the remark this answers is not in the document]
nquinquenel: Good point

src/main/java/org/sonarlint/intellij/editor/SonarExternalAnnotator.java
nquinquenel: I'll confirm this with PM, in the meantime I'm removing all the annotation part, just having them visible in the list is enough probably

nquinquenel: It should not be annotated

PR #782, opened 2023-08-28
Fix UI freeze when analyzing a file

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/analysis/OnTheFlyFindingsHolder.kt
```diff
@@ -95,6 +96,9 @@ class OnTheFlyFindingsHolder(private val project: Project) : FileEditorManagerLi
     }
 
     fun clearCurrentFile() {
+        if (selectedFile == null) {
+            selectedFile = SonarLintUtils.getSelectedFile(project)
+        }
```
nquinquenel, line 101: `if else` instead of double `if`?

nquinquenel (approved):
LGTM, one minor comment

PR #791, opened 2023-09-05
Feature/mmf 3312 Version update and tests

eray-felek-sonarsource:
With the version update


damien-urruty-sonarsource (approved):
LGTM, but you probably need to rebase this PR onto master so that quickfix preview stuff disappear from here

PR #793, opened 2023-09-11
SLI-1099 Emphasize new issues in the current file tab

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/ui/nodes/SummaryNode.java
```diff
@@ -27,15 +27,20 @@
 public class SummaryNode extends AbstractNode {
   private String emptyText;
   private final boolean forSecurityHotspot;
+  private final boolean forOldIssues;
+
+  private static final String FORMAT = "Found %d %s in %d %s";
```
nquinquenel, line 32: Good practice to have `static final` fields at the very top

src/main/java/org/sonarlint/intellij/ui/tree/IssueTreeModelBuilder.java
```diff
@@ -77,9 +77,14 @@ public IssueTreeModelBuilder() {
   /**
    * Creates the model with a basic root
    */
-  public DefaultTreeModel createModel() {
+  public DefaultTreeModel createModel(boolean isOldIssue) {
     latestIssues = Collections.emptyMap();
-    summary = new SummaryNode();
+    if(isOldIssue){
```
nquinquenel, line 82: Please use the Sonar code style 🤯

src/main/java/org/sonarlint/intellij/ui/nodes/SummaryNode.java
```diff
@@ -61,10 +66,13 @@ public String getText() {
     }
 
     if (forSecurityHotspot) {
-      return String.format("Found %d %s in %d %s", findings, findings == 1 ? "Security Hotspot" : "Security Hotspots", files, files == 1 ? "file" : "files");
+      return String.format(FORMAT, findings, findings == 1 ? "Security Hotspot" : "Security Hotspots", files, files == 1 ? "file" : "files");
     }
 
-    return String.format("Found %d %s in %d %s", findings, findings == 1 ? "issue" : "issues", files, files == 1 ? "file" : "files");
+    if(forOldIssues){
```
nquinquenel, line 72: Code style

src/main/java/org/sonarlint/intellij/ui/nodes/SummaryNode.java
```diff
@@ -61,10 +66,13 @@ public String getText() {
...
+      return String.format(FORMAT, findings, findings == 1 ? "issue" : "issues", files, files == 1 ? "file since old analysis" : "files since old analysis");
+    }
+    return String.format(FORMAT, findings, findings == 1 ? "issue" : "issues", files, files == 1 ? "file since new analysis" : "files since new analysis");
```
nquinquenel, line 75: I think you can set a `String` for the text `since old/new analysis`, and just return this `String` inside this line, so it avoids having duplication and the `if (forOldIssues)`

src/main/java/org/sonarlint/intellij/ui/CurrentFilePanel.java
```diff
@@ -57,6 +58,7 @@ public class CurrentFilePanel extends AbstractIssuesPanel {
   private static final String SPLIT_PROPORTION_PROPERTY = "SONARLINT_ISSUES_SPLIT_PROPORTION";
   private final JBPanelWithEmptyText issuesPanel;
   private final JScrollPane treeScrollPane;
+
```
nquinquenel, line 61:
```suggestion
```

src/main/java/org/sonarlint/intellij/ui/CurrentFilePanel.java
```diff
@@ -65,7 +67,11 @@ public CurrentFilePanel(Project project) {
     // Issues panel
     setToolbar(actions());
 
-    treeScrollPane = ScrollPaneFactory.createScrollPane(tree);
+    var treePanel = new JBPanel<ReportPanel>(new VerticalFlowLayout(0, 0));
```
nquinquenel, line 70: `JBPanel<ReportPanel>` I don't think we are in the report tab?

nquinquenel (approved):
LGTM for the current state (until we come back to it with real data), just a few comments

src/main/java/org/sonarlint/intellij/ui/nodes/SummaryNode.java
eray-felek-sonarsource: Since those text will change after backend integration I think it is not needed for now

nquinquenel: I think it will be the same after, something like this:
`var sinceText = forOldIssues ? "since old analysis" : "since new analysis";`
`return String.format(FORMAT, findings, findings == 1 ? "issue" : "issues", files, files == 1 ? "file " + sinceText : "files " + sinceText);`

Later, we'll just replace "old analysis" by whatever we retrieve from core

PR #794, opened 2023-09-11
SLI-1094 Let users change the new code focus option

nquinquenel:


src/main/java/org/sonarlint/intellij/actions/SonarFocusOnNewCode.kt
```diff
@@ -0,0 +1,45 @@
...
+
+import com.intellij.ide.DataManager
+import com.intellij.openapi.actionSystem.ActionPlaces
+import com.intellij.openapi.actionSystem.AnActionEvent
+import org.sonarlint.intellij.common.ui.SonarLintConsole
+import org.sonarlint.intellij.config.Settings
+
+class SonarFocusOnNewCode(private var isFocusOnNewCode: Boolean) : AbstractSonarToggleAction() {
+
+    constructor() : this(false)
```
damien-urruty-sonarsource, line 30: Do you need 2 constructors here?

src/main/java/org/sonarlint/intellij/actions/SonarFocusOnNewCode.kt
```diff
@@ -0,0 +1,45 @@
...
+    fun isFocusOnNewCode(): Boolean = isFocusOnNewCode
+
+    override fun isSelected(e: AnActionEvent): Boolean = isFocusOnNewCode
+
+    override fun setSelected(e: AnActionEvent, isSelected: Boolean) {
+        if (isFocusOnNewCode != isSelected) {
+            isFocusOnNewCode = isSelected
+            DataManager.getInstance().dataContextFromFocusAsync
+                .onSuccess { Settings.getGlobalSettings().setFocusOnNewCode(ActionPlaces.TOOLWINDOW_CONTENT, it, isFocusOnNewCode) }
+                .onError { e.project?.let { SonarLintConsole.get(it).error("Could not get data context for action to focus on new code") } }
```
damien-urruty-sonarsource, line 41: As both onSuccess and onError are async, you need to check if the project is present and not disposed before doing something

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -148,6 +153,19 @@ public void setRulesByKey(Map<String, Rule> rules) {
...
+    return isFocusOnNewCode;
+  }
+
+  public void setFocusOnNewCode(String place, DataContext dataContext, boolean focusOnNewCode) {
+    isFocusOnNewCode = focusOnNewCode;
+    var action = (SonarFocusOnNewCode) ActionManager.getInstance().getAction("SonarLint.SetFocusNewCode");
+    if (action.isFocusOnNewCode() != isFocusOnNewCode) {
+      var anActionEvent = AnActionEvent.createFromDataContext(place, null, dataContext);
+      action.setSelected(anActionEvent, focusOnNewCode);
+    }
```
damien-urruty-sonarsource, line 166: I'm not sure in terms of responsibility if this class is the right place to hold this, we usually have mostly setters and getters here

src/main/java/org/sonarlint/intellij/ui/traffic/light/SonarLintTrafficLightAction.kt
```diff
@@ -0,0 +1,46 @@
...
+class SonarLintTrafficLightAction(private val editor: Editor) : AnAction(), CustomComponentAction {
+
+    override fun createCustomComponent(presentation: Presentation, place: String): JComponent {
+        return SonarLintTrafficLightWidget(this, presentation, place, editor)
+    }
+
+    override fun update(e: AnActionEvent) {
+        val component = e.presentation.getClientProperty(CustomComponentAction.COMPONENT_KEY) as SonarLintTrafficLightWidget?
+        component?.refresh()
+    }
```
damien-urruty-sonarsource, line 40: I implemented that to have something quickly, but I'm not sure it's sufficient to refresh the view. How often is the update called? Shouldn't we listen for changes in the findings? If it's refreshed very often maybe it's sufficient

src/main/java/org/sonarlint/intellij/actions/SonarFocusOnNewCode.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.sonarlint.intellij.common.ui.SonarLintConsole
+import org.sonarlint.intellij.config.Settings
+
+class SonarFocusOnNewCode(private var isFocusOnNewCode: Boolean) : AbstractSonarToggleAction() {
+
+    constructor() : this(false)
+
+    fun isFocusOnNewCode(): Boolean = isFocusOnNewCode
+
+    override fun isSelected(e: AnActionEvent): Boolean = isFocusOnNewCode
```
damien-urruty-sonarsource, line 34: In fact I wonder if you couldn't get rid of the `isFocusOnNewCode` var and directly use the value from the global settings. This might make the code snippet present in the settings and that I mentioned above useless

damien-urruty-sonarsource (approved):
A few questions but looks great!

src/main/java/org/sonarlint/intellij/actions/SonarFocusOnNewCode.kt
nquinquenel: `Action class must have a no-argument constructor`

And due to Kotlin not adding a default constructor, I need to have this one there 🤷🏻

nquinquenel: Edit: not needed anymore after refactoring to no variable

src/main/java/org/sonarlint/intellij/actions/SonarFocusOnNewCode.kt
nquinquenel: You're right, for some reasons I thought it wouldn't get updated on changes but I was wrong, makes like easier!

src/main/java/org/sonarlint/intellij/actions/SonarFocusOnNewCode.kt
nquinquenel: removed after refactoring

src/main/java/org/sonarlint/intellij/ui/traffic/light/SonarLintTrafficLightAction.kt
nquinquenel: It gets called very often (up to multiple times in seconds), and I think it is not necessary (I'm even afraid it has negative impacts on the resources), I will try to come up with a better solution 👍🏻

PR #795, opened 2023-09-13
MMF-3245 Have an option to focus on new code

nquinquenel:


src/main/java/org/sonarlint/intellij/actions/SonarFocusOnNewCode.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    override fun setSelected(e: AnActionEvent, isSelected: Boolean) {
+        Settings.getGlobalSettings().isFocusOnNewCode = isSelected
+        e.project?.let { getService(it, SonarLintToolWindow::class.java).refreshViews() }
+    }
+
+    override fun updatePresentation(project: Project, presentation: Presentation) {
+        val enabled = Settings.getSettingsFor(project).isBound
+        presentation.setEnabled(enabled)
+        presentation.text = "Set Focus on New Code" + if (enabled) "" else " (connected mode only)"
```
nquinquenel, line 42: We may have to capitalize `connected mode only` as well

src/main/java/org/sonarlint/intellij/ui/ReportPanel.java
```diff
@@ -330,4 +420,10 @@ private void disableEmptyDisplay(Boolean state) {
   public void dispose() {
     lastAnalysisPanel.dispose();
   }
+
+  public void refreshView() {
+    if(lastAnalysisResult != null) {
```
nquinquenel, line 425:
```suggestion
    if (lastAnalysisResult != null) {
```

src/main/java/org/sonarlint/intellij/actions/SonarFocusOnNewCode.kt
damien-urruty-sonarsource: You are probably right, fixing this

PR #798, opened 2023-09-14
SLI-1102 Emphasize focus on new code in the 'Taint Vulnerabilities' tab

nquinquenel:


eray-felek-sonarsource (approved):
LGTM!

src/main/java/org/sonarlint/intellij/ui/nodes/SummaryNode.java
```diff
@@ -64,13 +66,18 @@ public String getText() {
       return emptyText;
     }
 
-    if (forSecurityHotspot) {
-      return String.format(FORMAT, findings, findings == 1 ? "Security Hotspot" : "Security Hotspots", files, files == 1 ? "file" : "files");
+    String sinceText = "";
+    String newOrOldOrNothing = "";
```
eray-felek-sonarsource, line 70: Maybe we can have a better name here, how about focusText?

eray-felek-sonarsource (approved):
Small comment other than that LGTM!

src/main/java/org/sonarlint/intellij/ui/nodes/SummaryNode.java
damien-urruty-sonarsource: Not sure the suggestion is clearer. Naming is hard... Let's revisit later

PR #801, opened 2023-09-21
SLI-1101 Emphasize new focus on Security Hotspots

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/finding/hotspot/LiveSecurityHotspot.java
```diff
@@ -94,4 +94,7 @@ public boolean isResolved() {
     return status.isResolved();
   }
 
+  public boolean isNewFinding() {
+    return isNewFinding();
+  }
```
damien-urruty-sonarsource, line 99: I think this is dead code (and happily in fact, else would raise a stack overflow!)

damien-urruty-sonarsource (approved):
LGTM

PR #805, opened 2023-09-28
SLI-1101 Fix automatic expansion for hotspots

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/ui/SecurityHotspotsPanel.java
```diff
@@ -184,6 +184,7 @@ public int updateHotspots(Map<VirtualFile, Collection<LiveSecurityHotspot>> hots
       var newHotspotsAfterFiltering = securityHotspotTreeBuilder.applyCurrentFiltering(project);
       var hotspotsAfterFiltering = oldHotspotsAfterFiltering + newHotspotsAfterFiltering;
       displaySecurityHotspotsAfterFiltering(hotspotsAfterFiltering);
+      oldSecurityHotspotTree.collapseRow(0);
```
damien-urruty-sonarsource, line 187: I am not sure that would work: what if a user has previously expanded the tree? It would become collapsed at the next analysis? I think we should do this only when the view is built, but then we should not intervene anymore

damien-urruty-sonarsource (approved):

PR #806, opened 2023-09-28
Trigger an analysis when the sync is made

damien-urruty-sonarsource:


eray-felek-sonarsource (approved):
LGTM, there is 1 UI test failing maybe it is fp

PR #817, opened 2023-10-02
Fix SQ bugs

nquinquenel:
#CleanAsYouCode

### DIFF
--- a/git/src/main/kotlin/org/sonarlint/intellij/git/GitRepo.kt
+++ b/git/src/main/kotlin/org/sonarlint/intellij/git/GitRepo.kt
@@ -47,16 +47,11 @@ class GitRepo(private val repo: GitRepository, private val project: Project, pri
                 val distance = distance(project, repo, head, localBranchHash.asString()) ?: continue
                 branchesPerDistance.computeIfAbsent(distance) { HashSet() }.add(serverBranchName)
             }
-            if (branchesPerDistance.isEmpty()) {
-                return null
-            }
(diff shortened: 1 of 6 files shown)


git/src/main/kotlin/org/sonarlint/intellij/git/GitRepo.kt
```diff
@@ -51,11 +51,12 @@ class GitRepo(private val repo: GitRepository, private val project: Project, pri
...
-            if (bestCandidates.contains(serverMainBranch)) {
-                // Favor the main branch when there are multiple candidates with the same distance
-                serverMainBranch
-            } else bestCandidates.first()
+            branchesPerDistance[minDistance]?.let {
+                if (it.contains(serverMainBranch)) {
+                    // Favor the main branch when there are multiple candidates with the same distance
+                    serverMainBranch
+                } else it.first()
+            } ?: throw Exception("Could not find the branch with the minimum distance")
```
damien-urruty-sonarsource, line 59: I think the code was easier to read before. What about replacing:
```
if (branchesPerDistance.isEmpty()) {
  return null
}
val minDistance = branchesPerDistance.keys.stream().min(Comparator.naturalOrder()).get()
val bestCandidates: Set<String?> = branchesPerDistance[minDistance]!!
```
with something like:
`val bestCandidates = branchesPerDistance.minByOrNull { it.key }?.value ?: return null`

I think it would be equivalent (to be tested)

PR #822, opened 2023-10-05
Handle "Open in IDE" feature for issues

nquinquenel:


src/main/java/org/sonarlint/intellij/ui/vulnerabilities/TaintVulnerabilitiesPanel.kt
```diff
@@ -307,6 +313,39 @@ class TaintVulnerabilitiesPanel(private val project: Project) : SimpleToolWindow
         currentStatus?.let { populateTrees(it) }
     }
 
+    fun doesTaintExist(taintKey: String): LocalTaintVulnerability? {
```
damien-urruty-sonarsource, line 316: Reading the name I was expecting a boolean to be returned. Maybe something like `findIssueByKey`?
I think there is the same for "normal" issues

src/main/java/org/sonarlint/intellij/actions/ShowFindingCallable.kt
```diff
@@ -0,0 +1,104 @@
...
+                    )
+            }
+        }
+    }
+
+    private fun showIssue(toolWindow: SonarLintToolWindow) {
+        toolWindow.openCurrentFileTab()
+        toolWindow.bringToFront()
+
+        SonarLintUtils.getService(project, SonarLintToolWindow::class.java).getAndSelectIssue(showFinding)
```
damien-urruty-sonarsource, line 95: Another suggestion for the name: `trySelectIssue`

src/main/java/org/sonarlint/intellij/actions/ShowFindingCallable.kt
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private fun showTaintVulnerability(toolWindow: SonarLintToolWindow) {
+        toolWindow.openTaintVulnerabilityTab()
+        toolWindow.bringToFront()
+
+        SonarLintUtils.getService(project, SonarLintToolWindow::class.java).getAndSelectTaintVulnerability(showFinding)
```
damien-urruty-sonarsource, line 102: Same here

src/main/java/org/sonarlint/intellij/finding/ShowFinding.kt
```diff
@@ -0,0 +1,68 @@
...
+    val type: Class<T>,
+) {
+
+    companion object {
+
+        fun handleFlows(project: Project, flows: List<FlowDto>): List<Flow> {
+            val matcher = TextRangeMatcher(project)
+            return flows.stream().map { flow ->
+                flow.locations.stream().map {
+                    SonarLintIntelliJClient.tryFindFile(project, it.filePath)?.let { file ->
```
damien-urruty-sonarsource, line 48: It seems a bit strange to me to rely on the client here, maybe we should have an utility somewhere else that both this class and the client would rely on

src/main/java/org/sonarlint/intellij/ui/vulnerabilities/TaintVulnerabilitiesPanel.kt
```diff
@@ -307,6 +313,39 @@ class TaintVulnerabilitiesPanel(private val project: Project) : SimpleToolWindow
...
+        val issue = treeBuilder.findTaintByKey(taintKey)?.let {
+            oldTreeBuilder.findTaintByKey(taintKey)
+        }
+        return issue
```
damien-urruty-sonarsource, line 320: You can probably use:
> return treeBuilder.findTaintByKey(taintKey) ?: oldTreeBuilder.findTaintByKey(taintKey)

src/main/java/org/sonarlint/intellij/ui/vulnerabilities/TaintVulnerabilitiesPanel.kt
```diff
@@ -307,6 +313,39 @@ class TaintVulnerabilitiesPanel(private val project: Project) : SimpleToolWindow
...
+    }
+
+    fun <T: Finding> trySelectFilteredTaintVulnerability(taintVulnerability: LocalTaintVulnerability?, showFinding: ShowFinding<T>) {
```
damien-urruty-sonarsource, line 323: The name of the method made me think, but this is more a general question: what if the issue/hotspot/taint exists and was detected but is filtered out in the views? For example if the finding is resolved. I saw that you already check in both trees in the case of new code period so this should work, I'm not sure about resolved findings

damien-urruty-sonarsource (approved):
Some questions/suggestions, else LGTM

src/main/java/org/sonarlint/intellij/ui/vulnerabilities/TaintVulnerabilitiesPanel.kt
```diff
@@ -307,6 +313,39 @@ class TaintVulnerabilitiesPanel(private val project: Project) : SimpleToolWindow
...
+                    "The taint vulnerability could not be detected by SonarLint in the current code.")
+            return
+        }
+        val matcher = TextRangeMatcher(project)
+        val rangeMarker = computeReadActionSafely<RangeMarker>(project) {
+            matcher.matchWithCode(showFinding.file, showFinding.textRange, showFinding.codeSnippet)
+        }
+        if (rangeMarker == null) {
+            SonarLintProjectNotifications.get(project)
+                .notifyUnableToOpenFinding("issue", "The issue could not be detected by SonarLint in the current code.")
```
nquinquenel, line 340: Note to me, it should be taint vulnerability here

src/main/java/org/sonarlint/intellij/ui/vulnerabilities/TaintVulnerabilitiesPanel.kt
nquinquenel: The `findTaintByKey` method looks at any issue in the tree, regardless of the filters. So this is effectively working, but if it was resolved and the filter is not enabled, the issue won't appear in the list (but still, it will be highlighted and the rule description is shown, going with the flow of a regular finding, not a server finding).

This is an improvement that I'd like to have: when opening the finding in the issue, we could have the finding selected/focused in the list, so the user is aware of which one it is. But it also requires a bit more work because if the finding is filtered out, we need the user to enable the filters. We did something similar for security hotspot, we showed a notification with an action to clear the filtering. But for hotspot the workflow is different...

PR #823, opened 2023-10-06
SLI-1098 Fix resolved issue bug

eray-felek-sonarsource:


damien-urruty-sonarsource (commented):
I think the `allowResolvedIssues` fix is low risk, in the other one we modify a field and I'm not sure of the impact, maybe Nicolas will know better

PR #825, opened 2023-10-10
Improve build pipeline duration

damien-urruty-sonarsource:


.cirrus.yml
```diff
@@ -42,28 +43,28 @@ eks_builder_container: &BUILDER_CONTAINER_DEFINITION
 
 ec2_instance: &WINVM_DEFINITION
   experimental: true
-  image: base-windows-jdk17-v*
+  image: base-windows-jdk11-v*
```
nquinquenel, line 46: Is there a benefit to move back to jdk11?

nquinquenel (approved):
LGTM, that's some clean up!

.cirrus.yml
damien-urruty-sonarsource: Gradle was downloading a jdk11 behind the scenes, so I tried forcing it to 11 here, but now it is downloading a JDK17. I will drop this commit as anyway we are moving everything to Java 17 this week

PR #830, opened 2023-10-16
SLI-1136 Increase CPU

eray-felek-sonarsource:


nquinquenel:
I think the increase should essentially be for the `validate_windows` task in which we hit the CPU and memory limit:

image

Also, `validate_windows` is our main bottleneck task overall

nquinquenel (approved):

PR #835, opened 2023-10-17
10.0 Hardening

nquinquenel:


src/main/java/org/sonarlint/intellij/editor/ApplyQuickFixIntentionAction.kt
```diff
@@ -63,7 +63,7 @@ class ApplyQuickFixIntentionAction(private val fix: QuickFix, private val ruleKe
      *  interacting with the telemetry.
      */
     override fun invoke(project: Project, editor: Editor, file: PsiFile) {
-        if (!fix.isApplicable()) {
+        if (!fix.isApplicable() || !fix.isWithinBounds(editor.document)) {
```
damien-urruty-sonarsource, line 66: Small suggestion: move this second test inside `isApplicable`, passing the document as parameter. The fix is applicable if it is within the bounds of the document (plus other conditions)

damien-urruty-sonarsource (approved):
Only one small comment, LGTM!

PR #840, opened 2023-10-17
SLI-1086 Update Intellij build version

eray-felek-sonarsource:


its/src/main/resources/META-INF/plugin.xml
```diff
@@ -26,7 +26,7 @@
   <vendor url="https://sonarsource.com">SonarSource</vendor>
   <description>Fake Plugin To Run SonarLint UI tests with multiple flavors</description>
 
-  <idea-version since-build="213.5744.223"/>
+  <idea-version since-build="223"/>
```
damien-urruty-sonarsource, line 29: I think we should set a more precise version here, else we don't know exactly against which version we build

src/main/resources/META-INF/plugin.xml
```diff
@@ -161,7 +161,7 @@
     </change-notes>
 
     <!-- please see http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/build_number_ranges.html for description -->
-    <idea-version since-build="213.5744.223"/>
+    <idea-version since-build="223"/>
```
damien-urruty-sonarsource, line 164: Same here

gradle.properties
```diff
@@ -1,9 +1,9 @@
 version=9.2-SNAPSHOT
 org.gradle.jvmargs=-Xmx8G -XX:MaxMetaspaceSize=512M
 sonarlintCoreVersion=9.4.0.74778
-intellijBuildVersion=IC-2021.3
-clionBuildVersion=CL-2021.3
-riderBuildVersion=RD-2021.3
+intellijBuildVersion=IC-2022.3
+clionBuildVersion=CL-2022.3
+riderBuildVersion=RD-2022.2
```
damien-urruty-sonarsource, line 6: I think it's the first time we use a different version for one flavor. It means for Rider we build against a lower version than the one we officially support (as the real minimal version is defined in the `plugin.xml`). I don't know if this could be a problem, maybe if some APIs were dropped, but I don't think we used deprecated APIs. What is the reason for not using the same version?

.cirrus.yml
```diff
@@ -159,13 +159,13 @@ qa_task:
     variable_name: GOLAND_KEY
   matrix:
     - env:
-        IDEA_VERSION: 'IC-2021.3.3'
+        IDEA_VERSION: 'IC-2022.3'
         QA_CATEGORY: 'Idea2021'
```
damien-urruty-sonarsource, line 163: You should also change the `QA_CATEGORY` labels

gradle.properties
eray-felek-sonarsource: Rider 2022.3 has a bug that is well-known with Gradle. In order to support 2022.3 and forward I am using this version.

damien-urruty-sonarsource (approved):

PR #841, opened 2023-10-17
Should close Got It tooltips if showing

nquinquenel:


its/src/test/kotlin/org/sonarlint/intellij/its/BaseUiTest.kt
```diff
@@ -171,10 +171,13 @@ open class BaseUiTest {
     }
 
     private fun closeAllGotItTooltips() {
-        remoteRobot.findAll(ComponentFixture::class.java, GotItTooltipFixture.firstButton()).forEach {
-            it.click()
-            // Quick hack because another Got It Tooltip might not be visible directly
-            Thread.sleep(500)
+        for (i in 0..1) {
```
damien-urruty-sonarsource, line 174: What if we introduce another Got-it toolitp in the future?
Maybe we should call findAll as many times as it returns something?

nquinquenel: I wanted to avoid that because it can find a tooltip but since it's not showing it cannot be closed. I added a `while` but with a max tries mechanism in case it's stuck

damien-urruty-sonarsource (approved):

PR #847, opened 2023-10-20
Run plugin verifier at night for all JetBrains IDEs

nquinquenel:


damien-urruty-sonarsource (approved):
The download of the appropriate version is a bit like black magic for me, but I trust it works. 
What if the build fails, do we receive a notification somehow?

nquinquenel:
I agree this is some black magic, likely hard to maintain. I tried to make it so we have the least manual changes to do (especially regarding EAP builds that are changing every week). There is no big risk, it's just a nightly pipeline but I can understand if we should not merge it in this state 😅 

Currently, we don't have any notification. I don't know if there is an easy way with Cirrus. I know SonarCloud used a [python script ](https://github.com/SonarSource/sonarcloud-core/blob/master/.cirrus/slack-notifier/sonarcloud/slack/notifier.py)to generate slack message, maybe we could just copy them but I have no time for this now.

PR #848, opened 2023-10-20
SLI-1141 Fix focus on new code checkbox stay unchecked

eray-felek-sonarsource:


nquinquenel (approved):
Good catch

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalOptionsPanel.java
```diff
@@ -126,6 +126,7 @@ public void save(SonarLintGlobalSettings settings) {
     getService(CleanAsYouCodeService.class).setFocusOnNewCode(focusOnNewCode.isSelected());
     settings.setAutoTrigger(autoTrigger.isSelected());
     settings.setNodejsPath(nodeJsPath.getText());
+    settings.setFocusOnNewCode(focusOnNewCode.isSelected());
```
damien-urruty-sonarsource, line 129: This is normally done through line 126, but the problem is that the Settings instance used in the CaYCService is not right. So good catch!
I suggest keeping this responsibility in the service, but passing the settings instance to the setFocusOnNewCode()

src/main/java/org/sonarlint/intellij/cayc/CleanAsYouCodeService.kt
```diff
@@ -36,11 +37,20 @@ class CleanAsYouCodeService {
 
     fun setFocusOnNewCode(isFocusOnNewCode: Boolean) {
         getGlobalSettings().isFocusOnNewCode = isFocusOnNewCode
+        refresh()
```
damien-urruty-sonarsource, line 40: Last suggestion: this first method could call the second by providing getGlobalSettings() as the parameter.

damien-urruty-sonarsource (approved):

PR #850, opened 2023-10-23
Dependencies should be downloaded from Repox in priority

nquinquenel:
Looking at Cirrus logs, we download every dependency from maven central (Sonar scanners, etc.). We should probably do it from Repox if the dependencies are available there.


damien-urruty-sonarsource (approved):
Good catch!

PR #855, opened 2023-10-26
SLI-1146 Avoid relying on deprecated APIs

damien-urruty-sonarsource:


.cirrus.yml
```diff
@@ -106,7 +106,7 @@ validate_task:
   eks_container:
     <<: *BUILDER_CONTAINER_DEFINITION
     cpu: 6
-    memory: 8G
+    memory: 10G
```
nquinquenel, line 109: I wonder how this works, the instance used is t3.small, which has 2GB of memory (and 2 vCPU)

nquinquenel (approved):
LGTM

.cirrus.yml
damien-urruty-sonarsource: I will try to clarify that separately

damien-urruty-sonarsource: From what I observe in the graphics on the build, this value is used as we are now a bit below 10G in effective usage

PR #876, opened 2023-11-27
SLI-1178 Support SonarLint for CLion Nova

damien-urruty-sonarsource:


clion/src/main/java/org/sonarlint/intellij/clion/AnalyzerConfiguration.java
```diff
@@ -175,19 +175,33 @@ static ForcedLanguage getSonarLanguage(OCLanguageKind languageKind) {
...
+    } catch (ClassNotFoundException e) {
+      SonarLintConsole.get(project).debug("Could not check if remote or WSL toolchain is used");
+      return false;
     }
-    return cppEnvironment != null && (cppEnvironment.getToolSet().isRemote() || cppEnvironment.getToolSet().isWSL() || cppEnvironment.getToolSet().isDocker());
+  }
+
+  private static void tryIfClassAccessible() throws ClassNotFoundException {
+    Class.forName("com.jetbrains.cidr.cpp.cmake.workspace.CMakeWorkspace");
+    Class.forName("com.jetbrains.cidr.cpp.toolchains.CPPEnvironment");
```
damien-urruty-sonarsource, line 204: Do you really need to check both? Why? If so, a comment might be useful

clion/src/main/java/org/sonarlint/intellij/clion/CFamilyAnalysisConfigurator.java
```diff
@@ -47,4 +52,13 @@ public AnalysisConfiguration configure(Module module, Collection<VirtualFile> fi
...
   }
 
+  private static boolean isSupported() {
+    try {
+      Class.forName("com.jetbrains.rider.cpp.fileType.psi.CppFile");
+      return false;
+    } catch (ClassNotFoundException e) {
+      return true;
+    }
+  }
```
damien-urruty-sonarsource, line 62: Hmm I find this a bit surprising: if the class is present, then we consider it is not supported? Are the return values inverted, or should the method name be improved?

clion/src/main/java/org/sonarlint/intellij/clion/CFamilyFileExclusionContributor.java
```diff
@@ -40,4 +43,14 @@ public ExcludeResult shouldExclude(Module module, VirtualFile fileToAnalyze) {
     }
     return ExcludeResult.excluded(configurationResult.getSkipReason());
   }
+
+  private static boolean isSupported() {
```
damien-urruty-sonarsource, line 47: Same here

gradle.properties
```diff
@@ -1,9 +1,11 @@
...
+resharperBuildVersion=RD-2024.1-SNAPSHOT
 intellijBuildVersion=IC-2022.3.1
 # Should be the same as the one embedded into SonarLint Core
 protobufVersion=3.25.1
 omnisharpVersion=1.39.10
 org.gradle.jvmargs=-Xmx8G -XX:MaxMetaspaceSize=512M
-org.gradle.caching=true
\ No newline at end of file
+org.gradle.caching=true
+kotlin.stdlib.default.dependency = false
```
damien-urruty-sonarsource, line 11: Is this line related to CLion? I think I added it myself but I don't remember, if you know, adding a comment telling why it's needed would help

clion/src/main/java/org/sonarlint/intellij/clion/AnalyzerConfiguration.java
```diff
@@ -175,19 +175,33 @@ static ForcedLanguage getSonarLanguage(OCLanguageKind languageKind) {
...
+          if (cppEnvironment != null) {
+            return cppEnvironment.getToolSet().isRemote() || cppEnvironment.getToolSet().isWSL() || cppEnvironment.getToolSet().isDocker();
+          }
         }
       }
+      SonarLintConsole.get(project).debug("Not using remote or WSL toolchain");
+      return false;
+    } catch (ClassNotFoundException e) {
+      SonarLintConsole.get(project).debug("Could not check if remote or WSL toolchain is used");
+      return false;
```
damien-urruty-sonarsource, line 198: I prefer what you did in other places with a `isSupported` call that lets you return early from the function if needed, instead of a big try-catch

clion/src/main/java/org/sonarlint/intellij/clion/AnalyzerConfiguration.java
```diff
@@ -175,19 +175,33 @@ static ForcedLanguage getSonarLanguage(OCLanguageKind languageKind) {
...
-        cppEnvironment = getCMakeCppEnvironment(cMakeWorkspace, configuration);
-      } else {
-        cppEnvironment = tryReflection(initializedWorkspace);
-        if (cppEnvironment != null) {
-          break;
+    try {
+      tryIfClassAccessible();
+      final var initializedWorkspaces = CidrWorkspace.getInitializedWorkspaces(project);
+      for (var initializedWorkspace : initializedWorkspaces) {
+        if (Class.forName("com.jetbrains.cidr.cpp.cmake.workspace.CMakeWorkspace").isInstance(initializedWorkspace)) {
```
damien-urruty-sonarsource, line 182: Why do you need reflection here? Can't we assume this class is present if the 2 others are? It's also a change of behavior since it might throw, further `initializedWorkspaces` might not be checked

clion/src/test/java/org/sonarlint/intellij/clion/BuildWrapperJsonGeneratorTest.java
```diff
@@ -26,6 +26,8 @@
 import java.util.Map;
 import java.util.TreeMap;
 import org.junit.jupiter.api.Test;
+import org.sonarlint.intellij.clion.AnalyzerConfiguration;
+import org.sonarlint.intellij.clion.BuildWrapperJsonGenerator;
```
damien-urruty-sonarsource, line 30: Those added imports look suspicious, are they needed?

src/main/resources/META-INF/plugin-resharper.xml
damien-urruty-sonarsource, line 1: Using the name `plugin-resharper.xml` might be confusing as we also support Rider. The optional module is `com.intellij.modules.rider.cpp.core` which is probably a subpart of Resharper. 

This [doc](https://plugins.jetbrains.com/docs/intellij/plugin-compatibility.html#modules-specific-to-functionality) mentions:
> module `com.intellij.modules.rider`: Connection to ReSharper Process in Background

So the name does not reflect that we are relying on the cpp part of Resharper. Maybe `clion-resharper` to disambiguate? The same remark for the gradle module name, folder and package names

resharper/src/main/java/org/sonarlint/intellij/resharper/AnalyzerConfiguration.java
```diff
@@ -0,0 +1,312 @@
...
+        }
+      }
+    }
+    return cppEnvironment != null && (cppEnvironment.getToolSet().isRemote() || cppEnvironment.getToolSet().isWSL() || cppEnvironment.getToolSet().isDocker());
+  }
+
+  @Nullable
+  private CPPEnvironment tryReflection(CidrWorkspace initializedWorkspace) {
+    // Use reflection to check if workspace is instanceof com.jetbrains.cidr.project.workspace.WorkspaceWithEnvironment interface has
+    // getEnvironment() method
```
damien-urruty-sonarsource, line 205: Do we still need to use reflection in the `resharper` module? Also we try to get things from `com.jetbrains.cidr` I thought this was not present anymore

resharper/src/main/java/org/sonarlint/intellij/resharper/AnalyzerConfiguration.java
```diff
@@ -0,0 +1,312 @@
...
+      return cMakeWorkspace.getProfileInfoFor(cMakeConfiguration).getEnvironment();
+    } catch (ExecutionException e) {
+      throw new IllegalStateException(e);
+    }
+  }
+
+  private static Method getPreprocessorDefinesMethod() {
+    if (preprocessorDefinesMethod == null) {
+      try {
+        preprocessorDefinesMethod = OCCompilerSettings.class.getMethod("getPreprocessorDefines");
```
damien-urruty-sonarsource, line 252: Same here

resharper/src/main/java/org/sonarlint/intellij/resharper/CFamilyFileExclusionContributor.java
```diff
@@ -0,0 +1,49 @@
...
+
+import static org.sonarlint.intellij.common.util.SonarLintUtils.isCLion;
+
+public class CFamilyFileExclusionContributor implements FileExclusionContributor {
+
+  @Override
+  public ExcludeResult shouldExclude(Module module, VirtualFile fileToAnalyze) {
+    if (!isCLion()) {
+      return ExcludeResult.notExcluded();
+    }
```
damien-urruty-sonarsource, line 37: Is this check needed? Won't this extension point be loaded only when it is the case?

resharper/src/main/java/org/sonarlint/intellij/resharper/CFamilyAnalysisConfigurator.java
```diff
@@ -0,0 +1,57 @@
...
+import org.sonarlint.intellij.common.ui.SonarLintConsole;
+
+import static org.sonarlint.intellij.common.util.SonarLintUtils.isCLion;
+
+public class CFamilyAnalysisConfigurator implements AnalysisConfigurator {
+
+  @Override
+  public AnalysisConfiguration configure(Module module, Collection<VirtualFile> filesToAnalyze) {
+    var result = new AnalysisConfiguration();
+    if (!isCLion()) {
```
damien-urruty-sonarsource, line 35: Same here

damien-urruty-sonarsource (commented):
I left some comments, it looks mostly OK.

I think we should run the ITs on this new version of CLion to check if everything works as expected, probably on a new axis.

There is a lot of code duplication. I am not very concerned, because this part of SonarLint is never changed, and the "legacy" code will be removed at some point when we raise the minimal supported version. One thing you could do is introduce a `clion-commons` module to put the common code if needed

clion/src/main/java/org/sonarlint/intellij/clion/AnalyzerConfiguration.java
nquinquenel: On the latest version of CLion, when ReSharper is off, these 2 classes/packages got removed. So in the current state, we don't support remote/WSL in this case.

clion/src/main/java/org/sonarlint/intellij/clion/CFamilyAnalysisConfigurator.java
nquinquenel: CppFile is only present when ReSharper is on, so when this is the case I don't want to run the regular CLion analysis

resharper/src/main/java/org/sonarlint/intellij/resharper/CFamilyFileExclusionContributor.java
nquinquenel: It's loaded when `com.intellij.modules.rider.cpp.core` is present, which can be the case in Rider, or in CLion with ReSharper on. But we don't want to run the 'ReSharper CLion' analysis on Rider for instance

clion/src/test/java/org/sonarlint/intellij/clion/BuildWrapperJsonGeneratorTest.java
nquinquenel: Unused imports from leftovers after the copy paste, to be removed indeed

gradle.properties
nquinquenel: You added it and I don't remember the reason as well, seems like no difference. IIRC it was due to some warnings during build time, but might be wrong

clion/src/main/java/org/sonarlint/intellij/clion/AnalyzerConfiguration.java
damien-urruty-sonarsource: What I mean is: wouldn't it be sufficient to check only for one of them? Why having 2 checks?

clion/src/main/java/org/sonarlint/intellij/clion/CFamilyAnalysisConfigurator.java
damien-urruty-sonarsource: Then maybe we could find a more informative name for the method, because `isSupported` does not tell what is supported. Something like `isClionResharperOn()`? And then invert the logic

gradle.properties
damien-urruty-sonarsource: Could it be related to the `LinkageError` with the `GotItTooltip` we had originally? I also found this [doc](https://plugins.jetbrains.com/docs/intellij/using-kotlin.html#kotlin-standard-library). If it's not needed, then we can drop it

resharper/src/main/java/org/sonarlint/intellij/resharper/CFamilyFileExclusionContributor.java
damien-urruty-sonarsource: Could you add a comment? I think it would be useful to remember why this condition exists

resharper/src/main/java/org/sonarlint/intellij/resharper/AnalyzerConfiguration.java
nquinquenel: Not everything is gone from `com.jetbrains.cidr`, it's only specific packages, there is no official documentation about the changes so we discover case by case

clion/src/main/java/org/sonarlint/intellij/clion/AnalyzerConfiguration.java
nquinquenel: Moved

mpaladin (approved):

mpaladin (approved):
@nquinquenel the sooner you put it in dogfood the sooner I can ask my squad to test the plugin with the new engine.

PR #885, opened 2023-12-05
SLI-1200 Provide contextual help in Current File Tab

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/util/HelpLabelUtils.kt
```diff
@@ -22,16 +22,30 @@ package org.sonarlint.intellij.util
...
 
         @JvmStatic
-        fun createCleanAsYouCode() =
+        fun currentFileHelp() =
                 ContextHelpLabel.createWithLink(null,
-                        "Deliver clean code by focusing on code that was recently modified",
-                        "Learn more about Clean as You Code", true) { BrowserUtil.browse(FOCUS_ON_NEW_CODE_LINK) }
+                        "SonarLint analyzes in real-time the active file only and reports any issues found in this view. " +
+                            "SonarCloud complements SonarLint by detecting more across the whole project.",
+                        "Try SonarCloud for free", true) { BrowserUtil.browse(SONARCLOUD_PRODUCT_SIGNUP_LINK) }
```
damien-urruty-sonarsource, line 36: Shouldn't you have the link on `currentFileHelpNotConnected` instead of `currentFileHelp`?

src/main/java/org/sonarlint/intellij/util/HelpLabelUtils.kt
```diff
@@ -22,16 +22,30 @@ package org.sonarlint.intellij.util
...
     companion object {
 
         @JvmStatic
-        fun createCleanAsYouCode() =
+        fun currentFileHelp() =
                 ContextHelpLabel.createWithLink(null,
-                        "Deliver clean code by focusing on code that was recently modified",
-                        "Learn more about Clean as You Code", true) { BrowserUtil.browse(FOCUS_ON_NEW_CODE_LINK) }
+                        "SonarLint analyzes in real-time the active file only and reports any issues found in this view. " +
+                            "SonarCloud complements SonarLint by detecting more across the whole project.",
```
damien-urruty-sonarsource, line 35: Also, shouldn't the part about SonarCloud be displayed only in standalone mode?

src/main/java/org/sonarlint/intellij/util/HelpLabelUtils.kt
damien-urruty-sonarsource, line 36: This is a call to action and will need to be recorded with telemetry. You can have a look at `LinkTelemetry.browseWithTelemetry` and check other usages of this method

src/main/java/org/sonarlint/intellij/util/HelpLabelUtils.kt
```diff
@@ -22,16 +22,30 @@ package org.sonarlint.intellij.util
...
+                        "Try SonarCloud for free", true) { BrowserUtil.browse(SONARCLOUD_PRODUCT_SIGNUP_LINK) }
+
+        @JvmStatic
+        fun currentFileHelpNotConnected() =
```
damien-urruty-sonarsource, line 39: The other similar methods in this file start with `create`, I would align the names for consistency

src/main/java/org/sonarlint/intellij/ui/CurrentFileConnectedModePanel.java
```diff
@@ -79,22 +78,34 @@ private void createPanel() {
...
+
     connectedCard = new ConnectedCard();
+    connectedCardWrapper.add(helpLabelTextConnected);
+    connectedCardWrapper.add(helpLabelConnected);
+    connectedCardWrapper.add(connectedCard, gce);
 
     panel.add(new EmptyCard(), EMPTY);
     panel.add(new NotConnectedCard(), NOT_CONNECTED);
-    panel.add(connectedCard, CONNECTED);
+    panel.add(notConnectedCard, NOT_CONNECTED);
```
damien-urruty-sonarsource, line 104: Those 2 lines seem a bit strange, it gives the impression that we are adding 2 times the same thing

damien-urruty-sonarsource (commented):
I tested quickly, looks good on the UI. Only suggestion: add a small space between the text and the question mark

src/main/java/org/sonarlint/intellij/ui/CurrentFileConnectedModePanel.java
```diff
@@ -79,24 +86,57 @@ private void createPanel() {
...
           SonarLintActions.getInstance().configure(), e, panel, null, true);
       }
     };
-    Stream.of(panel.getComponents()).forEach(c -> c.addMouseListener(clickListener));
-
-    layout.show(panel, NOT_CONNECTED);
+    connectedCard.addMouseListener(clickListener);
+    notConnectedCard.addMouseListener(clickListener);
+    emptyCard.addMouseListener(clickListener);
+    errorCard.addMouseListener(clickListener);
```
damien-urruty-sonarsource, line 139: Do you need all these click listeners? This has the effect of changing the cursor on hover over "What"s in this view" to a finger, but nothing happens on the click

src/main/java/org/sonarlint/intellij/util/HelpLabelUtils.kt
```diff
@@ -28,10 +28,24 @@ class HelpLabelUtils {
...
+                        "SonarLint analyzes in real-time the active file only and reports any issues found in this view. " +
+                            "SonarCloud complements SonarLint by detecting more across the whole project.",
+                        "Try SonarCloud for free", true) {
+                    LinkTelemetry.SONARCLOUD_SIGNUP_PAGE.browseWithTelemetry()
+                }
+
+        @JvmStatic
+        fun createCurrentFileHelp() =
+            ContextHelpLabel.create(
+                "SonarLint analyzes in real-time the active file only and reports any issues found in this view.")
```
damien-urruty-sonarsource, line 42: You can probably extract the text in a variable instead of duplicating it twice. I expect we will always use the same text

src/main/java/org/sonarlint/intellij/ui/CurrentFileConnectedModePanel.java
```diff
@@ -79,24 +86,57 @@ private void createPanel() {
     layout = new CardLayout();
     panel = new JPanel(layout);
 
+    var notConnectedCardWrapper = new JPanel(new GridBagLayout());
```
damien-urruty-sonarsource, line 89: This `CurrentFileConnectedModePanel` class was probably badly named before and even more now. Not sure which name to pick instead

src/main/java/org/sonarlint/intellij/ui/CurrentFileConnectedModePanel.java
```diff
@@ -79,24 +86,57 @@ private void createPanel() {
...
+    notConnectedCardWrapper.add(helpLabelNotConnected, gcs);
+    notConnectedCardWrapper.add(notConnectedCard, gce);
 
-    panel.add(new EmptyCard(), EMPTY);
-    panel.add(new NotConnectedCard(), NOT_CONNECTED);
-    panel.add(connectedCard, CONNECTED);
-    panel.add(new ErrorCard(), ERROR);
+    connectedCardWrapper.add(helpLabelTextConnected);
+    connectedCardWrapper.add(helpLabelConnected, gcs);
+    connectedCardWrapper.add(connectedCard, gce);
```
damien-urruty-sonarsource, line 116: In fact, I wonder if we shouldn't extract the logic for the "What's in this view" in a different panel. One effect is that if no file is opened, the "empty" card gets selected, so the "What's in this view" disappears. We can discuss that online if you want

damien-urruty-sonarsource (commented):
Another round of comments, I can help you implement some of them

src/main/java/org/sonarlint/intellij/ui/CurrentFileConnectedModePanel.java
```diff
@@ -166,18 +166,18 @@ private void switchCards() {
...
         var projectBindingManager = getService(project, ProjectBindingManager.class);
         projectBindingManager.tryGetServerConnection().ifPresentOrElse(serverConnection -> {
-          try {
-            var module = SonarLintAppUtils.findModuleForFile(selectedFile, project);
-            if (module == null) {
-              switchCard(EMPTY);
-            } else {
-              connectedCard.updateTooltip(module, serverConnection);
-              switchCard(CONNECTED);
+            try {
```
damien-urruty-sonarsource, line 169: I think you could revert this change, I don't see anything new compared to the old code. Or was it badly formatted?

src/main/java/org/sonarlint/intellij/ui/WhatsInThisViewPanel.kt
```diff
@@ -0,0 +1,107 @@
...
+        panel!!.add(notConnectedCardWrapper, NOT_CONNECTED)
+        panel!!.add(connectedCardWrapper, CONNECTED)
+        layout!!.show(panel, NOT_CONNECTED)
+    }
+
+    private fun switchCards() {
+        ApplicationManager.getApplication().assertIsDispatchThread()
+
+        // Checking connected mode state may take time, so lets move from EDT to pooled thread
+        runOnPooledThread(project!!) {
```
damien-urruty-sonarsource, line 91: There are several places where you use the `!!` operator, do you need it? `project` is not nullable

src/main/java/org/sonarlint/intellij/ui/WhatsInThisViewPanel.kt
```diff
@@ -0,0 +1,107 @@
...
+        createPanel()
+        switchCards()
+        CurrentFileStatusPanel.subscribeToEventsThatAffectCurrentFile(project) { this.switchCards() }
+        project.messageBus.connect(project).subscribe<VcsListener>(
+            VcsListener.TOPIC,
+            VcsListener { _: Module?, _: String? ->
+                runOnUiThread(
+                    project
+                ) { this.switchCards() }
+            })
```
damien-urruty-sonarsource, line 60: You don't need this part here, as VCS events don't influence the content of the panel

src/main/java/org/sonarlint/intellij/ui/WhatsInThisViewPanel.kt
```diff
@@ -0,0 +1,107 @@
...
+import javax.swing.JPanel
+import org.sonarlint.intellij.common.util.SonarLintUtils
+import org.sonarlint.intellij.common.vcs.VcsListener
+import org.sonarlint.intellij.core.ProjectBindingManager
+import org.sonarlint.intellij.ui.UiUtils.Companion.runOnUiThread
+import org.sonarlint.intellij.util.HelpLabelUtils.Companion.createCurrentFileHelp
+import org.sonarlint.intellij.util.HelpLabelUtils.Companion.createCurrentFileHelpNotConnected
+import org.sonarlint.intellij.util.runOnPooledThread
+
+class WhatsInThisViewPanel(project: Project) {
```
damien-urruty-sonarsource, line 39: You could use val here and avoid having to declare a var line 42

damien-urruty-sonarsource (commented):
Another round of comments, hopefully the last one :)

damien-urruty-sonarsource (approved):

PR #886, opened 2023-12-06
SLI-1203 Fix LiveFindingCache deadlock

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/finding/persistence/LiveFindingCache.java
```diff
@@ -155,12 +152,12 @@ private void flushAll() {
...
    */
-  public synchronized void clear() {
+  public void clear() {
     persistence.clear();
     cache.clear();
   }
 
-  public synchronized boolean contains(VirtualFile virtualFile) {
+  public boolean contains(VirtualFile virtualFile) {
     return getLive(virtualFile) != null;
```
henryju, line 161: Optimization, in the `getLive()` method we are creating a new ArrayList, while we don't actually use it here.

src/main/java/org/sonarlint/intellij/finding/persistence/LiveFindingCache.java
```diff
@@ -119,26 +120,22 @@ public Collection<Trackable> getPreviousFindings(VirtualFile file) {
...
-    cache.computeIfAbsent(virtualFile, f -> new ArrayList<>()).add(finding);
-  }
-
   /**
-   * Flushes all cached entries to disk.
+   * Flushes all provided entries to disk.
    * It does not clear the cache.
    */
-  private void flushAll() {
+  private void flushAll(Map<VirtualFile, Collection<T>> newFindingsPerFile) {
```
henryju, line 136: Before, this method was never called concurrently, because of the `synchronized` on the `replaceFindings` method. Now it can be called concurrently. Is it safe? Is `persistence.save(key, liveFindings);` thread safe?

src/main/java/org/sonarlint/intellij/finding/persistence/LiveFindingCache.java
damien-urruty-sonarsource: Fixed

src/main/java/org/sonarlint/intellij/finding/persistence/LiveFindingCache.java
damien-urruty-sonarsource: I pass the map as a parameter to avoid iterating over the cache, which could cause concurrency issues.
About the persistence, in fact we already have synchronization on `persistence.save(key, liveFindings)` but it might be a problem as it contains a read action. I will rework that

src/main/java/org/sonarlint/intellij/finding/persistence/LiveFindingCache.java
```diff
@@ -155,13 +152,13 @@ private void flushAll() {
...
-  public synchronized void clear() {
+  public void clear() {
     persistence.clear();
     cache.clear();
   }
 
-  public synchronized boolean contains(VirtualFile virtualFile) {
-    return getLive(virtualFile) != null;
+  public boolean contains(VirtualFile virtualFile) {
+    return cache.get(virtualFile) != null;
```
henryju, line 161:
```suggestion
    return cache.containsKey(virtualFile);
```
maybe?

henryju (approved):
Minor comment then LGTM!

PR #889, opened 2023-12-11
SLI-1198 Provide in context notification for advanced rules

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/notifications/SonarLintProjectNotifications.java
```diff
@@ -30,21 +30,27 @@
...
 import org.sonarsource.sonarlint.core.clientapi.client.smartnotification.ShowSmartNotificationParams;
 import org.sonarsource.sonarlint.core.commons.log.ClientLogOutput;
 
 import static org.sonarlint.intellij.common.util.SonarLintUtils.getService;
 import static org.sonarlint.intellij.documentation.SonarLintDocumentation.Intellij.CONNECTED_MODE_LINK;
+import static org.sonarlint.intellij.documentation.SonarLintDocumentation.Marketing.CONNECTED_MODE_BENEFITS_LINK;
 
 @Service(Service.Level.PROJECT)
 public final class SonarLintProjectNotifications {
+  private static final NotificationGroup IN_CONTEXT_AD_GROUP = NotificationGroupManager.getInstance()
```
damien-urruty-sonarsource, line 52: Field name still mentioning "AD"

src/main/java/org/sonarlint/intellij/promotion/DontAskAgainAction.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.intellij.openapi.actionSystem.AnActionEvent
+import org.sonarlint.intellij.actions.AbstractSonarAction
+import org.sonarlint.intellij.config.Settings
+
+class DontAskAgainAction(text: String, description: String?) : AbstractSonarAction(
+    text, description, null
+) {
+
+    override fun actionPerformed(e: AnActionEvent) {
+        Settings.getGlobalSettings().isPromotionDisabled = true;
```
damien-urruty-sonarsource, line 31: Shouldn't we also expire the notification?

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -0,0 +1,157 @@
...
+
+@Service(Service.Level.PROJECT)
+class PromotionProvider(private val project: Project) {
+
+    private val pattern = "hh:mm:ss a, EEE M/d/uuuu"
+    private val dateTimeFormatter = DateTimeFormatter.ofPattern(pattern)
+
+    private val promotedAdvancedLanguages = listOf(Language.JAVA, Language.PYTHON, Language.PHP, Language.JS, Language.TS, Language.CS)
+
+    private val promotedLanguages = listOf(Language.SCALA, Language.PLSQL, Language.SWIFT)
```
damien-urruty-sonarsource, line 43: The names are very close. Maybe `languagesHavingAdvancedRules` and `extraLanguagesInConnectedMode` would be clearer? By the way the second one should not be a static list, PLSQL support depends on the IDE

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -0,0 +1,157 @@
...
+import java.time.temporal.ChronoUnit
+import org.sonarlint.intellij.common.util.SonarLintUtils
+import org.sonarlint.intellij.config.Settings
+import org.sonarlint.intellij.notifications.SonarLintProjectNotifications
+import org.sonarsource.sonarlint.core.commons.Language
+
+@Service(Service.Level.PROJECT)
+class PromotionProvider(private val project: Project) {
+
+    private val pattern = "hh:mm:ss a, EEE M/d/uuuu"
```
damien-urruty-sonarsource, line 38: Instead of defining your own pattern, it would be slightly simpler to store the date as a `Long` by using `Instant.toEpochMilli()`

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -0,0 +1,157 @@
...
+
+        if (lastPromotionDate == null) {
+            return true
+        } else {
+            val dateDiff = ChronoUnit.SECONDS.between(lastPromotionDate, LocalDateTime.now())
+
+            if (dateDiff > PROMOTION_PERIOD) {
+                return true
+            }
+        }
```
damien-urruty-sonarsource, line 149: I think this check is wrong, we are checking SECONDS which I think is wrong. You should have a look at a previous comment I made, `Instant` has a `isBefore()` method, and the code would read better

src/main/java/org/sonarlint/intellij/core/EmbeddedPlugins.kt
```diff
@@ -91,7 +92,7 @@ object EmbeddedPlugins {
     val extraEnabledLanguagesInConnectedMode: Set<Language>
         get() {
             val extraEnabledLanguages = EnumSet.copyOf(ADDITIONAL_ENABLED_LANGUAGES_IN_CONNECTED_MODE)
-            amendEnabledLanguages(extraEnabledLanguages, true)
+            PLSQLLanguageActivator().amendLanguages(extraEnabledLanguages, true)
```
damien-urruty-sonarsource, line 95: What is the reason of this change?

src/main/java/org/sonarlint/intellij/notifications/SonarLintProjectNotifications.java
```diff
@@ -113,12 +113,25 @@ public void notifyWiderLanguageSupport(String language) {
...
+  public void notifyAdvancedLanguageSupport(String language) {
+    var notification = IN_CONTEXT_PROMOTION_GROUP.createNotification(
+      TITLE_SONARLINT_SUGGESTIONS,
+      "Enable advanced " + language + " rules by connecting your project",
+      NotificationType.WARNING);
+    addNotifications(notification);
+    notification.notify(myProject);
+  }
+
+  private static void addNotifications(Notification notification) {
```
damien-urruty-sonarsource, line 129: Maybe `addActions` would be a better name?

src/main/java/org/sonarlint/intellij/notifications/SonarLintProjectNotifications.java
```diff
@@ -113,12 +113,25 @@ public void notifyWiderLanguageSupport(String language) {
...
       NotificationType.WARNING);
+    addNotifications(notification);
+    notification.notify(myProject);
+  }
+
+  public void notifyAdvancedLanguageSupport(String language) {
+    var notification = IN_CONTEXT_PROMOTION_GROUP.createNotification(
+      TITLE_SONARLINT_SUGGESTIONS,
+      "Enable advanced " + language + " rules by connecting your project",
+      NotificationType.WARNING);
```
damien-urruty-sonarsource, line 124: I missed it in the first PR but I think a INFO type would be more adapted for those notifications

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -26,49 +26,82 @@ import com.intellij.openapi.project.Project
...
 
-    private fun findLanguage(extension: String) = extraEnabledLanguagesInConnectedMode.find {
-        it.defaultFileSuffixes.any { suffix ->
-            suffix.equals(extension) || suffix.equals(".$extension")
+    private fun showAdvancedLanguagePromotion(notifications: SonarLintProjectNotifications, language: String) {
+        val lastModifiedDate: Instant? = getLastModifiedDate()
+
+        if (shouldNotify(lastModifiedDate)) {
+            notifications.notifyAdvancedLanguageSupport(language)
+            Settings.getGlobalSettings().lastPromotionNotificationDate = Instant.now().toEpochMilli()
```
damien-urruty-sonarsource, line 100: I fear that we will end up duplicating the same logic for every type of notification, maybe there is a way to factorize some parts?

damien-urruty-sonarsource (approved):

src/main/java/org/sonarlint/intellij/notifications/SonarLintProjectNotifications.java
```diff
@@ -108,17 +108,21 @@ public void notifyProjectStorageInvalid() {
...
       TITLE_SONARLINT_SUGGESTIONS,
-      "Enable " + language +" analysis by connecting your project",
-      NotificationType.WARNING);
+      content,
+      NotificationType.INFORMATION);
+    addActions(notification);
+    notification.notify(myProject);
+  }
+
+  private static void addActions(Notification notification) {
```
damien-urruty-sonarsource, line 120: This separate method makes less sense now as it will be called only once. You can also keep it, as you prefer

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -26,53 +26,73 @@ import com.intellij.openapi.project.Project
...
+        Language.JAVA, Language.PYTHON, Language.PHP, Language.JS, Language.TS, Language.CS
+    )
+
     fun subscribeToTriggeringEvents() {
         val busConnection = project.messageBus.connect()
         with(busConnection) {
             subscribe(FileEditorManagerListener.FILE_EDITOR_MANAGER, object : FileEditorManagerListener {
                 override fun fileOpened(source: FileEditorManager, file: VirtualFile) {
+                    val extension = file.extension ?: return
+                    val notifications = getSonarLintProjectNotifications(source)
```
damien-urruty-sonarsource, line 49: You could move this in the if

damien-urruty-sonarsource (approved):

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
eray-felek-sonarsource: I would like to keep it here

PR #890, opened 2023-12-12
SLI-1195 notification for full project analysis

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.java
```diff
@@ -94,6 +95,13 @@ public Cancelable autoAnalyzeFiles(Collection<VirtualFile> files, TriggerType tr
       return null;
     }
     var callback = new UpdateOnTheFlyFindingsCallable(onTheFlyFindingsHolder);
+
+    var lastAnalysisDate = getGlobalSettings().getLastAutoAnalysisDate();
+
+    if (lastAnalysisDate == 0) {
+      getGlobalSettings().setLastAutoAnalysisDate(Instant.now().toEpochMilli());
+    }
```
damien-urruty-sonarsource, line 103: Could we find a way to not modify the AnalysisSubmitter and rely only on the event?

src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.java
```diff
@@ -119,6 +127,7 @@ public void analyzeFilesOnUserAction(Collection<VirtualFile> files, AnActionEven
       callback = new ShowUpdatedCurrentFileCallable(project, onTheFlyFindingsHolder);
     } else {
       callback = new ShowReportCallable(project);
+      getGlobalSettings().setReportUsed(true);
```
damien-urruty-sonarsource, line 130: Same here

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -87,6 +88,22 @@ public boolean isSecretsNeverBeenAnalysed() {
...
+  }
+
+  public long getLastAutoAnalysisDate() {
+    return lastAutoAnalysisDate;
+  }
+
+  public boolean isReportUsed() {
+    return isReportUsed;
+  }
+
```
damien-urruty-sonarsource, line 106: Is there a way to not store those details in the global settings? I think we should stick to user preferences here

damien-urruty-sonarsource (commented):
As we discussed, some concerns about the implementation. We can discuss more about it

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -64,6 +100,22 @@ class PromotionProvider(private val project: Project) {
...
 
+    private fun getSonarLintProjectNotifications(): SonarLintProjectNotifications {
+        return SonarLintUtils.getService(
+            project,
+            SonarLintProjectNotifications::class.java
+        )
+    }
+
+    private fun processFullProjectPromotion(notifications: SonarLintProjectNotifications) {
+        val firstAutoAnalysis = Instant.ofEpochMilli(PropertiesComponent.getInstance().getLong(FIRST_AUTO_ANALYSIS_DATE, 0L))
```
damien-urruty-sonarsource, line 111: We read the value twice in a row (here and just before calling this method), you could get the value only once and check at the same time

damien-urruty-sonarsource (approved):
Small improvement, else LGTM

PR #891, opened 2023-12-14
MMF 3529 Deeper analysis

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/documentation/SonarLintDocumentation.kt
```diff
@@ -34,8 +34,10 @@ object SonarLintDocumentation {
     object Marketing {
         private const val BASE_MARKETING_URL = "https://www.sonarsource.com"
         const val COMPARE_SERVER_PRODUCTS_LINK = "${BASE_MARKETING_URL}/open-source-editions"
+        const val CONNECTED_MODE_BENEFITS_LINK = "${Intellij.BASE_DOCS_URL}/team-features/connected-mode/#benefits"
```
nquinquenel, line 37: It should be in the IntelliJ section

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -112,15 +113,18 @@ class PromotionProvider(private val project: Project) {
         val wasReportEverUsed = PropertiesComponent.getInstance().getBoolean(WAS_REPORT_EVER_USED, false)
 
         if (!wasReportEverUsed && firstAutoAnalysis.isBefore(Instant.now().minus(FULL_PROJECT_PROMOTION_DURATION))) {
-            showPromotion(notifications, "Detect issues in your whole project")
+            UiUtils.Companion.runOnUiThread(project, Runnable {
+                SonarLintUtils.getService(project, PromotionProvider::class.java)
+                    .showPromotion(notifications, "Detect issues in your whole project")
+            })
```
nquinquenel, line 119: You can simplify everything to just
```
            UiUtils.runOnUiThread(project) {
                showPromotion(notifications, "Detect issues in your whole project")
            }
```
And then, I don't necessarily understand what is the need for this to run of the UI thread, what was the issue?

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -0,0 +1,185 @@
...
+
+        if (language != null) {
+            showPromotion(notifications, "Enable " + language.label + " analysis by connecting your project")
+        }
+    }
+
+    private fun findLanguage(extension: String, languages: Set<Language>): Language? {
+        return languages.find {
+            it.defaultFileSuffixes.any { suffix ->
+                suffix.equals(extension) || suffix.equals(".$extension")
```
nquinquenel, line 142: You can replace the `.equals` by `==`

src/main/java/org/sonarlint/intellij/ui/WhatsInThisViewPanel.kt
```diff
@@ -0,0 +1,96 @@
...
+import javax.swing.JPanel
+import org.sonarlint.intellij.common.util.SonarLintUtils
+import org.sonarlint.intellij.core.ProjectBindingManager
+import org.sonarlint.intellij.ui.UiUtils.Companion.runOnUiThread
+import org.sonarlint.intellij.util.HelpLabelUtils.Companion.createCurrentFileHelp
+import org.sonarlint.intellij.util.HelpLabelUtils.Companion.createCurrentFileHelpNotConnected
+import org.sonarlint.intellij.util.runOnPooledThread
+
+class WhatsInThisViewPanel(val project: Project) {
+    private val NOT_CONNECTED = "Not Connected"
```
nquinquenel, line 38: Since those are static variables not meant to be modified, you can put both in a `companion object` block, and add the `const` identifier

src/main/java/org/sonarlint/intellij/util/HelpLabelUtils.kt
```diff
@@ -26,12 +26,26 @@ import org.sonarlint.intellij.telemetry.LinkTelemetry
 
 class HelpLabelUtils {
     companion object {
+        private var currentFileText: String = "SonarLint analyzes in real-time the active file only and reports any issues found in this view. ";
```
nquinquenel, line 29: Do you know if this sentence was reviewed? I'd personally go with:
> SonarLint only analyzes the active file in real-time and reports any issues found in this view.
WDYT?

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -0,0 +1,185 @@
...
+        if (shouldNotify(lastModifiedDate)) {
+            notifications.notifyLanguagePromotion(content)
+            PropertiesComponent.getInstance().setValue(LAST_PROMOTION_NOTIFICATION_DATE, Instant.now().toEpochMilli().toString())
+        }
+    }
+
+    private fun getLastModifiedDate(): Instant? {
+        val instantValue = PropertiesComponent.getInstance().getLong(LAST_PROMOTION_NOTIFICATION_DATE, 0L)
+
+        val lastModifiedDate: Instant? = if (instantValue != 0L) {
```
nquinquenel, line 159: Personal preference but you don't need to put the type in this case

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
eray-felek-sonarsource: There are 2 different notifications can be triggered at the same time from multiple threads and we might end up showing 2 notifications which is something we want to avoid(We have 7 days rule)

src/main/java/org/sonarlint/intellij/util/HelpLabelUtils.kt
eray-felek-sonarsource: Yes, we agreed on this sentence with Marco

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
eray-felek-sonarsource: I prefer .equals

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -41,10 +41,20 @@
...
   private boolean isFocusOnNewCode = false;
+  private boolean isPromotionDisabled = false;
+
+  public boolean isPromotionDisabled() {
+    return isPromotionDisabled;
+  }
+
+  public void setPromotionDisabled(boolean promotionDisabled) {
+    isPromotionDisabled = promotionDisabled;
+  }
```
nquinquenel, line 53: Please put the getter/setter after all the attributes declaration

src/main/java/org/sonarlint/intellij/util/HelpLabelUtils.kt
```diff
@@ -26,12 +26,26 @@ import org.sonarlint.intellij.telemetry.LinkTelemetry
 
 class HelpLabelUtils {
     companion object {
+        private const val CURRENT_FILE_TEXT: String = "SonarLint analyzes in real-time the active file only and reports any issues found in this view. ";
```
nquinquenel, line 29:
```suggestion
        private const val CURRENT_FILE_TEXT: String = "SonarLint analyzes in real-time the active file only and reports any issues found in this view. "
```

nquinquenel (approved):
Minor comments, otherwise LGTM

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
nquinquenel: IMO we should switch to `=`= because it raises a Sonar rule, so I'd change it because we eat our own dog food (unless it really does not make sense), and then because in standalone mode the issue will be raised again 😅

src/main/java/org/sonarlint/intellij/util/HelpLabelUtils.kt
nquinquenel: I think it's not super well written, at least add a comma before the "and" maybe 😄 

> "SonarLint analyzes in real-time the active file only, and reports any issues found in this view. "

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
eray-felek-sonarsource: The reason I prefer using .equals is that some languages have this == for comparing the value(Kotlin) and some languages (Java) have this for comparing the reference. In our codebase we use both languages, and if are in the habit of using == instead of .equals there is a chance that will use it thinking it compares the values of the strings but in fact it would be comparing the references

eray-felek-sonarsource: IMO we should disable this rule

PR #893, opened 2023-12-18
Hardening 10.2

nquinquenel:


src/main/resources/META-INF/plugin.xml
```diff
@@ -64,6 +64,7 @@
 
     <change-notes><![CDATA[
       <ul>
+        <li>10.2 - Add support for Gradle Kotlin DSL. Bug fixes, fewer FPs and improvements for many languages.</li>
```
eray-felek-sonarsource, line 67: We could mention S6785 from Python here.

eray-felek-sonarsource (approved):

src/main/resources/META-INF/plugin.xml
nquinquenel: Why do you think so? S6785 is not new, it's an improvement.

nquinquenel: I added the mention because it's new, thanks

PR #896, opened 2023-12-20
SLI-1199 In context notification for faster project analysis

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -67,7 +67,7 @@ class PromotionProvider(private val project: Project) {
                     val extension = file.extension ?: return
                     val notifications = getSonarLintProjectNotifications(source)
 
-                    if (!Settings.getSettingsFor(project).isBound && !Settings.getGlobalSettings().isPromotionDisabled) {
+                    if (isPromotionEnabled()) {
```
damien-urruty-sonarsource, line 70: It's more readable like this, well-done!

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -78,22 +78,41 @@ class PromotionProvider(private val project: Project) {
...
-                    if (nonReportAnalysisTriggers.contains(triggerType) && !wasAutoAnalyzed) {
-                        PropertiesComponent.getInstance().setValue(FIRST_AUTO_ANALYSIS_DATE, Instant.now().toEpochMilli().toString())
+                    if (nonReportAnalysisTriggers.contains(triggerType)) {
+                        processAutoAnalysisTriggers(wasAutoAnalyzed, notifications)
                     }
 
                     if (reportAnalysisTriggers.contains(triggerType)) {
-                        PropertiesComponent.getInstance().setValue(WAS_REPORT_EVER_USED, true)
+                        processReportAnalysisTriggers(files, notifications)
                     }
```
damien-urruty-sonarsource, line 87: I would expect those 2 `if`s to be moved inside the `isPromotionEnabled()`, I think it would make more sense. Do you see any issue with that?

damien-urruty-sonarsource (approved):

PR #898, opened 2023-12-22
SLI-1201 Provide help in report tab

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/ui/ReportTabStatusPanel.kt
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.sonarlint.intellij.ui
+
+import com.intellij.openapi.project.Project
+import com.intellij.ui.components.JBPanel
+import java.awt.BorderLayout
+
+class ReportTabStatusPanel internal constructor(private val project: Project) : JBPanel<ReportTabStatusPanel>(BorderLayout()) {
+    companion object {
+        private const val HELP_TEXT: String = "This view shows simple issues found on multiple files. "
```
damien-urruty-sonarsource, line 28: You can normally remove the `: String` part, Kotlin should be able to infer the type

damien-urruty-sonarsource, line 28: I think we should also say that it's a "snapshot", so that people know the view won't be updated after the report has been created. Also that issues appear here after a specific action, it's not automatic

src/main/java/org/sonarlint/intellij/ui/ReportTabStatusPanel.kt
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.sonarlint.intellij.ui
+
+import com.intellij.openapi.project.Project
+import com.intellij.ui.components.JBPanel
+import java.awt.BorderLayout
+
+class ReportTabStatusPanel internal constructor(private val project: Project) : JBPanel<ReportTabStatusPanel>(BorderLayout()) {
+    companion object {
+        private const val HELP_TEXT = "This view shows snapshot of simple issues found on multiple files. "
```
damien-urruty-sonarsource, line 28: shows a snapshot

damien-urruty-sonarsource (approved):

PR #901, opened 2023-12-29
SLI-1216 Simplify languages enabling

damien-urruty-sonarsource:


nquinquenel (approved):
Seems good, well done that's much cleaner!

src/main/java/org/sonarlint/intellij/core/EnabledLanguages.kt
```diff
@@ -153,4 +180,6 @@ object EmbeddedPlugins {
     private class EmbeddedPlugin(val pluginKey: String, val name: String, val jarFilePattern: String) {
         constructor(language: Language, name: String, jarFilePattern: String) : this(language.pluginKey, name, jarFilePattern)
     }
+
+    private fun isIdeModuleEnabled(pluginId: String) = PluginManagerCore.getPlugin(PluginId.getId(pluginId))?.isEnabled == true
```
nquinquenel, line 184: I don't think the `== true` is necessary

nquinquenel (approved):
👍🏻

src/main/java/org/sonarlint/intellij/core/EnabledLanguages.kt
nquinquenel: nevermind it is!

PR #907, opened 2024-01-19
SLI-1231 Use auth to download sonarlint-omnisharp distributions

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/core/SonarLintEngineFactory.java
```diff
@@ -143,7 +143,7 @@ private static Map<String, String> prepareExtraProps() {
 
   private static void addOmnisharpServerPaths(SonarLintPlugin plugin, Map<String, String> extraProps) {
     extraProps.put("sonar.cs.internal.omnisharpMonoLocation", plugin.getPath().resolve("omnisharp/mono").toString());
-    extraProps.put("sonar.cs.internal.omnisharpWinLocation", plugin.getPath().resolve("omnisharp/win").toString());
+    extraProps.put("sonar.cs.internal.omnisharpWinLocation", plugin.getPath().resolve("omnisharp/net472").toString());
```
nquinquenel, line 146: Interesting, what was the impact of this?

build.gradle.kts
```diff
@@ -69,11 +67,6 @@ allprojects {
                 }
             }
         }
-        mavenCentral {
```
nquinquenel, line 72: So we don't rely on mavenCentral at all anymore? Repox is sufficient?

nquinquenel (approved):
Looks super good!

build.gradle.kts
damien-urruty-sonarsource: Yes. All the dependencies we rely on today are on Repox

src/main/java/org/sonarlint/intellij/core/SonarLintEngineFactory.java
damien-urruty-sonarsource: I just changed the folder name where the omnisharp files are deployed to. Before it was `win`, now I keep the name of the artifact classifier, which is `net472`

PR #910, opened 2024-01-22
SLI-1232 Pass icon as optional in notification options

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/ui/vulnerabilities/TaintVulnerabilitiesPanel.kt
```diff
@@ -148,7 +149,7 @@ class TaintVulnerabilitiesPanel(private val project: Project) : SimpleToolWindow
             RefreshTaintVulnerabilitiesAction(),
             sonarLintActions.includeResolvedTaintVulnerabilitiesAction(),
             sonarLintActions.configure(),
-            OpenInBrowserAction(LEARN_MORE, "Learn more about taint vulnerabilities in SonarLint", TAINT_VULNERABILITIES_LINK)
+            OpenInBrowserAction(LEARN_MORE, "Learn more about taint vulnerabilities in SonarLint", TAINT_VULNERABILITIES_LINK, AllIcons.Actions.Help)
```
nquinquenel, line 152: You can remove this icon as well, we don't want it anymore in any places

src/main/java/org/sonarlint/intellij/ui/SecurityHotspotsPanel.java
```diff
@@ -139,7 +140,7 @@ private static DefaultActionGroup createActionGroup() {
     actionGroup.add(sonarLintActions.filterSecurityHotspots());
     actionGroup.add(sonarLintActions.includeResolvedHotspotAction());
     actionGroup.add(sonarLintActions.configure());
-    actionGroup.add(new OpenInBrowserAction("Learn More", "Learn more about Security Hotspots in SonarLint", SECURITY_HOTSPOTS_LINK));
+    actionGroup.add(new OpenInBrowserAction("Learn More", "Learn more about Security Hotspots in SonarLint", SECURITY_HOTSPOTS_LINK, AllIcons.Actions.Help));
```
nquinquenel, line 143: same

src/main/java/org/sonarlint/intellij/actions/OpenInBrowserAction.kt
```diff
@@ -19,12 +19,12 @@
...
 package org.sonarlint.intellij.actions
 
-import com.intellij.icons.AllIcons
 import com.intellij.ide.BrowserUtil
 import com.intellij.openapi.actionSystem.AnActionEvent
+import javax.swing.Icon
 
-class OpenInBrowserAction(text: String, description: String?, private val link: String) : AbstractSonarAction(
-    text, description, AllIcons.Actions.Help
+class OpenInBrowserAction(text: String, description: String?, private val link: String, icon: Icon?) : AbstractSonarAction(
```
nquinquenel, line 26: You could put a default value `, icon: Icon? = null` so you don't have to add `null` everywhere

src/main/java/org/sonarlint/intellij/ui/vulnerabilities/TaintVulnerabilitiesPanel.kt
eray-felek-sonarsource: It is displayed in Taint Vulnerabilities tab not in notification. I think we should keep it

src/main/java/org/sonarlint/intellij/ui/SecurityHotspotsPanel.java
eray-felek-sonarsource: same for Security Hotspot

src/main/java/org/sonarlint/intellij/ui/vulnerabilities/TaintVulnerabilitiesPanel.kt
nquinquenel: You're right, good point!

nquinquenel (approved):

PR #913, opened 2024-01-24
Feature/ef/analyzers updates 10.3

eray-felek-sonarsource:


src/main/resources/META-INF/plugin.xml
```diff
@@ -64,6 +64,7 @@
 
     <change-notes><![CDATA[
       <ul>
+        <li>10.3 - 9 new C++ MISRA 2023 rules. Remove support for Node.JS 14. Support TypeScript 5.3. 5 new rules for JavaScript</li>
```
nquinquenel, line 67:
```suggestion
        <li>10.3 - 9 new C++ MISRA 2023 rules. Remove support for Node.JS 14. Support TypeScript 5.3. 5 new rules for JavaScript.</li>
```

build.gradle.kts
```diff
@@ -1,15 +1,15 @@
 
 import com.jetbrains.plugin.blockmap.core.BlockMap
 import groovy.lang.GroovyObject
+import org.jetbrains.intellij.tasks.RunPluginVerifierTask
+import org.jfrog.gradle.plugin.artifactory.dsl.PublisherConfig
```
nquinquenel, line 5: I think it's supposed to be ordered alphabetically, do you use the right codestyle?

src/main/resources/META-INF/plugin.xml
nquinquenel, line 67: It could be worth mentioning that we now support rules for the Jakarta namespace, it was reported as an issue for some users

src/main/resources/META-INF/plugin.xml
```diff
@@ -64,6 +64,7 @@
 
     <change-notes><![CDATA[
       <ul>
+        <li>10.3 - Support Jakarta namespace. 9 new C++ MISRA 2023 rules. Remove support for Node.JS 14. Support TypeScript 5.3. 5 new rules for JavaScript.</li>
```
damien-urruty-sonarsource, line 67: About:
> Remove support for Node.JS 14

I have several questions:
* last time SonarJS dropped support for a version, they also deprecated the next one (you can see below in the release notes). Is it also the case here?
* What is the minimal version now? I think we have it documented in several places, including the [docs website](https://docs.sonarsource.com/sonarlint/intellij/getting-started/requirements/#jstscss-analysis), so you might need to check with Kevin

nquinquenel (approved):

PR #914, opened 2024-01-25
SLI-1235 Do not use unsafe VFS recursion

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/fs/VirtualFileSystemListener.kt
```diff
@@ -46,7 +48,7 @@ import org.sonarsource.sonarlint.plugin.api.module.file.ModuleFileEvent
  * The BulkFileListener is not tied to a specific project but global to the IDE instance
  */
 class VirtualFileSystemListener(
-    private val fileEventsNotifier: ModuleFileEventsNotifier = ModuleFileEventsNotifier()
+    private val fileEventsNotifier: ModuleFileEventsNotifier = ModuleFileEventsNotifier(),
```
nquinquenel, line 51: Is it from the sonar code style?

nquinquenel (approved):
LGTM

src/main/java/org/sonarlint/intellij/fs/VirtualFileSystemListener.kt
damien-urruty-sonarsource: Yes
image

nquinquenel: I have the same but it does not add this comma 🤔

PR #918, opened 2024-01-29
Hardening 10.3 last round

damien-urruty-sonarsource:


eray-felek-sonarsource:
We could mention there is a new rule for HTML:
 S5247

eray-felek-sonarsource (commented):
1 comment about HTML update else LGTM!

damien-urruty-sonarsource:
@eray-felek-sonarsource This rule is a security hotspot, we usually don't mention them as they don't work in standalone mode

eray-felek-sonarsource (approved):

PR #927, opened 2024-02-05
SLI-1244 SLI-1194 Inform users on local ruleset and file exclusions

nquinquenel:


damien-urruty-sonarsource (commented):
The code changes look good, I only have one comment about the messages. The title of the MMF is "Inform users for better alternatives" and I don't find this aspect in the wording. 

For example, the current: 
> When a project is connected to SonarQube or SonarCloud, configuration from the server applies.

just describes a fact, which is useful I agree, but not the advantage that connected mode brings in this situation, e.g. sharing rules configuration with your team, saving it in a more persistent way, ... I think it's part of the acceptance criteria to insist on that part: 

> SonarLint [...] should provide the user with the following information around those lines:
> Short version: Sonar offers a more efficient way to centrally configure once your ruleset [...]

Maybe we should be talking more about the benefits?

nquinquenel:
> Maybe we should be talking more about the benefits?

Agree, I added more information

damien-urruty-sonarsource (approved):

PR #928, opened 2024-02-07
SLI-1255 Updating a rule should correctly refresh the UI

nquinquenel:


damien-urruty-sonarsource (approved):
LGTM, tested and it works way better

PR #931, opened 2024-02-08
SLI-1249 Declare a config-file in optional dependencies declarations

eray-felek-sonarsource:


src/main/resources/META-INF/plugin-go.xml
```diff
@@ -0,0 +1,24 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+    Lesser General Public License for more details.
+
+    You should have received a copy of the GNU Lesser General Public
+    License along with this program; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+
+-->
+<open>
+</open>
```
nquinquenel, line 24: Do we actually need the `<xml>` and `<open>` tags? Have you tried with just an empty XML file with the license header?

nquinquenel (approved):

nquinquenel (commented):
In the other plugin-*.xml we have, it's only

```
<idea-plugin>
</idea-plugin>
```

Do we need the `<xml>` tag and the `xmlns:xi` field?

eray-felek-sonarsource:
> In the other plugin-*.xml we have, it's only
> 
> ```
> <idea-plugin>
> </idea-plugin>
> ```
> 
> Do we need the `<xml>` tag and the `xmlns:xi` field?

Updated file to not include these tags

PR #932, opened 2024-02-08
SLI-1257 Pop-up "Too complex, sorry" should not appear

nquinquenel:


damien-urruty-sonarsource (approved):
I think this works around the problem, which is already a good start!

I also think that we should take a step back and think about this "invalid binding" case. When does it happen? Is throwing an error the best way for SonarLint to react to this case?

From my experience, it can happen when a connection is removed, and a project bound to this connection is not opened at that time. Opening the project then causes the invalid binding error. In this case, if the intention of the user was to drop the connection, we could maybe simply drop the binding when the project is opened, and notify the user? We could at least be more constructive than just saying "the binding is invalid" without call-to-action, and instead suggest options e.g. to drop the binding or re-create a connection.

There might be other cases when this case happens, maybe after an update? All this to say, it might be good going back to the root cause, I'm sure there is room for improvement

nquinquenel:
> From my experience, it can happen when a connection is removed, and a project bound to this connection is not opened at that time. Opening the project then causes the invalid binding error.

This is effectively the way I reproduce it.

> I also think that we should take a step back and think about this "invalid binding" case. When does it happen? Is throwing an error the best way for SonarLint to react to this case?

I agree, that's the first solution I described in the ticket, ideally fixing the root cause issue would be best. Even if this pop-up does not appear anymore, analyses are still blocked because the connection is not there, plus some notification errors appear while refreshing taints, and so on. I will investigate this approach, having this fix as a backup.

PR #934, opened 2024-02-12
SLI-1151 Do not analyse Razor files in Connected Mode

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/analysis/LocalFileExclusions.java
```diff
@@ -167,6 +169,18 @@ public Map<Module, Collection<VirtualFile>> retainNonExcludedFilesByModules(Coll
     return filesByModule;
   }
 
+  private Collection<VirtualFile> excludeRazorFiles(Collection<VirtualFile> files) {
+    return files.stream().filter(this::checkIfNotRazorFile).collect(Collectors.toSet());
+  }
+
+  public boolean checkIfNotRazorFile(VirtualFile file){
+    if(file.getExtension() == null) {
```
nquinquenel, line 177:
```suggestion
    if (file.getExtension() == null) {
```

nquinquenel (commented):
We put the exclusion checks inside the method `defaultExclusionCheckers`, I think it would also make sense to use the method `excludeUnsupportedFileOrFileType` for this type of check.

Another point, the ticket mentions that it should be excluded in Connected Mode, I think in the current implementation it will be ignored even outside of Connected Mode?

src/main/java/org/sonarlint/intellij/analysis/LocalFileExclusions.java
```diff
@@ -221,12 +221,21 @@ private ExcludeResult checkProjectStructureExclusion(VirtualFile file) {
...
     }
     return ExcludeResult.notExcluded();
   }
 
+  public static boolean isRazorFile(VirtualFile file) {
+    if (file.getExtension() == null) {
+      return false;
+    }
+
+    return file.getExtension().equals("razor");
```
nquinquenel, line 236: can be simplified to:
`return file.getExtension() != null && file.getExtension().equals("razor");`

nquinquenel (approved):

PR #935, opened 2024-02-13
SLI-1258 Adjust notification condition for Connected Mode promotions

eray-felek-sonarsource:


nquinquenel (approved):
LGTM

Nitpick: use static import as it is quite verbose with all the `Settings.`

PR #936, opened 2024-02-15
SLI-1262 Add ITs for SonarCloud

nquinquenel:


eray-felek-sonarsource (commented):
We can gather the commits into 1 commit

nquinquenel:
> We can gather the commits into 1 commit

This will be done when merging, I'll squash before

eray-felek-sonarsource (approved):

PR #950, opened 2024-02-21
Dependencies update

nquinquenel:


build.gradle.kts
```diff
@@ -11,18 +11,19 @@ import java.util.zip.ZipOutputStream
 import org.jetbrains.intellij.tasks.RunPluginVerifierTask
 import org.jfrog.gradle.plugin.artifactory.dsl.PublisherConfig
 
+@Suppress("DSL_SCOPE_VIOLATION")
 plugins {
-    kotlin("jvm") version "1.8.10"
-    id("org.jetbrains.intellij") version "1.15.0"
-    id("org.sonarqube") version "4.4.1.3373"
+    kotlin("jvm") version "1.9.22"
```
damien-urruty-sonarsource, line 16: Out of curiosity, why is this one hardcoded?

build.gradle.kts
```diff
@@ -11,18 +11,19 @@ import java.util.zip.ZipOutputStream
 import org.jetbrains.intellij.tasks.RunPluginVerifierTask
 import org.jfrog.gradle.plugin.artifactory.dsl.PublisherConfig
 
+@Suppress("DSL_SCOPE_VIOLATION")
```
damien-urruty-sonarsource, line 14: Why do we need this?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,78 @@
...
+sonar-php = "3.33.0.11274"
+sonar-python = "4.14.0.14263"
+sonar-kotlin = "2.20.0.4382"
+sonar-slang = "1.15.0.4655"
+sonar-html = "3.13.0.4821"
+sonar-xml = "2.10.0.4108"
+sonar-omnisharp = "1.18.0.83276"
+sonar-text = "2.8.0.1635"
+sonar-iac = "1.26.0.8471"
+sonar-cpp = "6.52.0.66181"
```
damien-urruty-sonarsource, line 30: Small detail: for commodity, as this block is what probably changes often, I would move it closer to the top

damien-urruty-sonarsource (approved):
Minor comments, else LGTM!

build.gradle.kts
nquinquenel: Right, I forgot it's just the ID `org.jetbrains.kotlin.jvm` behind

PR #955, opened 2024-02-23
SLI-1291 Allow external contributors to build SLI

nquinquenel:
It seems that our omnisharp-roslyn is not pushed on maven central, so I avoid the download of this dependency if user has not defined credentials to Repox

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -68,6 +68,12 @@ allprojects {
                 }
             }
         }
+        mavenCentral {
+            content {
+                // avoid dependency confusion
+                excludeGroupByRegex("com\\.sonarsource.*")
+            }
+        }
     }
 
     java {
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -217,11 +223,11 @@ dependencies {
     "sqplugins"(libs.bundles.sonar.analyzers)
     if (artifactoryUsername.isNotEmpty() && artifactoryPassword.isNotEmpty()) {
         "sqplugins"(libs.sonar.cfamily)
+        "omnisharp"("org.sonarsource.sonarlint.omnisharp:omnisharp-roslyn:$omnisharpVersion:mono@zip")
+        "omnisharp"("org.sonarsource.sonarlint.omnisharp:omnisharp-roslyn:$omnisharpVersion:net472@zip")
+        "omnisharp"("org.sonarsource.sonarlint.omnisharp:omnisharp-roslyn:$omnisharpVersion:net6@zip")
```
damien-urruty-sonarsource, line 228: Could we use catalog versions for those artifacts?

damien-urruty-sonarsource (approved):
Small comment, else LGTM!

build.gradle.kts
nquinquenel: I would need to concatenate the version to the module name in order to get -> `org.sonarsource.sonarlint.omnisharp:omnisharp-roslyn:$omnisharpVersion` and the version being `mono@zip`, but since TOML does not allow String interpolation, I cannot do that

I couldn't find a good way to do this so I kept it there

damien-urruty-sonarsource: `mono@zip` is not the version, but the `classifier` and the `extension`, and indeed there is no way to specify them in the toml file ([source](https://github.com/gradle/gradle/issues/16665) and [source](https://stackoverflow.com/questions/71485996/how-to-specify-classifier-when-using-a-gradle-version-catalog)). That said the version could still be defined in the toml file, but then specific syntax needs to be used in the build.gradle, as shown in the second link. So maybe simpler to keep it like this

PR #957, opened 2024-02-26
SLI-1298 Fix open in IDE feature in Windows

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -123,6 +111,19 @@ import org.sonarsource.sonarlint.core.rpc.protocol.common.FlowDto
...
+import java.net.Proxy
+import java.net.URI
+import java.net.URL
+import java.nio.file.Path
+import java.nio.file.Paths
+import java.security.cert.CertificateException
+import java.security.cert.CertificateFactory
+import java.security.cert.X509Certificate
+import java.util.UUID
+import java.util.concurrent.CancellationException
```
damien-urruty-sonarsource, line 126: As discussed, probably a wrong configuration, those imports should not move

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -270,7 +271,7 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
         if (!project.isDisposed) {
             SonarLintProjectNotifications.get(project).expireCurrentFindingNotificationIfNeeded()
         }
-        val file = tryFindFile(project, filePath.toString())
+        val file = tryFindFile(project, filePath.systemIndependentPath)
```
damien-urruty-sonarsource, line 274: Please see my comment on the other PR, I think we could try to call this method at a lower place

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
nquinquenel: It depends, this is also a point that was raised by Julien at some point. In kotlin, the configuration for imports puts `java.*` and `javax.*` package separately. In Java it is ordered alphabetically, in Kotlin they are put at the bottom. It depends on how we want to align, I think Eray has the default ordering (java package are separated).

nquinquenel: I have the same but I think the default one is this:
image

damien-urruty-sonarsource (approved):

PR #958, opened 2024-02-26
SLI-1263 Internal dogfooding is not available anymore

nquinquenel:


src/main/java/org/sonarlint/intellij/dogfood/DogfoodPluginRepositoryAuthProvider.kt
```diff
@@ -0,0 +1,100 @@
...
+                    with(testUrlRepox.openConnection() as HttpURLConnection) {
+                        requestMethod = "GET"
+                        setRequestProperty("Authorization", encodedAuth)
+                        connectTimeout = 10_000
+                        connect()
+
+                        if (responseCode == 401) {
+                            SonarLintProjectNotifications.projectLessNotification(
+                                "Dogfooding credentials are not valid",
+                                "Connection to Repox was unauthorized, make sure the credentials 'username' and 'password' are valid in ~/.sonarlint/dogfood.properties" +
```
damien-urruty-sonarsource, line 69: Maybe add "and restart the IDE" here too? Since we read the properties file only once, users will have to restart

damien-urruty-sonarsource (approved):
I think that can work. We will need to update the Dogfooding instructions on the Xtranet, and somehow communicate to dogfooders they will have to do something. 

I still think discussing with REs could be useful, they might have some other inputs

src/main/java/org/sonarlint/intellij/dogfood/DogfoodPluginRepositoryAuthProvider.kt
nquinquenel: It's already on the next line

PR #960, opened 2024-02-27
SLI-1301 Fix Taint Vulnerability Display on Windows

eray-felek-sonarsource:


damien-urruty-sonarsource (commented):
I think we could try to centralize this conversion instead of calling `systemIndependentPath` from several places. We could try to keep `Path` instances all the way except at the very end of the calls chain.

eray-felek-sonarsource:
> @eray-felek-sonarsource is this PR still necessary, or dit you fix it in the other PR?

Yes, it is valid. Turns. It also led me to another issue regarding server-sent events for security hotspots so I included that fix here as well.

damien-urruty-sonarsource (commented):
I have the impression that the code you change in `SonarLintIntelliJClient` and `TaintVulnerabilityMatcher` are pretty similar, and also there are some similarities with the code you changed yesterday. Would it be worth extracting this logic in a common location?

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -460,7 +461,7 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
 
             ProjectRootManager.getInstance(project).contentRoots.forEach {
                 if (it.isDirectory) {
-                    val matchedFile = it.findFileByRelativePath(filePath.toString())
+                    val matchedFile = findByRelativePath(it, filePath)
```
damien-urruty-sonarsource, line 464: Sorry to annoy you again, but I think we could also use `ProjectUtils.tryFindFile` here

src/main/java/org/sonarlint/intellij/util/SonarLintAppUtils.java
```diff
@@ -128,6 +130,15 @@ private static String getPathRelativeToProjectBaseDir(Project project, VirtualFi
     return VfsUtilCore.getRelativePath(file, projectDir);
   }
 
+  @CheckForNull
+  public static VirtualFile findByRelativePath(VirtualFile file, Path path) {
+    return file.findFileByRelativePath(getSystemIndependentPath(path));
+  }
+
+  public static String getSystemIndependentPath(Path filePath) {
```
damien-urruty-sonarsource, line 138: If possible, I would try to make this one private, to not encourage people to use it

damien-urruty-sonarsource (approved):

PR #969, opened 2024-03-04
Log when RPC requests are canceled instead of throwing an exception

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -679,34 +685,38 @@ class BackendService @NonInjectable constructor(private var backend: Sloop) : Di
...
+                        } else {
+                            val localOnlyIssue = serverOrLocalHotspot.right
+                            liveHotspot.backendId = localOnlyIssue.id
+                            liveHotspot.setOnNewCode(true)
+                        }
                     }
                 }
             }
+                .join()
+        } catch(e: CancellationException) {
```
nquinquenel, line 717:
```suggestion
        } catch (e: CancellationException) {
```

nquinquenel (approved):
LGTM

PR #972, opened 2024-03-05
SLI-1311 Parallelize Tests

eray-felek-sonarsource:


.cirrus.yml
```diff
@@ -46,7 +46,7 @@ eks_builder_container: &BUILDER_CONTAINER_DEFINITION
     JDK_VERSION: ${JDK_VERSION}
   builder_role: cirrus-builder
   builder_image: docker-builder-v*
-  builder_instance_type: t3.small
+  builder_instance_type: t3.2xlarge
```
thahnen, line 49: Hi @eray-felek-sonarsource,

FYI this does not change the machine type. This is the configuration just for the "external" machine, that is building the docker image for this very pipeline. It is not used for the pipeline itself.

For now, the machine type must be configured on the task level (for now), take a look at the pipelines of [SLE](https://github.com/SonarSource/sonarlint-eclipse/blob/10.0.0.81687/.cirrus.default.yml#L74).

This will change in the future once RE changes their logic to only have one (powerful) type of build node used for all pipelines - then we don't have to configure anything on our side anymore.

eray-felek-sonarsource: Hi Tobias, thank you I was wondering why I couldn't see much difference now it makes sense

nquinquenel (commented):
The pipeline is much faster with the improved hardware, but is it really faster by splitting the tests into multiple classes? What takes a lot of time is actually to start the SonarQube server, it would be great to compare the difference strictly on this (without the better machine).

The idea is great but I think it makes the whole ITs a bit messy, and it's also a lot of hard-coded values, so a bit annoying to maintain.

eray-felek-sonarsource:
> The pipeline is much faster with the improved hardware, but is it really faster by splitting the tests into multiple classes? What takes a lot of time is actually to start the SonarQube server, it would be great to compare the difference strictly on this (without the better machine).
> 
> The idea is great but I think it makes the whole ITs a bit messy, and it's also a lot of hard-coded values, so a bit annoying to maintain.

I did not change the hardware for ITs and even though I reversed every hardware change on this branch it is still the same result. 13 minutes(down from 21). I agree it is a bit messy at this point and may be improved. I am open to suggestions. However, I believe that 8 minutes is a substantial time reduction and a good trade-off. For the SonarQube start-up, I think we can also think about solutions for reducing that time but that is a different topic.

.cirrus.yml
```diff
@@ -270,8 +282,15 @@ qa_task:
     ffmpeg -loglevel warning -f x11grab -video_size 1280x960 -i ${DISPLAY} -codec:v libx264 -r 12 ${CIRRUS_WORKING_DIR}/recording_${IDEA_VERSION}.mp4
   run_its_script: |
     echo "Run ITs on ${IDEA_VERSION}"
+    echo "Run ITs on ${QA_CATEGORY}"
     source .cirrus/use-gradle-wrapper.sh
-    gradle :its:check --stacktrace -i -PijVersion=${IDEA_VERSION} -PslPluginDirectory=${CIRRUS_WORKING_DIR}/staged-plugin
+    if [[ ${QA_CATEGORY} = "Idea2022_Standalone" ]] || [[ ${QA_CATEGORY} = "Idea2023_Standalone" ]]; then
```
nquinquenel, line 287: I would check if the name ends by Standalone or Connected2, so you don't have to duplicate everything for each year

.cirrus.yml
```diff
@@ -270,8 +282,15 @@ qa_task:
     ffmpeg -loglevel warning -f x11grab -video_size 1280x960 -i ${DISPLAY} -codec:v libx264 -r 12 ${CIRRUS_WORKING_DIR}/recording_${IDEA_VERSION}.mp4
   run_its_script: |
     echo "Run ITs on ${IDEA_VERSION}"
+    echo "Run ITs on ${QA_CATEGORY}"
```
nquinquenel, line 285: Can be concatenated with the previous `echo`

.cirrus.yml
```diff
@@ -270,8 +282,15 @@ qa_task:
...
     source .cirrus/use-gradle-wrapper.sh
-    gradle :its:check --stacktrace -i -PijVersion=${IDEA_VERSION} -PslPluginDirectory=${CIRRUS_WORKING_DIR}/staged-plugin
+    if [[ ${QA_CATEGORY} = "Idea2022_Standalone" ]] || [[ ${QA_CATEGORY} = "Idea2023_Standalone" ]]; then
+    gradle its:test --tests StandaloneIdeaTests --stacktrace -i -PijVersion=${IDEA_VERSION} -PslPluginDirectory=${CIRRUS_WORKING_DIR}/staged-plugin
```
nquinquenel, line 288: What's the reason to use `test` instead of `check`?

its/src/test/kotlin/org/sonarlint/intellij/its/tests/ConnectedIdeaTests2.kt
```diff
@@ -0,0 +1,210 @@
...
+import org.sonarlint.intellij.its.utils.SonarCloudUtils.Companion.restoreSonarCloudProfile
+import org.sonarqube.ws.client.WsClient
+import org.sonarqube.ws.client.issues.SearchRequest
+import org.sonarqube.ws.client.usertokens.GenerateRequest
+import org.sonarqube.ws.client.usertokens.RevokeRequest
+import kotlin.random.Random
+
+// In order to run these test change the url triggerOpenHotspotRequest to some other port than 64120 depending on number of IntelliJ instances
+@DisabledIf("isCLionOrGoLand")
+class ConnectedIdeaTests2 : BaseUiTest() {
```
nquinquenel, line 47: ConnectedIdeaTests2 is maybe not the best name 😄

nquinquenel (commented):
It works but it'd be great to have a second opinion on this

.cirrus.yml
eray-felek-sonarsource: Check only runs tests by default; if there are other verifications, it also runs them. In our case, we do have other verifications like licenseFormat for ITs. So I am running the check without the tests as a separate task. Since we are running tests parallelly it is better to run and also add new parallel tests by using the test instead of check since we are running them explicitly. Excluding tests from check would become verbose for parallel execution.

.cirrus.yml
```diff
@@ -269,9 +302,15 @@ qa_task:
     echo 'Recording tests on video'
     ffmpeg -loglevel warning -f x11grab -video_size 1280x960 -i ${DISPLAY} -codec:v libx264 -r 12 ${CIRRUS_WORKING_DIR}/recording_${IDEA_VERSION}.mp4
   run_its_script: |
-    echo "Run ITs on ${IDEA_VERSION}"
+    echo "Run ITs on ${IDEA_VERSION} with category: ${QA_CATEGORY}"
     source .cirrus/use-gradle-wrapper.sh
-    gradle :its:check --stacktrace -i -PijVersion=${IDEA_VERSION} -PslPluginDirectory=${CIRRUS_WORKING_DIR}/staged-plugin
```
thahnen, line 274: I'm a bit confused, to be honest. It seems at least from what you changed, that not all integration tests will run anymore except for `StandaloneIdeaTests`, `CurrentFileTests`, and `HotspotVulnerabilityScalaTests` (formerly `ConnectedIdeaTests`). Is that the case?

thahnen (changes requested):
I'll change the review if it is actually the case that the other tests still do run.

.cirrus.yml
eray-felek-sonarsource: Hi Tobias,

Currently we have 6 test Classes:
ClionTests
CurrentFileTests
GoSonarLanguageTestsNode
HotspotVulnerabilityScalaTests
PLSQLTests
StandaloneTests

Parallelization is only applied to tests running on IC.

Clion,PLSQL and Go tests can not be run on IC.
CLion and Go requires Clion and GoLand IDEs
PLSQL requires Databases and SQL plugin which is not present on IC

We are not running any tests on Intellij Ultimate at the moment.

So, in short, we are indeed running all the tests that can be run on IC.

thahnen: We had a quick talk. Due to the bash if statement, on all axes there are not all tests executed anymore which has to be changed back.

eray-felek-sonarsource: Yes, added another elif statement and the following else that executes all tests for the CLion and GoLand IDEs to cover this case.

thahnen: Good addition but the else should then not run `StandaloneIdeaTests`, `HotspotVulnerabilityScalaTests`, `CurrentFileTests`.
I'm only aware that you can exclude test classes via a configuration in the `build.gradle.kts` that is triggered like a Maven profile on the call.

eray-felek-sonarsource: The Else is run for the IDEs Clion and GoLand. We have the annotation @DisabledIf("isCLionOrGoLand") on those classes to skip those tests. In fact, you can verify this on the cirrus CI you will see :(StandaloneIdeaTests > should_exclude_rule_() SKIPPED). I changed the code to use check instead of test to be consistent with previous versions.

nquinquenel (commented):
I speak I, I'm not a big fan of this, the tests and the CI are so tightly coupled that it's really complex to maintain and error-prone. It's also not easy to check whether all the tests are running or if we are missing some (cf previous discussion). I'd be happy if someone could change my mind, but I'd personally avoid merging this, even if the time gained is beneficial.

.cirrus.yml
damien-urruty-sonarsource: You could leverage the matrix here instead of defining if-else clauses. The purpose of the matrix is to create variations, and the list of excluded test suites is the variation here

damien-urruty-sonarsource:
To give another example, you could look at how tests are run in SonarQube: 
* [cirrus file](https://github.com/SonarSource/sonar-enterprise/blob/master/private/.cirrus.yml#L989)
* [gradle file](https://github.com/SonarSource/sonar-enterprise/blob/master/private/it-core/build.gradle#L127)

It is a bit similar, the only thing is that they have some kind of indirection by using labels in the CI file, and connecting labels to test suites to execute in the `build.gradle`. Not a big difference, and the idea behind is the same.

There are mechanisms in JUnit 5 to achieve the same, like [Tags](https://junit.org/junit5/docs/current/user-guide/#running-tests-tags), that could be used to avoid splitting test classes.

Personally, I don't like this kind of approach: 
* it needs extra configuration in the CI
* if we add a new test suite one day, it could be easy to forget to include it in one of the axis
* when adding a new test suite, in which category to include it? What is the impact on build duration? When to introduce a new category?

The ideal solution IMO would be more ambitious:
* have a pool of test runners, so in our case different IDE instances and flavors
* run tests in parallel
* distribute them evenly on those runners

The first one could be achieved at the CI level, and the last 2 ones at JUnit level. When more tests are added or if it's taking too long, we could add more runners to the pool.

Now this is probably a big rework, so in the meantime, the proposed solution could be a tradeoff, to at least speed things up.

its/build.gradle.kts
```diff
@@ -38,7 +38,19 @@ dependencies {
 }
 
 tasks.test {
-    useJUnitPlatform()
+    useJUnitPlatform() {
+        val tag = System.getenv("TEST_SUITE")
+
+        if (tag != null) {
+            if (tag.equals("Suite1")) {
+                excludeTags("Suite2", "Standalone")
```
damien-urruty-sonarsource, line 46: why not using `includeTags(tag)` ?

nquinquenel (approved):
I think it looks much better, with the latest changes it's much simpler and easier to maintain, great job 👍🏻

thahnen (approved):
This looks much smoother than the earlier implementation!

PR #976, opened 2024-03-05
SLI-1312 Disabling rule from action does not work

nquinquenel:


damien-urruty-sonarsource (approved):
So far we tried to keep the `Settings` classes as simple POJOs because they represent the state that is used for serialization/deserialization, with almost no custom logic. Not a blocker, maybe at some point it would be better to have a service coordinating the update of the settings and the backend notification

damien-urruty-sonarsource (approved):

PR #984, opened 2024-03-07
Add logs for backend init failure

nquinquenel:


damien-urruty-sonarsource:
I also added the traces in #987 to cut down on CI build time

PR #985, opened 2024-03-08
SLI-1322 Avoid blocking error while iterating over files

nquinquenel:


src/main/java/org/sonarlint/intellij/util/ProjectUtils.java
```diff
@@ -39,6 +37,8 @@
 import javax.annotation.CheckForNull;
 import org.sonarlint.intellij.finding.TextRangeMatcher;
 
+import static org.sonarlint.intellij.common.ui.ReadActionUtils.computeReadActionSafely;
```
damien-urruty-sonarsource, line 40: Is this added import used?

damien-urruty-sonarsource (approved):
I don't see how those changes prevent the error mentioned in the ticket from happening. Shouldn't we always surround the calls to `iterateContent` with a try-catch to avoid the `Cannot invoke "kotlin.Pair.getFirst()" because "rootsPair" is null`? Else LGTM

src/main/java/org/sonarlint/intellij/util/ProjectUtils.java
nquinquenel: Yes, it's just because the order of the static import moved to the bottom

PR #986, opened 2024-03-08
SLI-1321 Avoid retrieving ID for disposed module

nquinquenel:


damien-urruty-sonarsource:
This has been fixed as part of the rework in #987

PR #987, opened 2024-03-09
SLI-1325 Keep the plugin responsive if the service cannot start

damien-urruty-sonarsource:


src/main/java/org/sonarlint/intellij/ui/tree/TreeSummary.kt
```diff
@@ -20,60 +20,47 @@
...
-    }
-
-    private fun computeEmptyText(): String {
+    private fun computeEmptyText(newCodePeriod: String): String {
         if (isFocusOnNewCode()) {
             return if (holdsOldFindings) {
                 "No older ${treeContentKind.displayName}s"
             } else {
-                "No new ${treeContentKind.displayName}s${getCodePeriod()}"
+                "No new ${treeContentKind.displayName}s$newCodePeriod"
```
nquinquenel, line 63: Shouldn't we put a space before the new code period String? Given that the string `newCodePeriod` can be `"(unknown code period)"` I don't see any whitespace

damien-urruty-sonarsource: It was a regression indeed, fixed

PR #992, opened 2024-03-13
SLI-1330 Prevent lsp4j SEVERE logs to trigger uncaught exceptions

damien-urruty-sonarsource:


nquinquenel (approved):

src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -244,6 +247,15 @@ class BackendService : Disposable {
     }
 
     private fun startSloopProcess(): Sloop {
+        // SLI-1330
+        val lsp4jLogger = Logger.getLogger("org.eclipse.lsp4j.jsonrpc.RemoteEndpoint")
```
henryju, line 251: Be careful, if another IntelliJ plugin (or Intellij itself) is someday using lsp4j, it might redirect logs to the SL console as well.
At some point I tried shading and relocating lsp4j classes in the sonarlint-java-client, but it was difficult.

PR #995, opened 2024-03-14
Fix flaky tests about module analysis

nquinquenel:
My understanding is that the analysis is not necessarily ready yet, so waiting should fix the issue


thahnen (approved):
LGTM! To hell with all those flaky tests ^^

PR #999, opened 2024-03-21
SLI-1333 Add button for export

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -275,7 +289,8 @@ public void actionPerformed(ActionEvent e) {
       WEST, HORIZONTAL, insets, 0, 0));
     bindPanel.add(configureConnectionButton, new GridBagConstraints(2, 3, 1, 1, 0.0, 0.0,
       WEST, HORIZONTAL, insets, 0, 0));
-
+    bindPanel.add(exportConfigurationButton, new GridBagConstraints(2, 2, 1, 1, 0.0, 0.0,
```
nquinquenel, line 292: I think we could add a link to the doc somewhere. It could be "?" next to the button or in the dialog message?

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -385,6 +421,7 @@ public void itemStateChanged(ItemEvent e) {
       searchProjectButton.setEnabled(bound);
       connectionComboBox.setEnabled(bound);
       configureConnectionButton.setEnabled(bound);
+      exportConfigurationButton.setEnabled(org.sonarlint.intellij.config.Settings.getSettingsFor(project).isBound() && bound);
```
nquinquenel, line 424: You can get rid of the long package name

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -284,6 +299,27 @@ public void actionPerformed(ActionEvent e) {
...
+      if (confirm(project)) {
+        System.out.println("Confirmed to share");
+      }
+    });
+  }
+
+  private Boolean confirm(Project project) {
+    return MessageDialogBuilder.okCancel(
+        "Share this Connected Mode configuration?",
+        "A configuration file will be created in this working directory, making it easier for other team members to configure the binding" +
```
nquinquenel, line 315: We should add the information about the connection and project key that is going to be shared

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
eray-felek-sonarsource: We have another Settings and Java does not allow import renaming using "as"

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -284,6 +305,58 @@ public void actionPerformed(ActionEvent e) {
...
+        System.out.println("Confirmed to share");
+      }
+    });
+  }
+
+  private static boolean confirm(Project project) {
+    var binding = SonarLintUtils.getService(project, ProjectBindingManager.class).getBinding();
+    var isSonarCloud = false;
+
+    if(binding == null) {
```
nquinquenel, line 322:
```suggestion
    if (binding == null) {
```

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -284,6 +305,58 @@ public void actionPerformed(ActionEvent e) {
...
+      GlobalLogOutput.get().log("Binding is not available", ClientLogOutput.Level.ERROR);
```
nquinquenel, line 323: Use `SonarLintConsole.get(project)` since the panel is tied to a project

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -284,6 +305,58 @@ public void actionPerformed(ActionEvent e) {
...
+      return false;
+    }
+
+    try {
+      isSonarCloud = SonarLintUtils.getService(project, ProjectBindingManager.class).getServerConnection().isSonarCloud();
+    } catch (InvalidBindingException e) {
+      GlobalLogOutput.get().log(e.getMessage(), ClientLogOutput.Level.ERROR);
```
nquinquenel, line 330: Same here

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -284,6 +305,58 @@ public void actionPerformed(ActionEvent e) {
...
+      return false;
+    }
+
+    var projectKey = binding.getProjectKey();
+    var url = binding.getConnectionName();
+
+    String orgString;
+
+    if(isSonarCloud) {
```
nquinquenel, line 339:
```suggestion
    if (isSonarCloud) {
```

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -284,6 +305,58 @@ public void actionPerformed(ActionEvent e) {
...
+    var binding = SonarLintUtils.getService(project, ProjectBindingManager.class).getBinding();
+    var isSonarCloud = false;
+
+    if(binding == null) {
+      GlobalLogOutput.get().log("Binding is not available", ClientLogOutput.Level.ERROR);
+      return false;
+    }
+
+    try {
+      isSonarCloud = SonarLintUtils.getService(project, ProjectBindingManager.class).getServerConnection().isSonarCloud();
```
nquinquenel, line 328: I think we could use `tryGetServerConnection` instead, with `ifPresentOrElse` instead of catching an exception

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -284,6 +305,58 @@ public void actionPerformed(ActionEvent e) {
...
+    } catch (InvalidBindingException e) {
+      GlobalLogOutput.get().log(e.getMessage(), ClientLogOutput.Level.ERROR);
+      return false;
+    }
+
+    var projectKey = binding.getProjectKey();
+    var url = binding.getConnectionName();
+
+    String orgString;
```
nquinquenel, line 337: `orgString` is a bit strange as the text can be about SonarQube, it could be something like `connectionKind` instead

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -284,6 +305,58 @@ public void actionPerformed(ActionEvent e) {
...
+
+    try {
+      isSonarCloud = SonarLintUtils.getService(project, ProjectBindingManager.class).getServerConnection().isSonarCloud();
+    } catch (InvalidBindingException e) {
+      GlobalLogOutput.get().log(e.getMessage(), ClientLogOutput.Level.ERROR);
+      return false;
+    }
+
+    var projectKey = binding.getProjectKey();
+    var url = binding.getConnectionName();
```
nquinquenel, line 335: In case of SonarCloud is not the URL but the organization, could be renamed to `connectionName`

nquinquenel (commented):
Looks good, just a few comments

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
nquinquenel: Make sure you apply the code style

nquinquenel (approved):

PR #1002, opened 2024-03-22
SLI-1338 Rule modifications are not loaded at start up

nquinquenel:


src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettingsStore.java
```diff
@@ -45,6 +48,7 @@ public SonarLintGlobalSettings getState() {
   public void loadState(SonarLintGlobalSettings settings) {
     this.settings = settings;
     initializeRulesByKey();
+    getService(BackendService.class).updateStandaloneRulesConfiguration(settings.rulesByKey);
```
damien-urruty-sonarsource, line 51: The mistake is in fact on this line: https://github.com/SonarSource/sonarlint-intellij/blob/master/src/main/java/org/sonarlint/intellij/core/BackendService.kt#L316

We pass an empty map instead of the standalone rules config

nquinquenel: Thanks! I was pretty sure I was missing something more obvious, I missed this params 😄

damien-urruty-sonarsource (approved):

PR #1010, opened 2024-04-02
SLI-1334 Propose export after binding change

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/actions/ShareConfigurationAction.kt
```diff
@@ -0,0 +1,38 @@
...
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package org.sonarlint.intellij.actions
+
+import com.intellij.openapi.actionSystem.AnActionEvent
+import javax.swing.Icon
+import org.sonarlint.intellij.util.ConfigurationSharing.Companion.exportConfiguration
+
+class ShareConfigurationAction @JvmOverloads constructor(
```
nquinquenel, line 26: You don't need to add `@JvmOverloads` and you can remove the icon parameter if it's always null

src/main/java/org/sonarlint/intellij/actions/SonarLintToolWindow.java
```diff
@@ -469,5 +473,20 @@ private Content getTaintVulnerabilitiesContent() {
   @Override
   public void bindingChanged(@Nullable ProjectBinding previousBinding, @Nullable ProjectBinding newBinding) {
     runOnUiThread(project, this::refreshViews);
+
+    showSharedConfigurationNotification(project, "Project successfully bound with \"" +
```
nquinquenel, line 477: You could leverage the triple quotes `"""`

src/main/java/org/sonarlint/intellij/actions/SonarLintToolWindow.java
```diff
@@ -469,5 +473,20 @@ private Content getTaintVulnerabilitiesContent() {
...
+        newBinding.getProjectKey() + "\" on \"" + newBinding.getConnectionName() +
+      "\"." +
+        " If you share this configuration, a file will be created in this working directory," +
+        " making it easier for other team members to configure the binding for the same project." +
+        "You may also decide to share this configuration later from your list of bound projects.", "ShareConfigurationAction",
```
nquinquenel, line 482: For the `doNotShowAgainId` you can use a const with the following format:
`private const val SKIP_SHARED_CONFIGURATION_DIALOG_PROPERTY = "SonarLint.shareConfiguration"`

src/main/java/org/sonarlint/intellij/actions/SonarLintToolWindow.java
```diff
@@ -469,5 +473,20 @@ private Content getTaintVulnerabilitiesContent() {
...
+      new ShareConfigurationAction("Share configuration", CONNECTED_MODE_BENEFITS_LINK, null));
+  }
+
+  public void showSharedConfigurationNotification(Project project, String message, String doNotShowAgainId, AnAction action) {
+    if (!PropertiesComponent.getInstance().getBoolean(doNotShowAgainId)) {
+      SonarLintProjectNotifications.Companion.get(project).showSharedConfigurationNotification("Project successfully bound. Share configuration?"
+        ,message, doNotShowAgainId, action);
```
nquinquenel, line 489:
```suggestion
      SonarLintProjectNotifications.Companion.get(project).showSharedConfigurationNotification("Project successfully bound. Share configuration?", 
      message, doNotShowAgainId, action);
```

src/main/java/org/sonarlint/intellij/actions/SonarLintToolWindow.java
```diff
@@ -469,5 +473,20 @@ private Content getTaintVulnerabilitiesContent() {
...
+    showSharedConfigurationNotification(project, "Project successfully bound with \"" +
+        newBinding.getProjectKey() + "\" on \"" + newBinding.getConnectionName() +
+      "\"." +
+        " If you share this configuration, a file will be created in this working directory," +
+        " making it easier for other team members to configure the binding for the same project." +
+        "You may also decide to share this configuration later from your list of bound projects.", "ShareConfigurationAction",
+      new ShareConfigurationAction("Share configuration", CONNECTED_MODE_BENEFITS_LINK, null));
+  }
+
+  public void showSharedConfigurationNotification(Project project, String message, String doNotShowAgainId, AnAction action) {
```
nquinquenel, line 486: I don't think you need to pass the action as parameter if it's always the same one, same for the `doNotShowAgainId`

src/main/java/org/sonarlint/intellij/actions/SonarLintToolWindow.java
nquinquenel, line 477: `bindingChanged` is applied to new binding, or to binding removal. I'd suggest to move everything to `ProjetBindingManager`, in the `bindTo` method instead

src/main/java/org/sonarlint/intellij/notifications/SonarLintProjectNotifications.kt
```diff
@@ -349,6 +349,19 @@ class SonarLintProjectNotifications(private val myProject: Project) {
         }
     }
 
+    fun showSharedConfigurationNotification(title: String,message: String, doNotShowAgainId: String, action: AnAction?) {
```
nquinquenel, line 352:
```suggestion
    fun showSharedConfigurationNotification(title: String, message: String, doNotShowAgainId: String, action: AnAction?) {
```

src/main/java/org/sonarlint/intellij/util/ConfigurationSharing.kt
```diff
@@ -0,0 +1,145 @@
...
+
+class ConfigurationSharing {
+
+    companion object {
+        @JvmStatic
+        fun exportConfiguration(project: Project?) {
+            if (project == null) return
+
+            if (confirm(project)) {
+                val configScopeId: @SystemIndependent @NonNls String = projectId(project) ?: return
```
nquinquenel, line 52: What is `@SystemIndependent` and why do you need it?

src/main/java/org/sonarlint/intellij/util/ConfigurationSharing.kt
```diff
@@ -0,0 +1,145 @@
...
+import org.sonarlint.intellij.notifications.SonarLintProjectNotifications.Companion.get
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.connection.config.GetSharedConnectedModeConfigFileParams
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.connection.config.GetSharedConnectedModeConfigFileResponse
+
+class ConfigurationSharing {
+
+    companion object {
+        @JvmStatic
+        fun exportConfiguration(project: Project?) {
+            if (project == null) return
```
nquinquenel, line 49: I'd check if project is not disposed

src/main/java/org/sonarlint/intellij/util/ConfigurationSharing.kt
```diff
@@ -0,0 +1,145 @@
...
+                createFile(configScopeId, project)
+            }
+        }
+
+        private fun createFile(configScopeId: String, project: Project) {
+            if (project.getBasePath() == null) {
+                return
+            }
+            SonarLintUtils.getService<BackendService>(BackendService::class.java)
+                .getSharedConnectedModeConfigFileContents(GetSharedConnectedModeConfigFileParams(configScopeId))
```
nquinquenel, line 62: In `getSharedConnectedModeConfigFileContents` you should pass the configscope ID as parameter, and inside the method you should  create the `GetSharedConnectedModeConfigFileParams`

src/main/java/org/sonarlint/intellij/util/ConfigurationSharing.kt
```diff
@@ -0,0 +1,145 @@
...
+            isSonarCloud = if (connection.isPresent) {
+                connection.get().isSonarCloud
+            } else {
+                SonarLintConsole.get(project).error("Connection is not present")
+                return false
+            }
+            val projectKey = binding.projectKey
+            val connectionName = binding.connectionName
+            val connectionKind: String
+            connectionKind = if (isSonarCloud) {
```
nquinquenel, line 124:
```suggestion
            val connectionKind = connectionKind = if (isSonarCloud) {
```

src/main/java/org/sonarlint/intellij/util/ConfigurationSharing.kt
```diff
@@ -0,0 +1,145 @@
...
+            var isSonarCloud = false
+            if (binding == null) {
+                SonarLintConsole.get(project).error("Binding is not available")
+                return false
+            }
+            val connection = SonarLintUtils.getService(
+                project,
+                ProjectBindingManager::class.java
+            ).tryGetServerConnection()
+            isSonarCloud = if (connection.isPresent) {
```
nquinquenel, line 115: You can inline the `isSonarCloud` declaration here

src/main/java/org/sonarlint/intellij/actions/SonarLintToolWindow.java
eray-felek-sonarsource: I prefer this approach, IMO it is more readable

src/main/java/org/sonarlint/intellij/util/ConfigurationSharing.kt
eray-felek-sonarsource: It was added by automatic Kotlin conversion. Seems like not necessary

src/main/java/org/sonarlint/intellij/util/ConfigurationSharing.kt
eray-felek-sonarsource: I think this way makes it less readable

src/main/java/org/sonarlint/intellij/util/ConfigurationSharing.kt
```diff
@@ -0,0 +1,141 @@
...
+
+class ConfigurationSharing {
+
+    companion object {
+        @JvmStatic
+        fun exportConfiguration(project: Project?) {
+            if (project == null || project.isDisposed) return
+
+            if (confirm(project)) {
+                val configScopeId: @NonNls String = projectId(project)
```
nquinquenel, line 50:
```suggestion
                val configScopeId = projectId(project)
```

src/main/java/org/sonarlint/intellij/util/ConfigurationSharing.kt
```diff
@@ -0,0 +1,141 @@
...
+        private fun createFile(configScopeId: String, project: Project) {
+            if (project.getBasePath() == null) {
+                return
+            }
+            SonarLintUtils.getService<BackendService>(BackendService::class.java)
+                .getSharedConnectedModeConfigFileContents(configScopeId)
+                .thenAcceptAsync { sharedFileContent: GetSharedConnectedModeConfigFileResponse ->
+                    val root: Path = Paths.get(project.getBasePath())
+                    val filename: String
+                    filename = if (SonarLintUtils.isRider()) {
```
nquinquenel, line 64: I still think it's a better habit to inline everything, I don't see an added value in having `filename` declared just before. Also it's what we already do in the codebase, it's also important to stay consistent.

`val filename = if (SonarLintUtils.isRider()) {`

src/main/java/org/sonarlint/intellij/actions/SonarLintToolWindow.java
nquinquenel: I think it's better with a text block; you don't have to escape the quotes inside the text block and it removes all the plus signs on each line. Overall, I don't see a reason not to use it. Also, we started using this text block in other parts of the code, and as I mentioned in another comment, we should stay consistent in the codebase.

nquinquenel: You may have missed this one @eray-felek-sonarsource

nquinquenel (approved):
It's looking good, just a few last comments to tackle!

PR #1011, opened 2024-04-04
SLI-1335 Suggest auto binding for shared configuration

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -137,6 +142,57 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
         suggestionsByConfigScopeId.forEach { (configScopeId, suggestions) -> suggestAutoBind(findProject(configScopeId), suggestions) }
     }
 
+    override fun suggestConnection(suggestionsByConfigScope: MutableMap<String, MutableList<ConnectionSuggestionDto>>?) {
```
nquinquenel, line 145:
```suggestion
    override fun suggestConnection(suggestionsByConfigScope: Map<String, List<ConnectionSuggestionDto>>) {
```

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -137,6 +142,57 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
...
+        if (suggestionsByConfigScope != null && suggestionsByConfigScope.size == 1) {
```
nquinquenel, line 146: `suggestionsByConfigScope != null` can be removed as the parameter should not be nullable

nquinquenel, line 146: You should actually iterate over all the map entries, not only the first one. The baby step we discussed about, it's for the list of `ConnectionSuggestionDto`, we can assume it should be only one element in the suggestion list, but we could have multiple config scope (if you have multiple projects opened for example)

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -137,6 +142,57 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
...
+                project, String.format(
+                    """
+                    A Connected Mode configuration file is available to bind to project "%s" on %s "%s".
+                """.trimIndent(), projectKey, connectionKind, connectionName
+                ), SKIP_AUTO_SHARE_CONFIGURATION_DIALOG_PROPERTY
+            )
+        }
+    }
+
+    private fun getAutoShareConfigParams(dto: MutableMap.MutableEntry<String, MutableList<ConnectionSuggestionDto>>): Triple<String, String, String> {
```
nquinquenel, line 164: This method should probably transform a `Either<SQDto, SCDto>` to the information you need in the Triple

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -137,6 +142,57 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
...
+        val isSQ = dto.value.stream().findFirst().get().connectionSuggestion.isLeft
+
+        var sq = SonarQubeConnectionSuggestionDto(null, null)
```
nquinquenel, line 167: As a general practice, I'd not recommend using DTO object for internal usage on the client side, it's should be when transferring data between back and front

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -137,6 +142,57 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
...
+        var sc = SonarCloudConnectionSuggestionDto(null, null)
+
+        if (isSQ) {
```
nquinquenel, line 170: You can probably do a single `if (isLeft) {} else {}` than multiples if

src/main/java/org/sonarlint/intellij/notifications/SonarLintProjectNotifications.kt
```diff
@@ -362,6 +362,20 @@ class SonarLintProjectNotifications(private val myProject: Project) {
         }
     }
 
+    fun showAutoSharedConfigurationNotification(title: String, message: String, doNotShowAgainId: String, action: AnAction?) {
+        SONARLINT_GROUP.createNotification(
+            title,
+            message,
+            NotificationType.WARNING
```
nquinquenel, line 369: I think we could use INFO type?

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -137,6 +142,57 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
         suggestionsByConfigScopeId.forEach { (configScopeId, suggestions) -> suggestAutoBind(findProject(configScopeId), suggestions) }
     }
 
+    override fun suggestConnection(suggestionsByConfigScope: MutableMap<String, MutableList<ConnectionSuggestionDto>>?) {
+        if (suggestionsByConfigScope != null && suggestionsByConfigScope.size == 1) {
+            val dto = suggestionsByConfigScope.entries.stream().findFirst().get()
```
nquinquenel, line 147: Following the previous comment, this should be removed, but just for information, you can simplify this by `suggestionsByConfigScope.entries.first` to get the first entry, or even `suggestionsByConfigScope.first`

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -137,6 +140,47 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
         suggestionsByConfigScopeId.forEach { (configScopeId, suggestions) -> suggestAutoBind(findProject(configScopeId), suggestions) }
     }
 
+    override fun suggestConnection(suggestionsByConfigScope: Map<String, List<ConnectionSuggestionDto>>) {
+        for (suggestion in suggestionsByConfigScope) {
+            val project = BackendService.findModule(suggestion.key)?.project
+                ?: BackendService.findProject(suggestion.key) ?: return
```
nquinquenel, line 146: `continue` instead of `return`

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -137,6 +140,47 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
...
+
+            if (suggestion.value.size == 1) {
+                //It was decided to only handle the case where there is only one notification per configuration scope
```
nquinquenel, line 149: Only first a first baby step, probably we should revisit this at the end of the sprint or during a hardening

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -137,6 +140,47 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
...
+
+    private fun getAutoShareConfigParams(uniqueSuggestion: ConnectionSuggestionDto): Triple<String, String, String> {
+        val connectionKind: String
+        val projectKey: String
+        val connectionName: String
+
+        if (uniqueSuggestion.connectionSuggestion.isRight) {
+            connectionKind = "SonarCloud organization"
+            projectKey = uniqueSuggestion.connectionSuggestion.right.projectKey
+            connectionName = uniqueSuggestion.connectionSuggestion.right.organization
```
nquinquenel, line 174: You can directly return a `Triple()` inside the if and else, so you don't have to declare the variables

nquinquenel (approved):
LGTM, some comments still

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
eray-felek-sonarsource: Yep, almost forgot, thanks

PR #1012, opened 2024-04-05
MMF-3580 Share connected mode setup among contributors

nquinquenel:


eray-felek-sonarsource (approved):

nandorholozsnyak:
Hello there,
I'm curious if it will support module overrides per binding? I have a lot of maven modules with different rules, and I would like to share it with other devs too. It might be a bit strange, for example if they do not have the same workspace setup but for one big project it is an acceptable thing in my opinion.

nquinquenel:
Hello @nandorholozsnyak, thank you for your valuable feedback. We currently do not support module overriding for this feature. I suggest you to create a thread on our [Community Forum](https://community.sonarsource.com/c/sl/fault/6), so it can have visibility from our Product Managers.

nandorholozsnyak:
> Hello @nandorholozsnyak, thank you for your valuable feedback. We currently do not support module overriding for this feature. I suggest you to create a thread on our [Community Forum](https://community.sonarsource.com/c/sl/fault/6), so it can have visibility from our Product Managers.

Thank you!

PR #1015, opened 2024-04-11
SLI-1358 Bind auto shared config

eray-felek-sonarsource:


nquinquenel (approved):
Few comments

src/main/java/org/sonarlint/intellij/actions/filters/AutoShareTokenExchangeAction.kt
```diff
@@ -20,16 +20,39 @@
...
 
 class AutoShareTokenExchangeAction(
     text: String,
+    connectionSuggestionDto: ConnectionSuggestionDto,
+    project: Project,
 ) : AbstractSonarAction(
     text, null, null
 ) {
 
+    private val connectionSuggestionDto = connectionSuggestionDto
```
nquinquenel, line 36: To be put as constructor

src/main/java/org/sonarlint/intellij/actions/filters/AutoShareTokenExchangeAction.kt
```diff
@@ -20,16 +20,39 @@
...
+
+    private val project = project
```
nquinquenel, line 38: same

PR #1018, opened 2024-04-15
Fix assist binding

eray-felek-sonarsource:


nquinquenel:
@eray-felek-sonarsource what is this about?

PR #1019, opened 2024-04-16
SLI-1368 Implement telemetry

eray-felek-sonarsource:


jblievremont (approved):
I haven't seen updates to tests, is there a way to check that the correct telemetry methods are called?

Otherwise, LGTM 🚢 !

PR #1021, opened 2024-04-22
Handle SonarCloud import, apply IntelliJ UI guidelines, fix code style

nquinquenel:


its/src/test/kotlin/org/sonarlint/intellij/its/tests/domain/OpenInIdeTests.kt
```diff
@@ -38,8 +38,8 @@ class OpenInIdeTests {
         fun createConnection(token: String) {
             with(remoteRobot) {
                 idea {
-                    dialog("Do you trust this SonarQube server?") {
-                        button("Connect to this SonarQube server").click()
+                    dialog("Trust This SonarQube Server?") {
```
eray-felek-sonarsource, line 41: Was this change discussed, Trust this sound more like a command than a question :)

nquinquenel: IntelliJ guidelines [mention](https://jetbrains.github.io/ui/text/writing_short/#do-not-address-the-user) not to address the user, so we should avoid using "you" in sentences

PR #1022, opened 2024-04-22
SLI-1369 Improve visual feedback

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/util/ConfigurationSharing.kt
```diff
@@ -93,6 +93,12 @@ class ConfigurationSharing {
                         )
                         SonarLintConsole.get(project).error("Error while creating the shared file, IO exception : " + e.message)
                     }
+
+                    get(project).simpleNotification(
+                        null,
+                        "File \"$filename\" has been created. It might take some time for changes to be reflected in the project view.",
```
nquinquenel, line 99:
```suggestion
                        "File '$filename' has been created. It might take some time for changes to be reflected in the project view.",
```
Based on IntelliJ guidelines, we should not put double quotes in sentences

src/main/java/org/sonarlint/intellij/util/ConfigurationSharing.kt
```diff
@@ -93,6 +93,12 @@ class ConfigurationSharing {
                         )
                         SonarLintConsole.get(project).error("Error while creating the shared file, IO exception : " + e.message)
                     }
+
+                    get(project).simpleNotification(
```
nquinquenel, line 97: Move the notification after the the `BufferedWriter` wrote the file, otherwise it will get displayed even if it failed

nquinquenel (approved):
2 changes to make, otherwise LGTM

nquinquenel (approved):
2 changes to make, otherwise LGTM

src/main/java/org/sonarlint/intellij/util/ConfigurationSharing.kt
eray-felek-sonarsource: Nice catch

PR #1027, opened 2024-04-24
Hardening for 10.5, round 2

nquinquenel:


eray-felek-sonarsource (commented):
QAs are failing

src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -391,8 +392,18 @@ class BackendService : Disposable {
     }
 
     private fun getTimeoutProperty(propertyName: String): Duration? {
-        val property = System.getProperty(propertyName)
-        return property?.let { Duration.parse(it) }
+        return System.getProperty(propertyName)?.let {
```
eray-felek-sonarsource, line 395: Why are we doing this?

nquinquenel: When we worked on the HTTP part, we decided to allow both format (`PnDTnHnMn.nS` and minutes) for the timeout property.

Reason is, we think for most users it will be minutes only, but for tests we need to allow smaller durations as well. Because `PnDTnHnMn.nS` might not be user friendly, we allowed both.

Since we made the move to SLOOP, the new code did not follow the same behavior, as you can see we only take into account the format `PnDTnHnMn.nS` and not the minutes anymore (which is also in contradiction with the documentation).

eray-felek-sonarsource (approved):

PR #1048, opened 2024-05-07
Implement promotion method

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -726,7 +727,12 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
     }
 
     override fun promoteExtraEnabledLanguagesInConnectedMode(configurationScopeId: String, languagesToPromote: Set<Language>) {
-        // TODO: Move 'raw' analysis to SLCORE
+        if (languagesToPromote.isEmpty()) return
+
+        val project = BackendService.findModule(configurationScopeId)?.project
+            ?: BackendService.findProject(configurationScopeId) ?: return
+
+        project.getMessageBus().syncPublisher<PromotionListener>(PromotionListener.TOPIC).extraLanguagePromotionRaised(languagesToPromote)
```
nquinquenel, line 735: I think you can just directly call the methods to show the notification here; you don't need to create a listener?

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -188,7 +200,21 @@ class PromotionProvider(private val project: Project) {
...
+            languages.find {
+                it.name == language.name
+            }?.let { extraLanguagesToPromote.add(org.sonarsource.sonarlint.core.client.utils.Language.fromDto(it).label) }
+        }
+
+        return extraLanguagesToPromote
+    }
+
     private fun showPromotion(notifications: SonarLintProjectNotifications, content: String) {
+        PropertiesComponent.getInstance().setValue(LAST_PROMOTION_NOTIFICATION_DATE, Instant.now().toEpochMilli().toString())
```
nquinquenel, line 216: Is this change needed? It's all done after in the `if()`, I think there's an issue

nquinquenel (commented):
Looks good but I think it can be simplified

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
eray-felek-sonarsource: It was leftover code while testing forgot to delete

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
eray-felek-sonarsource: PromotionProvider is a project-level service. I had initially implemented to use as methods and converted to this design to have separation of concerns. Calling the method would actually require to change the design and I think it would not be a good design choice.

nquinquenel: `PromotionProvider` is currently based on listener because it fits what it needs to do: listen to a file opened and when an analysis is started.

Here we are already listening to a kind of event: we receive a request to `promoteExtraEnabledLanguagesInConnectedMode` when needed from the backend, it does not make much sense to add a listener on top of this.

It's also not an issue if it's a project-level service, we have access to the project so you can directly call getService(project, PromotionProvider::class.java)

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -188,6 +200,18 @@ class PromotionProvider(private val project: Project) {
...
+    private fun findExtraLanguages(languagesToPromote: Set<Language>, languages: Set<Language>): List<String> {
+        val extraLanguagesToPromote = ArrayList<String>()
+
+        languagesToPromote.forEach { language ->
+            languages.find {
+                it.name == language.name
+            }?.let { extraLanguagesToPromote.add(org.sonarsource.sonarlint.core.client.utils.Language.fromDto(it).label) }
+        }
+
+        return extraLanguagesToPromote
```
nquinquenel, line 212:
```suggestion
        return languagesToPromote.mapNotNull { language ->
            languages.find {
                it.name == language.name
            }?.let { org.sonarsource.sonarlint.core.client.utils.Language.fromDto(it).label }
        }.toList()
```

src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt
```diff
@@ -172,12 +182,14 @@ class PromotionProvider(private val project: Project) {
...
-        val language = findLanguage(extension, extraEnabledLanguagesInConnectedMode)
+    private fun processExtraLanguagePromotion(notifications: SonarLintProjectNotifications, languagesToPromote: Set<Language>) {
+        val languages = findExtraLanguages(languagesToPromote, extraEnabledLanguagesInConnectedMode)
 
-        if (language != null) {
-            showPromotion(notifications, "Enable " + language.label + " analysis by connecting your project")
-        }
+        if (languages.isEmpty()) return
+
+        val languagesText = languages.joinToString(", ")
```
nquinquenel, line 190: Maybe to stay consistent with SLE, we can do the following:
```suggestion
        val languagesText = languages.joinToString(" / ")
```

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
eray-felek-sonarsource: Makes sense, changed to use method

nquinquenel (approved):
LGTM!

PR #1061, opened 2024-05-22
Hardening 10.6

nquinquenel:


src/main/java/org/sonarlint/intellij/util/SonarLintAppUtils.java
```diff
@@ -180,11 +181,16 @@ public boolean visitFile(VirtualFile file) {
...
-      || ProjectFileIndex.getInstance(project).isInLibrary(file)
-      || (!ApplicationManager.getApplication().isUnitTestMode() && FileUtilRt.isTooLarge(file.getLength())));
-    return Boolean.FALSE == toSkip;
+    try {
+      var toSkip = computeReadActionSafely(project, () -> isGeneratedSourceByAnyFilter(file, project)
+        || ProjectFileIndex.getInstance(project).isExcluded(file)
+        || ProjectFileIndex.getInstance(project).isInLibrary(file)
+        || (!ApplicationManager.getApplication().isUnitTestMode() && FileUtilRt.isTooLarge(file.getLength())));
+      return Boolean.FALSE == toSkip;
+    } catch (Exception e) {
```
eray-felek-sonarsource, line 190: What are the exception/exceptions we are facing? We should try to avoid catching generic exceptions

eray-felek-sonarsource (approved):
Small comment else LGTM

src/main/java/org/sonarlint/intellij/util/SonarLintAppUtils.java
nquinquenel: I think any exception. The reasons might be various because we rely on IntelliJ APIs, it could be IOException, some index issues on VFS files, corrupted stuff...

The idea is to skip the file/directory if an error occurs and continue with the rest. At least it won't prevent the analysis from running.

PR #1065, opened 2024-05-24
SLI-1401 Avoid technical message while fetching projects

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -160,7 +160,7 @@ private Map<String, SonarProjectDto> downloadProjectList(ServerConnection select
     try {
       return ProgressManager.getInstance().run(downloadTask);
     } catch (Exception e) {
-      var msg = e.getMessage() != null ? e.getMessage() : "Failed to download list of projects";
+      var msg = "Failed to download list of projects. Make sure an authorized connection is present.";
```
nquinquenel, line 163: This is not always the root cause, it could be anything (no connection, wrong token, wrong permission, etc etc).

Maybe we could log the full stack trace in the console and propose the user to check it

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -160,7 +162,8 @@ private Map<String, SonarProjectDto> downloadProjectList(ServerConnection select
     try {
       return ProgressManager.getInstance().run(downloadTask);
     } catch (Exception e) {
-      var msg = e.getMessage() != null ? e.getMessage() : "Failed to download list of projects";
+      var msg = "Failed to download list of projects. Please check the logs for more details.";
+      getService(project, SonarLintConsole.class).debug(e.getMessage());
```
nquinquenel, line 166: I would log this as error, otherwise most users won't see the logs (if it's debug, you have to enable specific options to have them)

nquinquenel (approved):

PR #1066, opened 2024-05-24
SLI-1428 Plugin verifier task should not hide the master pipeline state

nquinquenel:
The nightly cronjob won't run on master anymore but on a dedicated branch "plugin-verifier".
Notifications will be received via slack.


thahnen (approved):
LGTM!

PR #1068, opened 2024-05-24
SLI-1357 List of connections should not be hidden

nquinquenel:
Works even in zoom 200%


eray-felek-sonarsource (commented):
Does not work on my machine

nquinquenel:
> Does not work on my machine

For Linux it is better, I cannot hide the connection selection anymore, even if 200% zoom. I think we can merge since it still an improvement, but I'll keep the ticket opened.

georgii-borovinskikh-sonarsource (approved):

PR #1074, opened 2024-05-30
SLI-1436 Analysis in CLion on Windows does not work

nquinquenel:


src/main/java/org/sonarlint/intellij/util/VirtualFileUtils.kt
```diff
@@ -28,7 +28,7 @@ import java.net.URISyntaxException
 object VirtualFileUtils {
     fun toURI(file: VirtualFile): URI? {
         return try {
-            URI(file.url.replace(" ", "%20"))
+            URI("${file.fileSystem.protocol}:///${file.path}".replace(" ", "%20"))
```
damien-urruty-sonarsource, line 31: My concern is that this won't work with remote files. But maybe we already don't support this case with SLI?
Also I think you could at least print a log in the catch below

damien-urruty-sonarsource (approved):
I think this works for most cases, except remote FS. Not sure we need to support this case

src/main/java/org/sonarlint/intellij/util/VirtualFileUtils.kt
damien-urruty-sonarsource: You could maybe check if the `file.isInLocalFileSystem()`? If it's not you will try to build a local URI that maybe does not exist

nquinquenel: We were effectively only computing URI for file in local file system before SLOOP

PR #1077, opened 2024-06-03
Enable filtering issues by severity in Current File and Report tab

strangelookingnerd:
grafik

This PR aims to add an option to the `Current File` and `Report` tab that allows to filter the list of issues by their severity. The advantage for the user is to set a threshold that is relevant when analyzing issues found by SonarLint in order to e.g. focus on the more severe issues rather than `info` severity.

This is a feature that was already requested in the past (see https://github.com/sonar-intellij-plugin/sonar-intellij-plugin/issues/95 or https://community.sonarsource.com/t/sonarlint-plugin-intellij-ide-group-by-critical-or-blocker/20062) which is why I think it is still relevant and should be considered for future versions of this plugin.

The implementation is heavily based upon `FilterSecurityHotspotAction` and related classes and provides the same filter for `Current File` and `Report`. It uses a similar approach to display only issues based on the threshold severity filter set by the user.

Since this is my first go at developing an IntelliJ Plugin there might be some oversights on my end that I'd be happy to fix 👍🏼 

Please review our [contribution guidelines](https://github.com/SonarSource/sonarlint-intellij/blob/master/contributing.md).

And please ensure your pull request adheres to the following guidelines: 

- [x] Please explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- [x] Use the following formatting style: [SonarSource/sonar-developer-toolset](https://github.com/SonarSource/sonar-developer-toolset#code-style)
- [x] Provide a unit test for any code you changed
- [ ] If there is a [JIRA](http://jira.sonarsource.com/browse/SLI) ticket available, please make your commits and pull request start with the ticket ID (SLI-XXXX)

### DIFF
--- a/src/main/java/org/sonarlint/intellij/actions/FilterIssueSeverityActionGroup.java
+++ b/src/main/java/org/sonarlint/intellij/actions/FilterIssueSeverityActionGroup.java
@@ -1,0 +1,67 @@
+/*
+ * SonarLint for IntelliJ IDEA
+ * Copyright (C) 2015-2024 SonarSource
+ * sonarlint@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package org.sonarlint.intellij.actions;
+
+import com.intellij.openapi.actionSystem.ActionGroup;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import javax.swing.Icon;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.sonarlint.intellij.actions.filters.FilterIssueSeverityAction;
+import org.sonarlint.intellij.actions.filters.FilterIssueSeveritySettings;
+import org.sonarlint.intellij.actions.filters.IssueSeverityFilters;
+
+public class FilterIssueSeverityActionGroup extends ActionGroup {
+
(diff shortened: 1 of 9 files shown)


AlexRage:
Hi @strangelookingnerd 

Thanks for your proposal and effort on this topic. 
Unfortunately we don't plan to add a filtering system in the IDE because it would go against the Clean As You Code methodology and more specifically the promotion of solving every issue in your code. 
We really appreciate your effort here and we will let you know if some change of priorities happens in the future.

strangelookingnerd:
> Unfortunately we don't plan to add a filtering system in the IDE because it would go against the Clean As You Code methodology and more specifically the promotion of solving every issue in your code.

I strongly disagree that filtering would in any way interfere with that methodology. Giving users a way to focus on the more important / severe issues first would - in my opinion - promote a structured way of solving issues in the first place.

I find it worrisome that so many good suggestions from the community are blatantly ignored because they apparently do not follow a rather vague interpretation of _Clean As You Code_. 

Still, there is hope that at some point this will change and we can revisit this PR.

AlexRage:
I totally understand your point of view about bringing some structure when solving issues. A potential side effect I can see from filtering issues by severity might be to only solve important / severe issues and not the rest, don't you think? 
I would be super happy to schedule an interview with you to have your thoughts on that topic. Would you be able to participate?

strangelookingnerd:
> I totally understand your point of view about bringing some structure when solving issues. A potential side effect I can see from filtering issues by severity might be to only solve important / severe issues and not the rest, don't you think?

From experience I'd say a developer that wants to solve _any_ issues would either start with the low-hanging fruits or the more severe ones and work their way through a file or project. Usually I'd say one would go even further and solve issue per rule, because it oftentimes makes life easier. You would rarely go line-by-line in these scenarios because it is considerably slower as you need to _jump_ from one rule to another and switch back and forth between different complexities. This is especially true when applying clean code principles to an entire project but also stands when looking at new / changed code alone.

> Clean as You Code from Sonar is a methodology that progressively improves the quality of the entire codebase with minimal cost and effort.
> -- https://www.sonarsource.com/resources/solution-briefs/how-to-implement-clean-as-you-code/

No matter which of the above paths you would take, a filter would support just that. It does not ultimately mean that a developer would stop fixing issue once everything from the filtered view is gone. In my opinion it would maybe also bring a feeling of accomplishment especially when dealing with an overhelming number of issues that keeps them going.

> I would be super happy to schedule an interview with you to have your thoughts on that topic. Would you be able to participate?

I'd love to! Let me know what works for you.

AlexRage:
Thanks for your answer. 
Please feel free to schedule some time [here](https://calendar.app.google/UrX5MLfVWTGYTCP5A).
Can't wait to speak to you about this!

nquinquenel:
Thank you for your contribution, @strangelookingnerd. I'm closing this PR until we can agree on this topic.

PR #1093, opened 2024-07-03
Avoid blocking thread while waiting for excluded files

nquinquenel:


nquinquenel:
We should remove this part at all

PR #1100, opened 2024-07-08
SLI-1492 Test exporting and importing shared configuration

eray-felek-sonarsource:


its/src/test/kotlin/org/sonarlint/intellij/its/tests/domain/OpenInIdeTests.kt
```diff
@@ -36,6 +36,26 @@ import org.sonarlint.intellij.its.tests.domain.SecurityHotspotTabTests.Companion
 class OpenInIdeTests {
 
     companion object {
+        fun shareConfiguration() {
```
nquinquenel, line 39: This class is for `Open in IDE`, I don't think it belongs here

its/projects/connectedmode-json/pom.xml
```diff
@@ -0,0 +1,29 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.sonarsource.sonarlint.core</groupId>
+  <artifactId>connectedmode-json</artifactId>
```
nquinquenel, line 5: I think `connectedmode-json` is not a super great name; what about `shared-connected-mode`, for instance? Or something else

its/projects/connectedmode-json/.sonarlint/connectedMode.json
```diff
@@ -0,0 +1,4 @@
+{
+    "sonarQubeUri": "http://127.0.0.1:54428",
```
nquinquenel, line 2: Is it always the same port in the end?

nquinquenel (commented):
Minor comments and one question

its/src/test/kotlin/org/sonarlint/intellij/its/tests/ConnectedIdeaTests.kt
```diff
@@ -283,6 +287,37 @@ class ConnectedIdeaTests : BaseUiTest() {
...
+        @Test
+        fun should_receive_shared_config() = uiTest {
+            openExistingProject("connectedmode-json")
+            enableConnectedModeFromCurrentFilePanel(CONNECTEDMODE_JSON_PROJECT_KEY, true, "Orchestrator")
+            notification("Share configuration").click()
+            shareConfiguration()
+            clearConnections()
+            goBackToWelcomeScreen()
+            openExistingProject("connectedmode-json")
+            notification("Use configuration")
```
nquinquenel, line 317: It could worth testing if the connection was successfully imported

eray-felek-sonarsource: Whenever we open a project, it uses the configuration information we initially stored in the main project(under its/shared-connected-mode/.sonarlint. So even after we use export it would be using the URL with the old port which makes it not possible to test this feature since we are not connected to it and we do not have the token for it.

its/src/test/kotlin/org/sonarlint/intellij/its/tests/ConnectedIdeaTests.kt
```diff
@@ -283,6 +287,35 @@ class ConnectedIdeaTests : BaseUiTest() {
...
+            )
+            // Build and analyze project to raise issue
+            executeBuildWithMaven("projects/shared-connected-mode/pom.xml", ORCHESTRATOR)
+            firstIssueKey = getFirstIssueKey(adminWsClient)
+        }
+
+        @Test
+        fun should_receive_shared_config() = uiTest {
+            openExistingProject("shared-connected-mode")
+            notification("Use configuration")
```
nquinquenel, line 311: This line does nothing

nquinquenel (approved):
I did a refactoring of the tests, it is now exporting and then importing the connection successfully (hopefully). Can you give it a review @eray-felek-sonarsource ?

its/src/test/kotlin/org/sonarlint/intellij/its/tests/ConnectedIdeaTests.kt
nquinquenel: You can test if you export and import in the same test - so you don't have to restart orchestrator (which would modify the port)

eray-felek-sonarsource:
> I did a refactoring of the tests, it is now exporting and then importing the connection successfully (hopefully). Can you give it a review @eray-felek-sonarsource ?

LGTM

PR #1103, opened 2024-07-09
SLI-1485 Open in IDE test should consistently find and open the issue

nquinquenel:


eray-felek-sonarsource (commented):
There seems to be some failing tests on validate task

eray-felek-sonarsource (approved):

PR #1114, opened 2024-07-24
SLI-1512 Handle the opening of a fix suggestion from SonarCloud

nquinquenel:


its/src/test/kotlin/org/sonarlint/intellij/its/fixtures/WelcomeFrame.kt
```diff
@@ -31,7 +31,7 @@ import com.intellij.remoterobot.stepsProcessing.step
 import java.time.Duration
 
 fun RemoteRobot.welcomeFrame(function: WelcomeFrame.() -> Unit) {
-    find(WelcomeFrame::class.java, Duration.ofSeconds(5)).apply(function)
+    find(WelcomeFrame::class.java, byXpath("//div[@class='FlatWelcomeFrame']"), Duration.ofSeconds(7)).apply(function)
```
thahnen, line 34: A more general question, why not use the `awitility` [library](https://github.com/awaitility/awaitility) for something like this?
I use it for SLE as Julien introduced it and it is superb!

nquinquenel: We use it for unit tests, but for ITs we directly leverage timeout from [IntelliJ UI test robot](https://github.com/JetBrains/intellij-ui-test-robot), it's part of the fixtures provided by the framework

thahnen (approved):
LGTM, but as usual, some more comments on the SLI codebase would be great, even though a lot of it is luckily well-developed and self-explanatory!

PR #1118, opened 2024-07-30
SLI-1514 Do not block UI thread while doing pre commit check

nquinquenel:


thahnen (approved):
Looks good to me, even though I don't really understand why this isn't a more generic job that is running and if it doesn't succeeds, schedules itself with a specific time instead of calling `Thread.sleep(...)`. I can only guess that something like this isn't available in IntelliJ :open_mouth:

nquinquenel:
> Looks good to me, even though I don't really understand why this isn't a more generic job that is running and if it doesn't succeeds, schedules itself with a specific time instead of calling `Thread.sleep(...)`. I can only guess that something like this isn't available in IntelliJ 😮

Right, I modified the logic 👍🏻

PR #1130, opened 2024-08-08
10.8.1 bug-fix release

nquinquenel:


src/main/java/org/sonarlint/intellij/finding/LiveFindings.kt
```diff
@@ -35,6 +35,21 @@ class LiveFindings(
             securityHotspotsPerFile.filterKeys { files.contains(it) })
     }
 
+    fun merge(other: LiveFindings?): LiveFindings {
```
eray-felek-sonarsource, line 38: We could add more clarification to this function. i.e comments, better naming

eray-felek-sonarsource (approved):
LGTM

PR #1135, opened 2024-08-21
SLI-1551 Fix AI Fix Suggestion Navigation

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/fix/ShowFixSuggestion.kt
```diff
@@ -37,9 +38,10 @@ import org.sonarsource.sonarlint.core.rpc.protocol.client.fix.FixSuggestionDto
 class ShowFixSuggestion(private val project: Project, private val file: VirtualFile, private val fixSuggestion: FixSuggestionDto) {
 
     fun show() {
+        var hasNavegated = false
```
thahnen, line 41: `hasNavigated` but I don't care for a typo, only if you fix the other typo as well.

src/main/java/org/sonarlint/intellij/fix/ShowFixSuggestion.kt
```diff
@@ -55,10 +57,23 @@ class ShowFixSuggestion(private val project: Project, private val file: VirtualF
             var successfullyOpened = true
 
             fixSuggestion.fileEdit().changes().forEachIndexed { index, change ->
-                fileEditorManager.selectedTextEditor?.let {
-                    val startLine = change.beforeLineRange().startLine
-                    val endLine = change.beforeLineRange().endLine
+                val startLine = change.beforeLineRange().startLine
+                val endLine = change.beforeLineRange().endLine
+
+                if(!hasNavegated) {
```
thahnen, line 63: Add a space between the `if` and its brackets. Only if you fix the other typo as well.

thahnen (approved):
LGTM, the fixes proposed are optional.

src/main/java/org/sonarlint/intellij/fix/ShowFixSuggestion.kt
eray-felek-sonarsource: Since there are 2 more bugs to be fixed, I will change them while fixing those

PR #1141, opened 2024-08-26
SLI-1556 In Rider, all solution's files should be analyzed

nquinquenel:
[SLI-1556](https://sonarsource.atlassian.net/browse/SLI-1556)

[SLI-1556]: https://sonarsource.atlassian.net/browse/SLI-1556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -712,31 +707,14 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
...
-            if (module.isDisposed) {
-                return@forEach
+        if (isRider()) {
+            computeRiderSharedConfiguration(module.project, configScopeId)?.let {
+                clientFiles.add(it)
             }
-            filesInContentRoots.addAll(visitAndAddFiles(contentRoot, module))
         }
 
-        val forcedLanguages = collectContributedLanguages(module, filesInContentRoots)
```
eray-felek-sonarsource, line 726: Why did we need to change forcedLanguages logic?

nquinquenel: It was not changed, it's still in `listModuleFiles`

Before, for retrieving list of files, we had a specific case for Rider, and for the rest. Now, we have a single case for everything, and on top of that we add extension point defined in plugin specific directory. Forced languages were duplicated for those 2 previous cases, now it's at the same place

eray-felek-sonarsource (approved):

PR #1151, opened 2024-08-30
SLI-1536 Warn users about potential proxy issues when an issue occurs

eray-felek-sonarsource:
[SLI-1536](https://sonarsource.atlassian.net/browse/SLI-1536)

[SLI-1536]: https://sonarsource.atlassian.net/browse/SLI-1536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarlint/intellij/config/global/wizard/ServerStep.java
+++ b/src/main/java/org/sonarlint/intellij/config/global/wizard/ServerStep.java
@@ -21,10 +21,12 @@ package org.sonarlint.intellij.config.global.wizard;
 
 import com.intellij.ide.wizard.AbstractWizardStepEx;
 import com.intellij.ide.wizard.CommitStepException;
+import com.intellij.openapi.application.ApplicationInfo;
 import com.intellij.ui.DocumentAdapter;
 import com.intellij.ui.HyperlinkAdapter;
 import com.intellij.ui.components.JBTextField;
 import com.intellij.util.net.HttpConfigurable;
+import com.intellij.util.text.VersionComparatorUtil;
 import com.intellij.util.ui.SwingHelper;
 import java.awt.event.MouseEvent;
 import java.net.MalformedURLException;
@@ -37,6 +39,7 @@ import javax.swing.JEditorPane;
 import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.JRadioButton;
+import javax.swing.JTextArea;
 import javax.swing.JTextField;
 import javax.swing.event.DocumentEvent;
 import javax.swing.event.DocumentListener;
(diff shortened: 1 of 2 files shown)


src/main/java/org/sonarlint/intellij/config/global/wizard/ServerStep.java
```diff
@@ -94,6 +107,17 @@ protected void textChanged(DocumentEvent e) {
 
     nameField.setToolTipText("Name of this configuration (mandatory field)");
 
+    var version = getIntelliJVersion();
+
+    if(isVersionHigher(version, "2024.2")) {
```
nquinquenel, line 112:
```suggestion
    if (isVersionHigher(version, "2024.2")) {
```

src/main/java/org/sonarlint/intellij/config/global/wizard/ServerStep.java
```diff
@@ -74,8 +77,18 @@ public class ServerStep extends AbstractWizardStepEx {
...
   private JEditorPane compareProducts;
+  private JTextArea proxyWarningText;
   private ErrorPainter errorPainter;
 
+  public static String getIntelliJVersion() {
+    ApplicationInfo appInfo = ApplicationInfo.getInstance();
+    return appInfo.getFullVersion();
+  }
+
+  public static boolean isVersionHigher(String currentVersion, String specificVersion) {
```
nquinquenel, line 88: I guess it should be `isVersionHigherOrEquals`?

src/main/java/org/sonarlint/intellij/config/global/wizard/ServerStep.java
```diff
@@ -74,8 +77,18 @@ public class ServerStep extends AbstractWizardStepEx {
   private JEditorPane sonarCloudFree;
   private JEditorPane sonarQubeFree;
   private JEditorPane compareProducts;
+  private JTextArea proxyWarningText;
   private ErrorPainter errorPainter;
 
+  public static String getIntelliJVersion() {
+    ApplicationInfo appInfo = ApplicationInfo.getInstance();
```
nquinquenel, line 84:
```suggestion
    var appInfo = ApplicationInfo.getInstance();
```

src/main/java/org/sonarlint/intellij/config/global/wizard/ServerStep.java
```diff
@@ -94,6 +107,17 @@ protected void textChanged(DocumentEvent e) {
 
     nameField.setToolTipText("Name of this configuration (mandatory field)");
 
+    var version = getIntelliJVersion();
+
+    if(isVersionHigher(version, "2024.2")) {
+      proxyWarningText.setEnabled(true);
```
nquinquenel, line 113: Here you put `proxyWarningText.setEnabled(true);` and below `proxyWarningText.setVisible(false);`

Shouldn't those two statements somewhat match? eg, both should be setEnable or setVisible?

src/main/java/org/sonarlint/intellij/config/global/wizard/ServerStep.java
```diff
@@ -74,8 +77,18 @@ public class ServerStep extends AbstractWizardStepEx {
   private JEditorPane sonarCloudFree;
   private JEditorPane sonarQubeFree;
   private JEditorPane compareProducts;
+  private JTextArea proxyWarningText;
   private ErrorPainter errorPainter;
 
+  public static String getIntelliJVersion() {
```
nquinquenel, line 83: It may be worth putting those two methods in a utils class

nquinquenel (commented):
We also have other places where we show the button to set up the proxy (in the automatic connected mode setup for example). It's maybe worth adding the info there as well.

Would be great to verify with PM if the changes are acceptable or not

PR #1178, opened 2024-09-23
SLI-1611 Tune the Garbage Collector to avoid high memory usage

nquinquenel:
[SLI-1611](https://sonarsource.atlassian.net/browse/SLI-1611)

[SLI-1611]: https://sonarsource.atlassian.net/browse/SLI-1611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarlint/intellij/core/BackendService.kt
+++ b/src/main/java/org/sonarlint/intellij/core/BackendService.kt
@@ -273,7 +273,11 @@ class BackendService : Disposable {
         sloopPath.toFile().walkTopDown().forEach { file ->
             getService(GlobalLogOutput::class.java).log(file.absolutePath, ClientLogOutput.Level.INFO)
         }
-        return sloopLauncher.start(sloopPath, Paths.get(jreHomePath))
+        return sloopLauncher.start(
+            sloopPath,
+            Paths.get(jreHomePath),
+            "-Xms384m -XX:+UseG1GC -XX:MaxHeapFreeRatio=20 -XX:MinHeapFreeRatio=10 -XX:+UseStringDeduplication -XX:MaxGCPauseMillis=50 -XX:ParallelGCThreads=2"
+        )
     }
 
     private fun listenForProcessExit(sloopProcess: Sloop) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -273,7 +273,11 @@ class BackendService : Disposable {
         sloopPath.toFile().walkTopDown().forEach { file ->
             getService(GlobalLogOutput::class.java).log(file.absolutePath, ClientLogOutput.Level.INFO)
         }
-        return sloopLauncher.start(sloopPath, Paths.get(jreHomePath))
+        return sloopLauncher.start(
+            sloopPath,
+            Paths.get(jreHomePath),
+            "-Xms512m -XX:+UseG1GC -XX:MaxHeapFreeRatio=50 -XX:MinHeapFreeRatio=20"
```
eray-felek-sonarsource, line 279: How did we decide on these values?

nquinquenel: I actually did further tuning and found better results with the following:

`-Xms384m`: SonarLint typically requires at least 200 to 300 MB to start, and grows to 400 or 500 MB during first analysis. Settings this value allows to avoid heap grow cycles at the beginning
`-XX:+UseG1GC`: This Garbage Collector was found to be the most suitable for our need (low latency & keep memory under control)
`-XX:MaxHeapFreeRatio=20 -XX:MinHeapFreeRatio=10`: This allows to reduce the heap size without degrading the performance too much (it was tested against other values)
`-XX:+UseStringDeduplication`: As we store a lot of String with the rules that might be duplicated with multiple connections, I think this is worth to use
`-XX:MaxGCPauseMillis=50`: For performance reasons, default is 200, at 50 it seems to perform well
`-XX:ParallelGCThreads=2`: Limit the number of threads used during GC, we don't want SonarLint to use all the threads available for performance reasons

eray-felek-sonarsource (approved):

PR #1183, opened 2024-09-27
SLI-1630 Download project list on background thread

eray-felek-sonarsource:
[SLI-1630](https://sonarsource.atlassian.net/browse/SLI-1630)

[SLI-1630]: https://sonarsource.atlassian.net/browse/SLI-1630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarlint/intellij/tasks/ServerDownloadProjectTask.java
+++ b/src/main/java/org/sonarlint/intellij/tasks/ServerDownloadProjectTask.java
@@ -31,6 +31,8 @@ import org.sonarlint.intellij.core.BackendService;
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.connection.projects.SonarProjectDto;
 
 import static org.sonarlint.intellij.common.util.SonarLintUtils.getService;
+import static org.sonarlint.intellij.util.ProgressUtils.waitForFuture;
+import static org.sonarlint.intellij.util.ThreadUtilsKt.computeOnPooledThreadWithoutCatching;
 
 public class ServerDownloadProjectTask extends Task.WithResult<Map<String, SonarProjectDto>, Exception> {
   private final ServerConnection server;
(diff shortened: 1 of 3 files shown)


src/main/java/org/sonarlint/intellij/tasks/ServerDownloadProjectTask.java
```diff
@@ -42,15 +44,17 @@ public ServerDownloadProjectTask(Project project, ServerConnection server) {
...
-    try {
-      return getService(BackendService.class)
-        .getAllProjects(server).get()
-        .getSonarProjects()
-        .stream().collect(Collectors.toMap(SonarProjectDto::getKey, p -> p));
-    } catch (Exception e) {
-      SonarLintConsole.get(myProject).error("Failed to download list of projects", e);
-      throw e;
-    }
+    return computeOnPooledThread("Download Projects", () -> {
```
nquinquenel, line 47: You can use `myProject` with `computeOnPooledThread`, to ensure more safety

src/main/java/org/sonarlint/intellij/tasks/ServerDownloadProjectTask.java
```diff
@@ -42,15 +44,17 @@ public ServerDownloadProjectTask(Project project, ServerConnection server) {
 
   @Override
   protected Map<String, SonarProjectDto> compute(@NotNull ProgressIndicator indicator) throws Exception {
```
nquinquenel, line 46: You can probably remove the `throws Exception`

nquinquenel (approved):
LGTM, just small comments

src/main/java/org/sonarlint/intellij/tasks/ServerDownloadProjectTask.java
```diff
@@ -42,15 +44,17 @@ public ServerDownloadProjectTask(Project project, ServerConnection server) {
...
-        .getAllProjects(server).get()
-        .getSonarProjects()
-        .stream().collect(Collectors.toMap(SonarProjectDto::getKey, p -> p));
-    } catch (Exception e) {
-      SonarLintConsole.get(myProject).error("Failed to download list of projects", e);
-      throw e;
-    }
+    return computeOnPooledThread("Download Projects", () -> {
+      try {
+        return waitForFuture(indicator, getService(BackendService.class)
```
nquinquenel, line 49: It would be better to put the `computeOnPooledThread` just around the `waitForFuture`, inside the try/catch

src/main/java/org/sonarlint/intellij/tasks/ServerDownloadProjectTask.java
nquinquenel, line 46: `throws Exception` is likely to be useless, you could remove it

src/main/java/org/sonarlint/intellij/tasks/ServerDownloadProjectTask.java
nquinquenel: Actually this is not that better, ideally the catch should be outside, as it is catching any exception and it returns the error message to the dialog in the Wizard (so that the user can understand). But currently, the computeOnPooledThread is catching any exception as well and just returns null. I think it would be best if we add a new computeOnPooledThreadWithoutCatching, something like that, that would use a custom waitForTask that would throw the exception and not just catch it, so it can go back to the original call.
Probably a similar issue in other xxxTask where we used computeOnPooledThread

nquinquenel (approved):
LGTM

PR #1186, opened 2024-09-30
SLI-1633 Cancel previous redundant analysis when triggering a new one

nquinquenel:
[SLI-1633](https://sonarsource.atlassian.net/browse/SLI-1633)

[SLI-1633]: https://sonarsource.atlassian.net/browse/SLI-1633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/sonarlint/intellij/trigger/TriggerType.kt
```diff
@@ -0,0 +1,50 @@
...
+    BINDING_UPDATE("Binding update", true),
+    OPEN_FINDING("Open finding", true),
+    SERVER_SENT_EVENT("Server-sent event", false);
+
+    companion object {
+        // Forced analysis that will appear in the report tab
+        val analysisSnapshot = listOf(RIGHT_CLICK, ALL, CHANGED_FILES)
+
+        // Non snapshot do not include CHECK_IN analysis and OPEN_FINDING analysis, as they are specific cases
+        val nonAnalysisSnapshot =
```
eray-felek-sonarsource, line 42: I would prefer a name like "dynamicAnalaysis"

src/main/java/org/sonarlint/intellij/analysis/RunningAnalysesTracker.kt
```diff
@@ -44,6 +45,22 @@ class RunningAnalysesTracker {
...
+            it.cancel()
+            finish(it)
+        }
+    }
+
+    fun cancelSimilarAnalysis(analysisState: AnalysisState, console: SonarLintConsole) {
+        for (analysis in analysisStateById.values) {
+            if (analysis.isRedundant(analysisState)) {
+                console.info("Cancelling analysis ${analysis.id}")
+                analysis.cancel()
```
eray-felek-sonarsource, line 59: Shouldn't we cancel the older ones, looks like we are cancelling the new one here

src/test/java/org/sonarlint/intellij/analysis/AnalysisStateTests.kt
```diff
@@ -0,0 +1,106 @@
...
+    private lateinit var randomUuid2: UUID
+
+    @BeforeEach
+    fun init() {
+        randomUuid1 = UUID.randomUUID()
+        randomUuid2 = UUID.randomUUID()
+    }
+
+    @Test
+    fun should_find_redundant_analysis_for_non_snapshot() {
```
eray-felek-sonarsource, line 48: I think we could add a medium test that starts an actual analysis to make sure we are canceling the old analysis

src/main/java/org/sonarlint/intellij/trigger/TriggerType.kt
nquinquenel: My only concern is that `CURRENT_FILE_ACTION` is a forced analysis, so non dynamic, but it's still part of this list "dynamicAnalysis". The idea is really to differentiate `Current file tab` to the `Report tab` (which ultimately boils down to being a snapshot vs non snapshot). I'm not a big fan of dynamic but if you have other proposition I'm open to them.

src/main/java/org/sonarlint/intellij/analysis/RunningAnalysesTracker.kt
nquinquenel: We are iterating over `analysisStateById`, which contains all the running analysis, and we are cancelling each of them that are considered redundant against the one passed in the parameters

src/test/java/org/sonarlint/intellij/analysis/AnalysisStateTests.kt
nquinquenel: I tried but I had difficulties making a test that triggered a long enough analysis that I could cancel it with a new one

eray-felek-sonarsource (approved):

PR #1205, opened 2024-10-17
SLI-1653 SLI-1652 SLI-1650 SLI-1649 Analyzer update for 10.12

nquinquenel:


src/main/resources/META-INF/plugin.xml
```diff
@@ -64,6 +64,7 @@
 
     <change-notes><![CDATA[
       <ul>
+        <li>10.12 - Support of Java 22 analysis. Improvements on Helm and Kubernetes analyzers. Bug fixes, fewer FPs and improvements for many languages.</li>
```
eray-felek-sonarsource, line 67: I am not sure if "Improvements on Helm and Kubernetes analyzers" is useful information for our users. We can either be more specific or opt-out mentioning it

nquinquenel: Ok, I think we can remove it then, I believe it's mostly technical stuff, release note is not really clear

eray-felek-sonarsource (approved):

PR #1211, opened 2024-10-31
SLI-1657 Allow displaying both standard and MQR rule descriptions

nquinquenel:


src/main/java/org/sonarlint/intellij/SonarLintIcons.kt
```diff
@@ -82,12 +82,42 @@ object SonarLintIcons {
...
+
     private val SEVERITY_ICONS = mapOf(
-        IssueSeverity.BLOCKER to getIcon("/images/severity/blocker.svg"),
-        IssueSeverity.CRITICAL to getIcon("/images/severity/critical.svg"),
-        IssueSeverity.INFO to getIcon("/images/severity/info.svg"),
-        IssueSeverity.MAJOR to getIcon("/images/severity/major.svg"),
-        IssueSeverity.MINOR to getIcon("/images/severity/minor.svg")
+        IssueSeverity.BLOCKER to getIcon("/images/impact/blocker.svg"),
+        IssueSeverity.CRITICAL to getIcon("/images/impact/high.svg"),
+        IssueSeverity.MAJOR to getIcon("/images/impact/medium.svg"),
```
eray-felek-sonarsource, line 118: We could align the image name i.e major ->impact/major

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -515,9 +515,9 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
         }
         val connection = connectionOpt.get()
         return if (connection.token != null) {
-            Either.forLeft(TokenDto(connection.token))
+            Either.forLeft(TokenDto(connection.token!!))
```
eray-felek-sonarsource, line 518: Can they be null or are we handling the exception if they are?

src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -445,8 +445,8 @@ class BackendService : Disposable {
     }
 
     fun getAllProjects(server: ServerConnection): CompletableFuture<GetAllProjectsResponse> {
-        val credentials: Either<TokenDto, UsernamePasswordDto> = server.token?.let { Either.forLeft(TokenDto(server.token)) }
-            ?: Either.forRight(UsernamePasswordDto(server.login, server.password))
+        val credentials: Either<TokenDto, UsernamePasswordDto> = server.token?.let { Either.forLeft(TokenDto(server.token!!)) }
+            ?: Either.forRight(UsernamePasswordDto(server.login!!, server.password!!))
```
eray-felek-sonarsource, line 449: same

src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -741,8 +735,8 @@ class BackendService : Disposable {
     }
 
     fun checkSmartNotificationsSupported(server: ServerConnection): CompletableFuture<CheckSmartNotificationsSupportedResponse> {
-        val credentials: Either<TokenDto, UsernamePasswordDto> = server.token?.let { Either.forLeft(TokenDto(server.token)) }
-            ?: Either.forRight(UsernamePasswordDto(server.login, server.password))
+        val credentials: Either<TokenDto, UsernamePasswordDto> = server.token?.let { Either.forLeft(TokenDto(server.token!!)) }
+            ?: Either.forRight(UsernamePasswordDto(server.login!!, server.password!!))
```
eray-felek-sonarsource, line 739: same

src/main/java/org/sonarlint/intellij/SonarLintIcons.kt
nquinquenel: Actually, the IssueSeverity shares the same icon as the ImpactSeverity (Standard/MQR mode), so I cannot really have the same names for both 😄

src/main/java/org/sonarlint/intellij/ui/vulnerabilities/tree/render/LocalTaintVulnerabilityRenderer.kt
```diff
@@ -48,25 +49,28 @@ object LocalTaintVulnerabilityRenderer : NodeRenderer<LocalTaintVulnerability> {
                 CompoundIcon(
                     CompoundIcon.Axis.X_AXIS,
                     gap,
-                    SonarLintIcons.impact(impact))
+                    SonarLintIcons.impact(impact!!)
```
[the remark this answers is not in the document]
nquinquenel: I'm checking just above for `node.getHighestImpact() != null`

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
nquinquenel: Normally it should not happen in our Wizard, it should be either token or password/login

nquinquenel: I reverted login/password `!!`, for token it's fine as we still check before

src/main/java/org/sonarlint/intellij/SonarLintIcons.kt
eray-felek-sonarsource: We could duplicate the images and store them under /images/severity/ folder so that if we need to change them in the future we would be confident it won't break the other icon

eray-felek-sonarsource (approved):

PR #1240, opened 2024-11-27
SLI-1741 Avoid sending file content for any file event in the project

nquinquenel:
[SLI-1741](https://sonarsource.atlassian.net/browse/SLI-1741)

[SLI-1741]: https://sonarsource.atlassian.net/browse/SLI-1741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -275,6 +275,7 @@ tasks {
     prepareTestingSandbox {
         doLast {
             copyPlugins(destinationDir, pluginName)
+            renameCsharpPlugins(destinationDir, pluginName)
             copyOmnisharp(destinationDir, pluginName)
             copySloop(destinationDir, pluginName)
         }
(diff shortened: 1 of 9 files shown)


eray-felek-sonarsource (approved):
SLI-1738 looks like analyzer update, is it a mistake?
Else LGTM

build.gradle.kts
```diff
@@ -275,6 +275,7 @@ tasks {
     prepareTestingSandbox {
         doLast {
             copyPlugins(destinationDir, pluginName)
+            renameCsharpPlugins(destinationDir, pluginName)
```
eray-felek-sonarsource, line 278: Why did we need this?

nquinquenel:
> [SLI-1738](https://sonarsource.atlassian.net/browse/SLI-1738) looks like analyzer update, is it a mistake? Else LGTM

Oops! I wanted to modify the title of another PR 😄

build.gradle.kts
nquinquenel: With the new C# repackaging, we rename the plugins in the /plugins/ directory so we can easily reference it in the code (with a static variable). This was done for the normal environment, but in a test environment I forgot to add this renaming, so I noticed some errors in some tests because the plugin couldn't be loaded. It didn't break the CI probably because we don't have a unit test for C#.

PR #1250, opened 2024-12-02
Use latest SLCORE version

nquinquenel:


eray-felek-sonarsource (commented):
There are failing QAs

PR #1283, opened 2025-01-21
SLI-1795 SLI-1798 SLI-1797 SLI-1799 SLI-1783 SLI-1802 Analyzer updates #2 for 10.13

nquinquenel:


src/main/resources/META-INF/plugin.xml
```diff
@@ -61,7 +61,7 @@
 
     <change-notes><![CDATA[
       <ul>
-        <li>10.15 - 1 new QuickFix for Java. Bug fixes, fewer FPs and improvements for many languages.</li>
+        <li>10.15 - 1 new QuickFix for Java. Fixed a performance regression on Python. Bug fixes, fewer FPs and improvements for many languages.</li>
```
eray-felek-sonarsource, line 64: Instead of "Fixed a performance regression on Python" we could use "Fix for a performance regression on Python"

eray-felek-sonarsource (approved):

PR #1286, opened 2025-01-23
SLI-1805 Avoid non-needed analysis triggering

eray-felek-sonarsource:
[SLI-1805](https://sonarsource.atlassian.net/browse/SLI-1805)

[SLI-1805]: https://sonarsource.atlassian.net/browse/SLI-1805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.java
+++ b/src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.java
@@ -93,14 +93,14 @@ public final class AnalysisSubmitter {
     TaskRunnerKt.startBackgroundableModalTask(project, ANALYSIS_TASK_TITLE, analysis::run);
   }
 
-  public void autoAnalyzeOpenFilesForModule(TriggerType triggerType, @Nullable Module module) {
+  public void autoAnalyzeSelectedFilesForModule(TriggerType triggerType, @Nullable Module module) {
     if (module == null) {
-      autoAnalyzeOpenFiles(triggerType);
+      autoAnalyzeSelectedFiles(triggerType);
       return;
     }
 
-    var openFiles = FileEditorManager.getInstance(project).getOpenFiles();
-    var filesToAnalyze = Arrays.stream(openFiles)
+    var selectedFiles = FileEditorManager.getInstance(project).getSelectedFiles();
+    var filesToAnalyze = Arrays.stream(selectedFiles)
       .filter(file -> module.equals(findModuleForFile(file, project)))
       .toList();
 
@@ -109,10 +109,11 @@ public final class AnalysisSubmitter {
     }
   }
 
-  public void autoAnalyzeOpenFiles(TriggerType triggerType) {
-    var openFiles = FileEditorManager.getInstance(project).getOpenFiles();
-    if (openFiles.length > 0) {
-      autoAnalyzeFiles(List.of(openFiles), triggerType);
+  public void autoAnalyzeSelectedFiles(TriggerType triggerType) {
+    var selectedFiles = FileEditorManager.getInstance(project).getSelectedFiles();
+
+    if (selectedFiles.length > 0) {
+      autoAnalyzeFiles(List.of(selectedFiles), triggerType);
     }
   }
 
@@ -122,6 +123,9 @@ public final class AnalysisSubmitter {
       return null;
     }
     var callback = new UpdateOnTheFlyFindingsCallable(onTheFlyFindingsHolder);
+    if (triggerType != TriggerType.SELECTION_CHANGED) {
+      onTheFlyFindingsHolder.clearNonDirtyAnalyzedFiles();
+    }
     return analyzeInBackground(files, triggerType, callback);
   }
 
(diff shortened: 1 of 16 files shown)


src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.java
```diff
@@ -121,8 +122,21 @@ public Cancelable autoAnalyzeFiles(Collection<VirtualFile> files, TriggerType tr
...
+      .filter(file -> !onTheFlyFindingsHolder.getNonDirtyAnalyzedFiles().contains(file))
+      .toList();
+
+    if (filesToAnalyze.isEmpty()) {
+      return null;
+    }
+
     var callback = new UpdateOnTheFlyFindingsCallable(onTheFlyFindingsHolder);
-    return analyzeInBackground(files, triggerType, callback);
+    var cancelable = analyzeInBackground(filesToAnalyze, triggerType, callback);
```
nquinquenel, line 135: You can probably return this directly and not put it in a variable

src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.java
```diff
@@ -121,8 +122,21 @@ public Cancelable autoAnalyzeFiles(Collection<VirtualFile> files, TriggerType tr
     if (!getGlobalSettings().isAutoTrigger()) {
       return null;
     }
+
+    var filesToAnalyze = files.stream()
+      .filter(file -> !onTheFlyFindingsHolder.getNonDirtyAnalyzedFiles().contains(file))
```
nquinquenel, line 127: I don't think we should do this check here; it will prevent many events from performing a new analysis. `autoAnalyzeSelectedFiles` is called from many places: build finished, settings changed, rule excluded, etc. In such cases, we should always trigger a new analysis on all the selected files. Clearing the non-dirty files list in every location may be complex.

We should be thinking in the reverse way: when do we want to take into account dirty files? Probably just during the selection changed event?

Also, `EditorOpenTrigger` is probably not relevant anymore

src/main/java/org/sonarlint/intellij/analysis/OnTheFlyFindingsHolder.kt
```diff
@@ -111,11 +117,13 @@ class OnTheFlyFindingsHolder(private val project: Project) : FileEditorManagerLi
     }
 
     override fun selectionChanged(event: FileEditorManagerEvent) {
+        getService(project, AnalysisSubmitter::class.java).autoAnalyzeSelectedFiles(TriggerType.SELECTION_CHANGED)
```
nquinquenel, line 120: Following my other comment, I would filter non dirty files here

nquinquenel (commented):
It's working, but I think some cases are broken, for example, updating the settings, after building, or when excluding a rule, it's not triggering an analysis again. CF my comment below. I think it still need a  bit of tuning

src/main/java/org/sonarlint/intellij/analysis/OnTheFlyFindingsHolder.kt
```diff
@@ -51,6 +53,10 @@ class OnTheFlyFindingsHolder(private val project: Project) : FileEditorManagerLi
             .subscribe(FileEditorManagerListener.FILE_EDITOR_MANAGER, this)
     }
 
+    fun getNonDirtyAnalyzedFiles(): Set<VirtualFile> = nonDirtyAnalyzedFiles
```
nquinquenel, line 56: I think this function can be removed, it's just the reference to `nonDirtyAnalyzedFiles`

src/main/java/org/sonarlint/intellij/analysis/OnTheFlyFindingsHolder.kt
```diff
@@ -51,6 +53,10 @@ class OnTheFlyFindingsHolder(private val project: Project) : FileEditorManagerLi
...
+    fun clearNonDirtyAnalyzedFiles() = nonDirtyAnalyzedFiles.clear()
+    private fun removeNonDirtyAnalyzedFile(file: VirtualFile) = nonDirtyAnalyzedFiles.remove(file)
```
nquinquenel, line 58: I'd also just call `nonDirtyAnalyzedFiles.remove(file)` directly, instead of having this function, it's just used at one location

src/main/java/org/sonarlint/intellij/analysis/OnTheFlyFindingsHolder.kt
```diff
@@ -111,11 +117,18 @@ class OnTheFlyFindingsHolder(private val project: Project) : FileEditorManagerLi
     }
 
     override fun selectionChanged(event: FileEditorManagerEvent) {
-        selectedFile = event.newFile
+        val file = event.newFile
```
nquinquenel, line 120: You can probably remove this variable and just use the `selectedFile` that is assigned just below

nquinquenel (commented):
Looks good, just a few minor comments left

src/main/java/org/sonarlint/intellij/analysis/OnTheFlyFindingsHolder.kt
eray-felek-sonarsource: I would prefer to keep it this way since there are other places changing the selectedFile variable

nquinquenel (approved):

PR #1292, opened 2025-01-28
SLI-1816 Interrupting the branch matching computation should not cause an error

eray-felek-sonarsource:
[SLI-1816](https://sonarsource.atlassian.net/browse/SLI-1816)

[SLI-1816]: https://sonarsource.atlassian.net/browse/SLI-1816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
+++ b/src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
@@ -60,6 +60,7 @@ import java.security.cert.X509Certificate
 import java.util.UUID
 import java.util.concurrent.CancellationException
 import java.util.concurrent.CompletableFuture
+import java.util.concurrent.TimeoutException
 import org.apache.commons.text.StringEscapeUtils
 import org.eclipse.lsp4j.jsonrpc.ResponseErrorException
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseError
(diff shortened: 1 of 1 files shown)


nquinquenel (changes requested):
I think we should just catch the interruption at the future level, not the whole method

nquinquenel (approved):
LGTM

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -644,11 +649,27 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
...
+                } catch (e: InterruptedException) {
+                    getService(project, SonarLintConsole::class.java).error(INTERRUPTED_MESSAGE, e)
+                } catch (e: TimeoutException) {
+                    getService(project, SonarLintConsole::class.java).error(TIMEOUT_MESSAGE, e)
+                }
             }
         })
-        return computeOnPooledThread(project, "Waiting for branch matching result") { resultFuture.get() }
+        return computeOnPooledThread(project, "Waiting for branch matching result") {
+            try {
```
nquinquenel, line 663: I don't think you need this try/catch here?

src/main/java/org/sonarlint/intellij/SonarLintIntelliJClient.kt
```diff
@@ -644,11 +649,27 @@ object SonarLintIntelliJClient : SonarLintRpcClientDelegate {
             PerformInBackgroundOption.ALWAYS_BACKGROUND
         ) {
             override fun run(indicator: ProgressIndicator) {
-                val result = repo.electBestMatchingServerBranchForCurrentHead(mainBranchName, allBranchesNames) ?: mainBranchName
-                resultFuture.complete(result)
+                try {
```
nquinquenel, line 652: I think you can reuse `FutureUtils.waitForTask`, it seems very similar

PR #1295, opened 2025-01-30
SLI-1827 Rely on NodeJs from IDE configuration

nquinquenel:
[SLI-1827](https://sonarsource.atlassian.net/browse/SLI-1827)

[SLI-1827]: https://sonarsource.atlassian.net/browse/SLI-1827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


damien-urruty-sonarsource (commented):
The changes look good, but I wonder: is it possible that the IDE Node.js path does not meet our version requirements? I didn't see any check around that, so it means we could set a path for the user that would fail later with SonarJS. Should we add this check? There is maybe something to expose from the core to be able to do that

nquinquenel:
> is it possible that the IDE Node.js path does not meet our version requirements? I didn't see any check around that, so it means we could set a path for the user that would fail later with SonarJS. Should we add this check? There is maybe something to expose from the core to be able to do that

It's definitely possible, but when the analysis happens, an error message will pop up. It's not any different today: when the user sets a Node.js path, there's nothing preventing him from setting an outdated path. It would probably be a good idea to have a check on the SLCORE side. I think it can be added later as an improvement, the goal here was really to focus on using the IDE's setting.

damien-urruty-sonarsource (approved):

damien-urruty-sonarsource:
One last comment: the setting is picked from the IDE only if the user visits the settings popup. By default, we don't use the IDE setting, which might not be optimal if we want to favor a config-less approach. A consequence of that: if the IDE setting changes, our setting won't change automatically. I think there is still room for improvement on this topic, let's plan  a 2nd round?

PR #1309, opened 2025-02-12
SLI-1851 SLI-1854 Support US Region

eray-felek-sonarsource:


common/src/main/java/org/sonarlint/intellij/common/util/SonarLintUtils.java
```diff
@@ -45,7 +45,7 @@ public class SonarLintUtils {
   public static final String DEFAULT_SONARCLOUD_URL = "https://sonarcloud.io";
   public static final String SONARCLOUD_URL = System.getProperty("sonarlint.internal.sonarcloud.url", DEFAULT_SONARCLOUD_URL);
   private static final Set<String> SONARCLOUD_ALIAS = Set.copyOf(List.of("https://sonarqube.com", "https://www.sonarqube.com",
-    "https://www.sonarcloud.io", DEFAULT_SONARCLOUD_URL, SONARCLOUD_URL));
+    "https://www.sonarcloud.io", "https://us-sc-staging.io", DEFAULT_SONARCLOUD_URL, SONARCLOUD_URL));
```
nquinquenel, line 48: To be updated

src/main/java/org/sonarlint/intellij/config/global/wizard/ServerStep.java
```diff
@@ -46,11 +47,12 @@
 import org.sonarlint.intellij.SonarLintIcons;
 import org.sonarlint.intellij.common.util.SonarLintUtils;
 import org.sonarlint.intellij.telemetry.LinkTelemetry;
+import org.sonarsource.sonarlint.core.SonarCloudRegion;
 
 import static org.sonarlint.intellij.common.util.SonarLintUtils.SONARCLOUD_URL;
 import static org.sonarlint.intellij.telemetry.LinkTelemetry.SONARCLOUD_FREE_SIGNUP_PAGE;
 
-public class ServerStep extends AbstractWizardStepEx {
+public class ServerStep extends AbstractWizardStepEx{
```
nquinquenel, line 55:
```suggestion
public class ServerStep extends AbstractWizardStepEx {
```

src/main/java/org/sonarlint/intellij/config/global/wizard/ServerStep.java
```diff
@@ -170,6 +182,9 @@ private void load(boolean editing) {
   private void selectionChanged() {
     boolean sq = radioSonarQube.isSelected();
 
+    radioEU.setEnabled(!sq);
+    radioUS.setEnabled(!sq);
```
nquinquenel, line 186: I think we should set one of the radio button to the right value when loading the panel

src/main/java/org/sonarlint/intellij/config/global/wizard/WizardModel.java
```diff
@@ -29,6 +29,7 @@
 import org.sonarlint.intellij.tasks.CheckNotificationsSupportedTask;
 import org.sonarlint.intellij.tasks.GetOrganizationTask;
 import org.sonarlint.intellij.tasks.GetOrganizationsTask;
+import org.sonarsource.sonarlint.core.SonarCloudRegion;
```
nquinquenel, line 32: That's a strange package location, we usually rely on `.core.rpc.protocol.` or `.core.client.utils.` for example

src/main/java/org/sonarlint/intellij/config/global/wizard/WizardModel.java
```diff
@@ -37,13 +38,23 @@ public class WizardModel {
   private ServerType serverType;
   private String serverUrl;
   private String token;
+
+  public SonarCloudRegion getRegion() {
```
nquinquenel, line 42: Why are these getter/setter in the middle of the class attributes? 😅

src/main/java/org/sonarlint/intellij/config/global/wizard/WizardModel.java
```diff
@@ -59,6 +70,9 @@ public WizardModel() {
   public WizardModel(ServerConnection connectionToEdit) {
     if (SonarLintUtils.isSonarCloudAlias(connectionToEdit.getHostUrl())) {
       serverType = ServerType.SONARCLOUD;
+      if(region != null) {
```
nquinquenel, line 73:
```suggestion
      if (region != null) {
```

nquinquenel, line 73: What if region is null?

src/main/java/org/sonarlint/intellij/config/global/wizard/WizardModel.java
```diff
@@ -251,8 +265,13 @@ private ServerConnection.Builder createUnauthenticatedConnection(@Nullable Strin
       .setName(name);
 
     if (serverType == ServerType.SONARCLOUD) {
-      builder.setHostUrl(SONARCLOUD_URL);
-
+      if (region == SonarCloudRegion.US) {
+        builder.setHostUrl(SonarCloudRegion.US.getProductionUri().toString());
+        builder.setRegion("US");
+      } else {
+        builder.setHostUrl(SONARCLOUD_URL);
```
nquinquenel, line 272: There's a mix of SQ:C URLs defined on the client side, and on the server side, we should maybe make it cleaner

src/main/java/org/sonarlint/intellij/config/global/wizard/WizardModel.java
```diff
@@ -251,8 +265,13 @@ private ServerConnection.Builder createUnauthenticatedConnection(@Nullable Strin
       .setName(name);
 
     if (serverType == ServerType.SONARCLOUD) {
-      builder.setHostUrl(SONARCLOUD_URL);
-
+      if (region == SonarCloudRegion.US) {
```
nquinquenel, line 268: Couldn't we make something automatic based on the region object? It would set the `region` and `hostUrl` via getter on this object

src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -525,6 +528,11 @@ class BackendService : Disposable {
         return SonarCloudConnectionConfigurationDto(
             createdConnection.name,
             createdConnection.organizationKey!!,
+            if(createdConnection.region != null) {
```
nquinquenel, line 531:
```suggestion
            if (createdConnection.region != null) {
```

src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -160,6 +161,7 @@ import org.sonarsource.sonarlint.core.rpc.protocol.common.UsernamePasswordDto
 import org.sonarsource.sonarlint.plugin.api.module.file.ModuleFileEvent
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.issue.CheckStatusChangePermittedParams as issueCheckStatusChangePermittedParams
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.issue.CheckStatusChangePermittedResponse as issueCheckStatusChangePermittedResponse
+import  org.sonarsource.sonarlint.core.rpc.protocol.common.SonarCloudRegion as rpcSonarCloudRegion
```
nquinquenel, line 164:
```suggestion
import org.sonarsource.sonarlint.core.rpc.protocol.common.SonarCloudRegion as rpcSonarCloudRegion
```

nquinquenel: I don't think you have the right codestyle in your IDE

src/main/java/org/sonarlint/intellij/core/BackendService.kt
nquinquenel: Could use `?.let { } ?:` for something bit more elegant 😄

src/main/java/org/sonarlint/intellij/tasks/ConnectionTestTask.kt
```diff
@@ -36,8 +37,10 @@ class ConnectionTestTask(private val server: ServerConnection) :
     override fun compute(indicator: ProgressIndicator): ValidateConnectionResponse? {
         indicator.text = "Connecting to " + server.hostUrl + "\u2026"
         indicator.isIndeterminate = true
+        val region = if(server.region != null) SonarCloudRegion.valueOf(server.region) else SonarCloudRegion.EU
```
nquinquenel, line 40:
```suggestion
        val region = if (server.region != null) SonarCloudRegion.valueOf(server.region) else SonarCloudRegion.EU
```

nquinquenel: Same here, maybe you can use `?.let { } ?:` structure

src/main/java/org/sonarlint/intellij/tasks/GetOrganizationTask.java
```diff
@@ -42,6 +44,14 @@ public GetOrganizationTask(ServerConnection server, String organizationKey) {
     super(null, "Fetch Organization From SonarQube Cloud", true);
     this.server = server;
     this.organizationKey = organizationKey;
+    this.region = SonarCloudRegion.EU;
+  }
+
+  public GetOrganizationTask(ServerConnection server, String organizationKey, SonarCloudRegion region) {
```
nquinquenel, line 50: Could maybe just have a single constructor with a nullable field? In which case do we not provide a region?

src/main/java/org/sonarlint/intellij/config/global/wizard/WizardModel.java
```diff
@@ -59,6 +70,9 @@ public WizardModel() {
   public WizardModel(ServerConnection connectionToEdit) {
     if (SonarLintUtils.isSonarCloudAlias(connectionToEdit.getHostUrl())) {
       serverType = ServerType.SONARCLOUD;
+      if(region != null) {
+        region = SonarCloudRegion.valueOf(connectionToEdit.getRegion());
```
nquinquenel, line 74: Why do we check if region is null, just to re-assign shortly after? Also, for existing regions, I believe the region is null

src/main/java/org/sonarlint/intellij/config/global/wizard/WizardModel.java
nquinquenel:
```
+--- org.sonarsource.sonarlint.core:sonarlint-rpc-impl:10.15.0.80213
|    +--- org.sonarsource.sonarlint.core:sonarlint-core:10.15.0.80213
```
sonarlint-core dependency is from sonarlint-rpc-impl, and we are going to [remove](https://sonarsource.atlassian.net/browse/SLI-1819) it, so we should make sure we don't use it anymore

src/main/java/org/sonarlint/intellij/config/global/wizard/WizardModel.java
eray-felek-sonarsource: I added US_SONARCLOUD_URL

src/main/java/org/sonarlint/intellij/config/global/wizard/WizardModel.java
eray-felek-sonarsource: For setting the region it's possible but for URL it is not since we want to decide based on a condition

src/main/java/org/sonarlint/intellij/config/global/wizard/ServerStep.java
eray-felek-sonarsource: It is set on the ServerStep.form

src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -769,39 +778,45 @@ class BackendService : Disposable {
         notifyBackend { it.sonarProjectBranchService.didVcsRepositoryChange(DidVcsRepositoryChangeParams(projectId(project))) }
     }
 
-    fun checkSmartNotificationsSupported(server: ServerConnection): CompletableFuture<CheckSmartNotificationsSupportedResponse> {
+    fun checkSmartNotificationsSupported(server: ServerConnection, region: SonarCloudRegion = SonarCloudRegion.EU):
```
nquinquenel, line 781: Same here for the region param, and all subsequent methods

src/main/java/org/sonarlint/intellij/sharing/AutomaticSharedConfigCreator.kt
```diff
@@ -241,23 +261,38 @@ class AutomaticSharedConfigCreator(
...
+                    GridBagConstraints(1, 7, 1, 1, 1.0, 0.0, GridBagConstraints.WEST, GridBagConstraints.HORIZONTAL, JBUI.emptyInsets(), 0, 0)
+                )
+            }
         }
 
         connectionNameLabel.text = "Connection name"
         centerPanel.add(
             connectionNameLabel,
-            GridBagConstraints(1, 6, 1, 1, 1.0, 0.0, GridBagConstraints.WEST, GridBagConstraints.HORIZONTAL, JBUI.emptyInsets(), 0, 0)
+            GridBagConstraints(1, 8, 1, 1, 1.0, 0.0, GridBagConstraints.WEST, GridBagConstraints.HORIZONTAL, JBUI.emptyInsets(), 0, 0)
```
nquinquenel, line 284: Offset will be wrong if not in dogfood

src/main/java/org/sonarlint/intellij/tasks/CheckNotificationsSupportedTask.java
```diff
@@ -35,12 +36,14 @@
  */
 public class CheckNotificationsSupportedTask extends Task.Modal {
   private final ServerConnection connection;
+  private final SonarCloudRegion region;
   private Exception exception;
   private boolean notificationsSupported = false;
 
-  public CheckNotificationsSupportedTask(ServerConnection connection) {
+  public CheckNotificationsSupportedTask(ServerConnection connection, SonarCloudRegion region) {
```
nquinquenel, line 43: I don't think you need the region, it's already in `ServerConnection`

nquinquenel (changes requested):
I think the general idea is good. However, I noticed some mistakes in design, code style, or logic. I'll let you review the comments.

Also, I think there is a lot of duplicated logic about the region. We could perhaps create a new class that would handle all this logic. For instance, we often check for the following:

```
                    if (region == SonarCloudRegion.US) {
                        US_SONARCLOUD_URL
                    } else {
                        SONARCLOUD_URL
                    }
```

Instead, I think we could simply have a function that would handle this logic and return the right value. In the future, if we add more regions, it will be complex to refactor everything without mistake.

src/main/java/org/sonarlint/intellij/config/global/ServerConnectionMgmtPanel.java
```diff
@@ -127,18 +129,41 @@ public void mouseClicked(MouseEvent evt) {
       @Override
       protected void customizeCellRenderer(JList list, ServerConnection server, int index, boolean selected, boolean hasFocus) {
         if (server.isSonarCloud()) {
```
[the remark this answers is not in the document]
eray-felek-sonarsource: I think we can get rid of if (!server.isSonarCloud()) but not sure how we could further simplify

eray-felek-sonarsource:
> I think the general idea is good. However, I noticed some mistakes in design, code style, or logic. I'll let you review the comments.
> 
> Also, I think there is a lot of duplicated logic about the region. We could perhaps create a new class that would handle all this logic. For instance, we often check for the following:
> 
> ```
>                     if (region == SonarCloudRegion.US) {
>                         US_SONARCLOUD_URL
>                     } else {
>                         SONARCLOUD_URL
>                     }
> ```
> 
> Instead, I think we could simply have a function that would handle this logic and return the right value. In the future, if we add more regions, it will be complex to refactor everything without mistake.

Added a RegionUtils class

src/main/java/org/sonarlint/intellij/config/global/ServerConnectionMgmtPanel.java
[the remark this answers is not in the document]
nquinquenel: I shared the snippet of the simplification, I think it's more readeable

src/main/java/org/sonarlint/intellij/config/global/wizard/ServerStep.java
```diff
@@ -70,13 +72,18 @@ public class ServerStep extends AbstractWizardStepEx {
...
+  private JLabel sonarCloudUrl;
 
   public ServerStep(WizardModel model, boolean editing, Collection<String> existingNames) {
     super("Server Details");
     this.model = model;
     this.existingNames = existingNames;
     radioSonarCloud.addChangeListener(e -> selectionChanged());
     radioSonarQube.addChangeListener(e -> selectionChanged());
+    radioEU.addChangeListener(e -> regionSelectionChanged());
+    radioUS.addChangeListener(e -> regionSelectionChanged());
```
nquinquenel, line 86: I guess you don't need a listener if it doesn't do anything more than the default behavior

src/main/java/org/sonarlint/intellij/sharing/AutomaticSharedConfigCreator.kt
```diff
@@ -147,7 +155,10 @@ class AutomaticSharedConfigCreator(
             if (isSQ) {
                 serverConnectionBuilder.setHostUrl(orgOrServerUrl)
             } else {
-                serverConnectionBuilder.setOrganizationKey(orgOrServerUrl).setHostUrl(SONARCLOUD_URL)
+                serverConnectionBuilder.setOrganizationKey(orgOrServerUrl).setHostUrl(
+                    RegionUtils.getUrlByRegion(region ?: SonarCloudRegion.EU)
+                )
+                serverConnectionBuilder.setRegion(region?.name.toString())
```
nquinquenel, line 161: If region is null you should put the EU region I think

nquinquenel (changes requested):
Looking good! Just a few comments

src/main/java/org/sonarlint/intellij/sharing/AutomaticSharedConfigCreator.kt
```diff
@@ -238,26 +250,44 @@ class AutomaticSharedConfigCreator(
...
-                    1, 5, 1, 1, 1.0, 0.0, GridBagConstraints.WEST, GridBagConstraints.HORIZONTAL, JBUI.insetsBottom(30), 0, 0
+                    1, ++gridY, 1, 1, 1.0, 0.0, GridBagConstraints.WEST, GridBagConstraints.HORIZONTAL, JBUI.insetsBottom(30), 0, 0
                 )
             )
+        } else {
+            if (SonarLintUtils.isDogfoodEnvironment()) {
+                val scURLField = JBTextField(US_SONARCLOUD_URL).apply {
+                    isEditable = false
+                }
+                val scURLFieldText = JBPanel<JBPanel<*>>(BorderLayout()).apply { add(scURLField, BorderLayout.CENTER) }
```
nquinquenel, line 261: I don't think you need to create a JBPanel if it has only one component inside

src/main/java/org/sonarlint/intellij/config/global/AutomaticServerConnectionCreator.kt
```diff
@@ -38,26 +38,31 @@ import java.awt.event.ActionEvent
...
     private val connectionNameField = JBTextField()
+    private val scURLField = JBTextField()
     private val connectedModeDescriptionLabel = SwingHelper.createHtmlViewer(false, null, null, null)
     private val serverUrlOrOrgLabel = SwingHelper.createHtmlViewer(false, null, null, null)
     private val serverUrlOrOrgField = JBTextField(serverOrOrg).apply {
         isEditable = false
     }
     private val redWarningIcon = JBLabel()
     private val warningLabel = SwingHelper.createHtmlViewer(false, null, null, JBUI.CurrentTheme.ContextHelp.FOREGROUND)
+    private val scURLLabel = SwingHelper.createHtmlViewer(false, null, null, null)
```
nquinquenel, line 65: I would group the 2 attributes together, just for clarity

src/main/java/org/sonarlint/intellij/config/global/AutomaticServerConnectionCreator.kt
```diff
@@ -153,32 +162,52 @@ class AutomaticServerConnectionCreator(private val serverOrOrg: String, private
...
+
+        if (!isSQ && SonarLintUtils.isDogfoodEnvironment()) {
+            centerPanel.add(
+                scURLLabel, GridBagConstraints(
+                    1, ++gridY, 1, 1, 1.0, 0.0, GridBagConstraints.WEST, GridBagConstraints.HORIZONTAL, JBUI.emptyInsets(), 0, 0
+                )
+            )
+
+            centerPanel.add(
+                scURLField, GridBagConstraints(
```
nquinquenel, line 185: This field should not be editable

src/main/java/org/sonarlint/intellij/util/RegionUtils.kt
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.sonarlint.intellij.util
+
+import org.sonarlint.intellij.common.util.SonarLintUtils
+import org.sonarsource.sonarlint.core.rpc.protocol.common.SonarCloudRegion
+
+class RegionUtils {
+
+    companion object {
+        @JvmStatic
```
nquinquenel, line 28: You don't need `@JvmStatic`

src/test/java/org/sonarlint/intellij/util/SonarLintUtilsTests.java
```diff
@@ -21,8 +21,12 @@
 
 import com.intellij.openapi.fileTypes.FileType;
 import com.intellij.openapi.vfs.VirtualFile;
+import org.apache.commons.lang3.SystemUtils;
+import org.junit.jupiter.api.Assertions;
```
nquinquenel, line 25: We use `assertj` for assertions

src/test/java/org/sonarlint/intellij/util/SonarLintUtilsTests.java
```diff
@@ -87,4 +91,17 @@ void testWithTrailingSlash() {
     assertThat(SonarLintUtils.withTrailingSlash(URL_WITHOUT_SLASH)).isEqualTo(URL_WITH_SLASH);
   }
 
+  @Test
+  void testIsDogfoodEnvironment() {
```
nquinquenel, line 95: I would split this into 2 tests. We try to apply the 3A pattern (Arrange/Act/Assert), here you re-arrange the data after asserting, it's not ideal

src/main/java/org/sonarlint/intellij/util/RegionUtils.kt
eray-felek-sonarsource: It is used in a Java class so we do need it

nquinquenel (approved):
LGTM

Let's do the remaining UI/UX changes in a separate ticket. Considerer writing UI tests if possible!

PR #1368, opened 2025-04-07
SLI-1962 Improve Walkthrough Persistence

eray-felek-sonarsource:
[SLI-1962](https://sonarsource.atlassian.net/browse/SLI-1962)

[SLI-1962]: https://sonarsource.atlassian.net/browse/SLI-1962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nquinquenel (commented):
It works but I think we can just leverage `SonarLintGlobalSettings`, where we already store a bunch of properties. I don't think we should store a new file dedicated for this.

src/main/java/org/sonarlint/intellij/OpenWelcomePageOnceOneProjectOpened.kt
```diff
@@ -36,9 +37,15 @@ class OpenWelcomePageOnceOneProjectOpened : StartupActivity {
...
 
         val properties = PropertiesComponent.getInstance()
+        val hasPropertiesWalkthroughRunOnce = properties.getBoolean(HAS_WALKTHROUGH_RUN_ONCE, false)
+        if (hasPropertiesWalkthroughRunOnce) {
+            getGlobalSettings().setHasWalkthroughRunOnce(true)
+            return
+        }
 
-        if (!properties.getBoolean(HAS_WALKTHROUGH_RUN_ONCE, false)) {
+        if (!getGlobalSettings().hasWalkthroughRunOnce()) {
```
nquinquenel, line 46: The 2 conditions should be grouped. In any case we set the values in both the global settings and in the property field

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -193,6 +195,14 @@ public void setNodejsPath(String nodejsPath) {
     this.nodejsPath = nodejsPath;
   }
 
+  public boolean hasWalkthroughRunOnce() {
+    return hasWalkthroughRunOnce;
+  }
+
+  public void setHasWalkthroughRunOnce(boolean hasWalkthroughRunOnce) {
```
nquinquenel, line 202: In `SonarLintGlobalConfigurable` you should also probably set the value in the new settings, otherwise it will get reset to false every time the user opens the settings

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettings.java
```diff
@@ -50,6 +50,8 @@ public final class SonarLintGlobalSettings {
 
   private List<ServerConnection> servers = new LinkedList<>();
   private List<String> fileExclusions = new LinkedList<>();
+  public boolean hasWalkthroughRunOnce = false;
```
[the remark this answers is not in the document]
eray-felek-sonarsource: Outdated code, it's already private

PR #1380, opened 2025-04-23
SLI-1942 Avoid mixing up ID and UUID for taints

nquinquenel:


sophio-japharidze-sonarsource (approved):
FWIW, the taint ID on newer versions of SonarQube Server _will_ be a UUID. But I agree it's still best to default to showing a rule description and "blindly" jumping to location instead of requiring taint to be returned by SLCORE.

PR #1386, opened 2025-04-30
SLI-2000 Remove Deprecated includedRules and excludedRules from Globa…

eray-felek-sonarsource:


nquinquenel (changes requested):
I think this does not compile, some tests are still relying on the fields you removed.

Also, you recently added a copy constructor with unit tests related to it, which you did not update. The number of fields has changed, and so the test should be updated accordingly.

nquinquenel (approved):

PR #1395, opened 2025-05-09
Feature/ef/analyzer updates second round 10.23

eray-felek-sonarsource:


nquinquenel (commented):
I think that would be better to clear the previous dependencies. It's going to be difficult next time to clean up the outdated checksums, here we know that we just have to remove ruby and text

nquinquenel (approved):

PR #1405, opened 2025-05-20
SLI-2025 Raise errors during connection creation wizard

eray-felek-sonarsource:


nquinquenel (approved):
LGTM, this is an improvement, but there's still a problem coming from SLOOP. The error is thrown and not catched at any point, as a result there's an error thrown at the IDE level due to the RPC failure. I will try to do the fix on SLCORE.

PR #1414, opened 2025-06-02
SLI-1531 Update to IntelliJ Platform Gradle Plugin (2.x)

nquinquenel:
[SLI-1531](https://sonarsource.atlassian.net/browse/SLI-1531)

[SLI-1531]: https://sonarsource.atlassian.net/browse/SLI-1531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -44,7 +44,6 @@ description = "SonarLint for IntelliJ IDEA"
 val intellijBuildVersion: String by project
 val omnisharpVersion: String by project
 val runIdeDirectory: String by project
-
 val verifierVersions: String by project
 
 // The environment variables ARTIFACTORY_PRIVATE_USERNAME and ARTIFACTORY_PRIVATE_PASSWORD are used on CI env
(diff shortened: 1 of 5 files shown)


eray-felek-sonarsource:
I am still doing some tests but my first impression is that it takes significantly longer to run runIde

nquinquenel:
> I am still doing some tests but my first impression is that it takes significantly longer to run runIde

Unfortunately, there's a bug preventing us from using a `localPath`. I [reported](https://github.com/JetBrains/intellij-platform-gradle-plugin/issues/1951) the issue and it will be fixed in version 2.6.1. For now, we have to provide dedicated running tasks that will use a specific version, so it will likely redownload the required IDE.

It's also resolving dependencies that involve the Plugin Verifier structure library for parsing and preparing the environment. This might take some time and it was also reported. Although locally it should take some time only the first time.

eray-felek-sonarsource:
> > I am still doing some tests but my first impression is that it takes significantly longer to run runIde
> 
> Unfortunately, there's a bug preventing us from using a `localPath`. I [reported](https://github.com/JetBrains/intellij-platform-gradle-plugin/issues/1951) the issue and it will be fixed in version 2.6.1. For now, we have to provide dedicated running tasks that will use a specific version, so it will likely redownload the required IDE.
> 
> It's also resolving dependencies that involve the Plugin Verifier structure library for parsing and preparing the environment. This might take some time and it was also reported. Although locally it should take some time only the first time.

It only seemed to be downloading the IDE at first but my 55 second duration was based on the runs after it. When it first downloaded it took more than 2 minutes

eray-felek-sonarsource:
We also need to update gitignore, we get changelogs like bundledPlugin-com.intellij.java-IC-223.8214.52.xml

PR #1424, opened 2025-06-10
SLI-2087 Allow sharing connection for modules

nquinquenel:
[SLI-2087](https://sonarsource.atlassian.net/browse/SLI-2087)

### DIFF
--- a/src/main/java/org/sonarlint/intellij/core/BackendService.kt
+++ b/src/main/java/org/sonarlint/intellij/core/BackendService.kt
@@ -674,6 +674,15 @@ class BackendService : Disposable {
         }
     }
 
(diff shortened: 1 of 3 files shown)


eray-felek-sonarsource:
I did some tests and it was not working as I expected

nquinquenel:
> I did some tests and it was not working as I expected

You're right, heuristics were based on the name of your project/folders, which did not work well in some cases. I refactored quite a bit how we proceed modules/projects and made it work in a better manner - it's purely based on the configuration scope IDs now.

eray-felek-sonarsource (commented):
I think having some information about the overridden modules on notification is not enough, user might overlook this and we should improve the UI on our modal binding pop-up to let users know they will have overridden modules when they connect.

eray-felek-sonarsource (approved):
LGTM

PR #1425, opened 2025-06-11
SLI-2054 Implement didOpenFile

eray-felek-sonarsource:
[SLI-2054](https://sonarsource.atlassian.net/browse/SLI-2054)

[SLI-2054]: https://sonarsource.atlassian.net/browse/SLI-2054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -1020,4 +1021,9 @@ class BackendService : Disposable {
         return sloop?.isAlive == true
     }
 
+    fun didOpenFile(configScopeId: String, fileUri: URI) {
```
nquinquenel: I would define two different `didOpenFile` for module/project. We try not to use the projectId/moduleId outside this class

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.java
```diff
@@ -54,13 +56,33 @@ public void onProjectOpened() {
...
+        var uri = VirtualFileUtils.INSTANCE.toURI(file);
+        if (configScopeId != null && uri != null) {
+          getService(BackendService.class).didOpenFile(configScopeId, uri);
+        }
+
+        if (isRider() && module != null) {
+          getService(BackendService.class).updateFileSystem(Map.of(module, List.of(new VirtualFileEvent(ModuleFileEvent.Type.CREATED, file))), true);
+        }
+      });
+    }
```
nquinquenel: I don't think we should be doing this here. This method is called on `StartupActivity`, it's for initializing bus and other services. What's the problem it's trying to solve?

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.java
```diff
@@ -54,13 +56,33 @@ public void onProjectOpened() {
...
   }
 
   @Override
   public void fileOpened(@NotNull FileEditorManager source, @NotNull VirtualFile file) {
     runOnPooledThread(source.getProject(), () -> {
       var module = findModuleForFile(file, source.getProject());
       if (module != null && FileUtils.Companion.isFileValidForSonarLintWithExtensiveChecks(file, source.getProject())) {
+        var uri = VirtualFileUtils.INSTANCE.toURI(file);
```
nquinquenel: Instead You could path the virtual file directly to the backend service and do the conversion there

src/main/java/org/sonarlint/intellij/core/BackendService.kt
nquinquenel: Also, we have a test class for methods in `BackendService`, you can add a small test about this one

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.java
eray-felek-sonarsource: We need to pass the open files information to the backend. Where would you suggest for doing this?

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.java
eray-felek-sonarsource: didOpenFile expects URI, how could we do that?

nquinquenel: The call to SLOOP yes, but I think the method in `BackendService` can take a `VirtualFile` as parameter, otherwise you'll end up doing this operation everywhere you call `BackendService.didOpenFile`

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.java
nquinquenel: I think when analysis readiness goes to green, it also triggers an analysis on open files for example?
Before, we did not have to rely on what you did here - there's a reason probably

eray-felek-sonarsource: Hi, only call I see is didChangeAnalysisReadiness made by SLCORE for such cases in which we trigger an analysis not SLCORE. I am not sure which call are you referring. Considering automatic analysis was not enabled not sure SLCORE was triggering an analysis as you mentioned before.

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.java
```diff
@@ -54,13 +56,33 @@ public void onProjectOpened() {
     myProject.getMessageBus()
       .connect()
       .subscribe(FileEditorManagerListener.FILE_EDITOR_MANAGER, this);
+
+    var fileEditorManager = FileEditorManager.getInstance(myProject);
+    for (VirtualFile file : fileEditorManager.getOpenFiles()) {
```
nquinquenel: You could take the opportunity to rewrite this class in Kotlin while it's small. You could benefit from the custom Extensions we have, such as `Project.getOpenFiles()`

nquinquenel:
```suggestion
    for (var file : fileEditorManager.getOpenFiles()) {
```

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.java
```diff
@@ -54,13 +56,33 @@ public void onProjectOpened() {
...
+      runOnPooledThread(myProject, () -> {
+        var module = findModuleForFile(file, myProject);
+        var configScopeId = module != null ? BackendService.Companion.moduleId(module) : BackendService.Companion.projectId(myProject);
+        var uri = VirtualFileUtils.INSTANCE.toURI(file);
+        if (configScopeId != null && uri != null) {
+          getService(BackendService.class).didOpenFile(configScopeId, uri);
+        }
+
+        if (isRider() && module != null) {
```
nquinquenel: You should do this before `didOpenFile`

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.java
eray-felek-sonarsource: After internal discussion we decided to do it here

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.kt
```diff
@@ -0,0 +1,100 @@
...
+            if (module != null && isFileValidForSonarLintWithExtensiveChecks(
+                    file, source.project
+                )
+            ) {
+
+                SonarLintUtils.getService(BackendService::class.java).didOpenFileForModule(module, file)
+
+                SonarLintUtils.getService(BackendService::class.java).updateFileSystem(
+                    Map.of(
+                        module, List.of(
```
nquinquenel: `listOf()` instead of `List.of()`

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.kt
```diff
@@ -0,0 +1,100 @@
...
+            val module = SonarLintAppUtils.findModuleForFile(file, source.project)
+            if (module != null && isFileValidForSonarLintWithExtensiveChecks(
+                    file, source.project
+                )
+            ) {
+
+                SonarLintUtils.getService(BackendService::class.java).didOpenFileForModule(module, file)
+
+                SonarLintUtils.getService(BackendService::class.java).updateFileSystem(
+                    Map.of(
```
nquinquenel: `mapOf()`

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.kt
```diff
@@ -0,0 +1,100 @@
...
+        EventScheduler(myProject, "open", TriggerType.EDITOR_OPEN, 1000, true)
+
+    fun onProjectOpened() {
+        myProject.messageBus.connect().subscribe(FileEditorManagerListener.FILE_EDITOR_MANAGER, this)
+
+        for (file in myProject.getOpenFiles()) {
+            runOnPooledThread(myProject) {
+                val module = SonarLintAppUtils.findModuleForFile(file, myProject)
+
+                if (SonarLintUtils.isRider() && module != null) {
```
nquinquenel: I think you can put the if condition inside the `if (module != null) {` below, for simplicity

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.kt
```diff
@@ -0,0 +1,100 @@
...
+                    SonarLintUtils.getService(BackendService::class.java).updateFileSystem(
+                        Map.of(
```
nquinquenel: `mapOf(xxx to yyy)`

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.kt
```diff
@@ -0,0 +1,100 @@
...
+import org.sonarlint.intellij.util.SonarLintAppUtils
+import org.sonarlint.intellij.util.getOpenFiles
+import org.sonarlint.intellij.util.runOnPooledThread
+import org.sonarsource.sonarlint.plugin.api.module.file.ModuleFileEvent
+
+@ThreadSafe
+@Service(Service.Level.PROJECT)
+class EditorOpenTrigger(private val myProject: Project) : FileEditorManagerListener, Disposable {
+    private val scheduler =
+        EventScheduler(myProject, "open", TriggerType.EDITOR_OPEN, 1000, true)
```
nquinquenel: You can probably remove the new line

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.kt
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    override fun fileOpened(source: FileEditorManager, file: VirtualFile) {
+        runOnPooledThread(source.project) {
+            val module = SonarLintAppUtils.findModuleForFile(file, source.project)
+            if (module != null && isFileValidForSonarLintWithExtensiveChecks(
+                    file, source.project
+                )
+            ) {
```
nquinquenel: I think you can put the whole if condition on one line

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.kt
```diff
@@ -0,0 +1,100 @@
...
+@Service(Service.Level.PROJECT)
+class EditorOpenTrigger(private val myProject: Project) : FileEditorManagerListener, Disposable {
+    private val scheduler =
+        EventScheduler(myProject, "open", TriggerType.EDITOR_OPEN, 1000, true)
+
+    fun onProjectOpened() {
+        myProject.messageBus.connect().subscribe(FileEditorManagerListener.FILE_EDITOR_MANAGER, this)
+
+        for (file in myProject.getOpenFiles()) {
+            runOnPooledThread(myProject) {
```
nquinquenel: I think the for can be included inside the `runOnPooledThread`

src/main/java/org/sonarlint/intellij/trigger/EditorOpenTrigger.kt
```diff
@@ -0,0 +1,100 @@
...
+        runOnPooledThread(source.project) {
+            val module = SonarLintAppUtils.findModuleForFile(file, source.project)
+            if (module != null && isFileValidForSonarLintWithExtensiveChecks(
+                    file, source.project
+                )
+            ) {
+
+                SonarLintUtils.getService(BackendService::class.java).didOpenFileForModule(module, file)
+
+                SonarLintUtils.getService(BackendService::class.java).updateFileSystem(
```
nquinquenel: `updateFileSystem` should be before `didOpenFileForModule`

nquinquenel (approved):
LGTM! Just some comments about clean code improvements

PR #1443, opened 2025-06-27
SLI-1568 Adopt SLCORE changes for triggering analysis

nquinquenel:


src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.kt
```diff
@@ -0,0 +1,191 @@
...
+                    response.analysisId?.let { analysisId ->
+                        val analysisState = AnalysisState(analysisId, callback, module)
+                        getService(project, RunningAnalysesTracker::class.java).track(analysisState)
+                    }
+                }
+            }
+        }
+    }
+
+    fun autoAnalyzeFiles(files: List<VirtualFile>) {
```
eray-felek-sonarsource: This seems to be used only in tests, can't we remove it?

src/main/java/org/sonarlint/intellij/analysis/AnalysisScope.java
```diff
@@ -1,90 +0,0 @@
-/*
- * SonarLint for IntelliJ IDEA
```
eray-felek-sonarsource, line 2: For full project analysis we were lettings users how many files and projects being analyzed in total in the progress bar. With the removal of this class, and the new adaptation, we are now only showing module by module. I think this would not help managing our user's expectation on how long they will wait for analysis to be finished.

nquinquenel: I added a message in the progress status bar that shows the number of modules being analyzed. Unfortunately, we don't have the information about the number of files anymore (this is part of SLCORE logic).

src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.kt
nquinquenel: Ideally yes, we should not rely on code made for test. I removed this method and changed the logic of the test to rely on didOpenFile

eray-felek-sonarsource (approved):
LGTM

PR #1444, opened 2025-06-27
SLI-2105 Add utm parameters to promotion links.

Krosovok:
[SLI-2105](https://sonarsource.atlassian.net/browse/SLI-2105)

[SLI-2105]: https://sonarsource.atlassian.net/browse/SLI-2105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Krosovok:
Some part of the functionality still missing because of the dependence on SLCore https://sonarsource.atlassian.net/browse/SLCORE-1504

common/build.gradle.kts
```diff
@@ -13,3 +13,13 @@ intellij {
         version.set(intellijBuildVersion)
     }
 }
+
+dependencies {
+    testImplementation(platform(libs.junit.bom))
+    testImplementation("org.junit.jupiter:junit-jupiter")
```
damien-urruty-sonarsource, line 39: For consistency, you could also declare the lib in `gradle/libs.versions.toml`

src/test/java/org/sonarlint/intellij/SonarLintIntelliJClientTests.kt
```diff
@@ -256,6 +258,15 @@ class SonarLintIntelliJClientTests : AbstractSonarLintLightTests() {
         )
     }
 
+    @Test
+    fun should_promote_languages_with_utm_params() {
+        client.promoteExtraEnabledLanguagesInConnectedMode(projectBackendId, setOf(Language.ANSIBLE, Language.SCALA))
+        val action = projectNotifications[0].actions[0]
+
+        assertThat(action).isInstanceOf(OpenTrackedLinkAction::class.java)
+        org.assertj.core.api.Assertions.assertThat((action as OpenTrackedLinkAction).promotion).isEqualTo(Promotion.ENABLE_LANGUAGE_ANALYSIS)
```
damien-urruty-sonarsource: Usually, we rely on static imports and use `assertThat` directly, it makes the assertion a little less verbose.
EDIT: ah I see that you did it the line just before

src/test/java/org/sonarlint/intellij/notifications/SonarLintProjectNotificationsTests.kt
```diff
@@ -82,14 +84,17 @@ class SonarLintProjectNotificationsTests : AbstractSonarLintLightTests() {
...
+        val action = notifications[0].actions[0]
 
         Assertions.assertThat(notifications).hasSize(1)
         Assertions.assertThat(notifications[0].title).isEqualTo("<b>SonarQube for IDE suggestions</b>")
         Assertions.assertThat(notifications[0].content).isEqualTo(content)
         Assertions.assertThat(notifications[0].type).isEqualTo(NotificationType.INFORMATION)
         Assertions.assertThat(notifications[0].icon).isEqualTo(SonarLintIcons.SONARQUBE_FOR_INTELLIJ)
         Assertions.assertThat(notifications[0].actions.size).isEqualTo(4)
+        Assertions.assertThat(action).isInstanceOf(OpenTrackedLinkAction::class.java)
+        Assertions.assertThat((action as OpenTrackedLinkAction).promotion).isEqualTo(Promotion.ENABLE_LANGUAGE_ANALYSIS)
```
damien-urruty-sonarsource: Same here, it was not the case before, but you could use static imports

src/main/java/org/sonarlint/intellij/telemetry/LinkTelemetry.kt
```diff
@@ -52,8 +54,21 @@ enum class LinkTelemetry(
...
+
+        BrowserUtil.browse(withParameters(promotion))
+    }
+
+    private fun withParameters(promotion: Promotion?): String {
+        var promotionParams = emptyMap<String, String>()
+        if (promotion != null) {
+            promotionParams = promotion.trackingParams
+        }
+        return UrlUtils.addParameters(url, promotionParams)
```
damien-urruty-sonarsource: you could simplify this with `UrlUtils.addParameters(url, promotion?.trackingParams ?: mapOf())`. And/or allow null in `UrlUtils.addParameters` and move the check there

src/main/java/org/sonarlint/intellij/promotion/Promotion.kt
```diff
@@ -0,0 +1,57 @@
...
+package org.sonarlint.intellij.promotion
+
+private val BASE_PARAMETERS = mapOf(
+    "utm_medium" to "referral",
+    "utm_source" to "sq-ide-product-intellij"
+)
+private val NOTIFICATION_PARAMETERS = BASE_PARAMETERS +
+    ("utm_content" to "notification")
+
+enum class Promotion(val trackingParams: Map<String, String>) {
```
damien-urruty-sonarsource: The name promotion is maybe a bit too broad, in this case what about `Utm`?

common/src/main/java/org/sonarlint/intellij/common/util/UrlUtils.kt
```diff
@@ -19,6 +19,7 @@
  */
 package org.sonarlint.intellij.common.util
 
+import io.ktor.http.URLBuilder
```
damien-urruty-sonarsource, line 22: Is ktor bundled with IntelliJ? I was a bit surprised but I found another occurrence of us using this library. I was about to advise not using it, it looks a bit internal or transitive, but maybe it's fine

src/main/java/org/sonarlint/intellij/promotion/Promotion.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    NEW_CONNECTION_PANEL(
+        BASE_PARAMETERS + mapOf(
+            "utm_content" to "create-new-connection-panel",
+            "utm_term" to "explore-sonarqube-cloud-free-tier"
```
damien-urruty-sonarsource: Are you using the Kotlin code style as described in our [toolset](https://github.com/SonarSource/sonar-developer-toolset?tab=readme-ov-file#code-style-configuration-for-intellij)? I suspect there is a missing trailing comma here.

common/src/main/java/org/sonarlint/intellij/common/util/UrlUtils.kt
nquinquenel: Good catch, ktor is bundled with IntelliJ Community but I don't think we need it. The other occurence is from `VirtualFileUtils#getEncoding`, `virtualFile.charset.name` should be replaced by `virtualFile.charset.name()` to remove the usage of ktor. Perhaps it's not in the scope of this ticket, this can be handled later

src/main/java/org/sonarlint/intellij/promotion/Promotion.kt
Krosovok: I have no idea what UTM actually stands for and I assume a random person from the future might also not know, so I wanted something easier to understand at the first glance. How about `PromotionType` or `PromotionLink` or something along these lines?

src/main/java/org/sonarlint/intellij/promotion/Promotion.kt
```diff
@@ -0,0 +1,57 @@
...
+            ("utm_term" to "analyze-project-ci-cd-pipeline-downloads")
+    ),
+    DETECT_SECURITY_ISSUES(
+        NOTIFICATION_PARAMETERS +
+            ("utm_term" to "enable-project-analysis-signup-free")
+    ),
+    ENABLE_LANGUAGE_ANALYSIS(
+        NOTIFICATION_PARAMETERS +
+            ("utm_term" to "enable-project-analysis-signup-free")
+    );
```
nquinquenel: These 2 values are identical, is it expected? I think the `DETECT_SECURITY_ISSUES` should maybe be `detect-security-issues-files-signup-free` or `detect-security-issues-files-downloads`, based on the ticket description

common/src/main/java/org/sonarlint/intellij/common/util/UrlUtils.kt
```diff
@@ -28,5 +29,23 @@ class UrlUtils {
...
+         */
+        fun addParameters(url: String, params: Map<String, String>): String {
+            if (params.isEmpty()) {
+                return url
+            }
+
+            val urlBuilder = URLBuilder(url)
+            for (entry in params) {
+                urlBuilder.parameters.append(entry.key, entry.value)
+            }
```
nquinquenel: Another way of writing this in kotlin style (as you prefer):
`params.map { (key, value) -> urlBuilder.parameters.append(key, value) }`

common/src/test/java/org/sonarlint/intellij/common/util/UrlUtilsTest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+import org.assertj.core.api.Assertions
+import org.junit.jupiter.api.Test
+
+private const val BASE_URL = "http://localhost:8080/url"
+
+class UrlUtilsTest {
+
+    @Test
+    fun should_not_change_url_when_no_params() {
```
nquinquenel: Just for sharing, you can also name the kotlin test this way:
```fun `should not change url when no params`() {```
Ideally we should be consistent in the naming, but I think in some tests we started writing it like this

File: src/main/java/org/sonarlint/intellij/promotion/PromotionProvider.kt [hunk 11]
```diff
@@ -138,23 +138,23 @@ class PromotionProvider(private val project: Project) {
...
-        showPromotion(notifications, "Analyze your project in your CI/CD pipeline")
+        showPromotion(notifications, "Analyze your project in your CI/CD pipeline", Promotion.ANALYZE_CI_CD)
     }
 
     private fun processAdvancedLanguagePromotion(notifications: SonarLintProjectNotifications, extension: String) {
         val language = findLanguage(extension, languagesHavingAdvancedRules)
 
         if (language != null) {
-            showPromotion(notifications, "Detect more security issues in your " + language.label + " files")
+            showPromotion(notifications, "Detect more security issues in your " + language.label + " files", Promotion.DETECT_SECURITY_ISSUES)
```
nquinquenel:
```suggestion
            showPromotion(notifications, "Detect more security issues in your ${language.label} files", Promotion.DETECT_SECURITY_ISSUES)
```

nquinquenel (commented):
Looking great!

common/src/main/java/org/sonarlint/intellij/common/util/UrlUtils.kt
damien-urruty-sonarsource: If it can help, we built our own [UrlBuilder](https://github.com/SonarSource/sonarqube-mcp-server/blob/master/src/main/java/org/sonarsource/sonarqube/mcp/serverapi/UrlBuilder.java) in MCP

src/main/java/org/sonarlint/intellij/promotion/Promotion.kt
damien-urruty-sonarsource: I see your point, and having easy to understand code is definitely a valid concern. In this case, I would suggest you to read a bit about [UTM](https://en.wikipedia.org/wiki/UTM_parameters) then, it's some kind of "standard", so I don't think it's a problem referencing it in the code, JB also did it in [SLCORE](https://github.com/SonarSource/sonarlint-core/pull/1392/files#diff-07f99c4973973a5856ac152e1cb3cadeb58bb8be97e5e9eba7bc3fff530592c2R46) for the same MMF. Maybe `UtmParameters`?

Krosovok: Renamed to `UtmParameters` and added a Javadoc with a little explanation.

common/build.gradle.kts
Krosovok: For some reason it seems to not be working both in root project and in commons. So I am leaving it as it is.

common/src/main/java/org/sonarlint/intellij/common/util/UrlUtils.kt
nquinquenel: My bad, it should be `params.forEach` and not `params.map`, map returns the list of transformed elements whereas `forEach` only applies something on each element

common/build.gradle.kts
```diff
@@ -1,3 +1,23 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
```
nquinquenel, line 19: We don't have HEADER on build files, was it added manually or with the license plugin?

common/src/test/java/org/sonarlint/intellij/common/util/UrlUtilsTest.kt
```diff
@@ -0,0 +1,78 @@
...
+    @Test
+    fun `should append to existing parameters`() {
+        val withParameters = UrlUtils.addParameters(
+            "$BASE_URL?existingParam=existingValue", mapOf(
+                "existingParam" to "value1",
+                "param2" to "value2"
+            )
+        )
+
+        Assertions.assertThat(withParameters).isEqualTo("$BASE_URL?existingParam=existingValue&existingParam=value1&param2=value2")
```
nquinquenel: We usually use static import to reduce noise, as `Assertions` is repeated quite often

src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -159,9 +144,27 @@ import org.sonarsource.sonarlint.core.rpc.protocol.common.SonarCloudRegion
...
+import java.time.format.DateTimeParseException
+import java.util.UUID
+import java.util.concurrent.CancellationException
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.logging.Filter
+import java.util.logging.Level
+import java.util.logging.Logger
```
nquinquenel: Your Kotlin code style seems to be different, I think we try to keep it in alphabetical order, here is my import setting:
image
Maybe this is not standardized in the developer-toolset repo for kotlin

build.gradle.kts
```diff
@@ -1,5 +1,29 @@
...
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
 
 import com.jetbrains.plugin.blockmap.core.BlockMap
+import org.apache.commons.compress.archivers.tar.TarArchiveEntry
+import org.apache.commons.compress.archivers.tar.TarArchiveInputStream
+import org.apache.commons.compress.compressors.gzip.GzipCompressorInputStream
+import org.jetbrains.intellij.tasks.RunPluginVerifierTask
```
damien-urruty-sonarsource: hmm it's strange that these imports were reordered, maybe a problem with code style. Have you followed all the steps described in the [toolset](https://github.com/SonarSource/sonar-developer-toolset)?

common/src/test/java/org/sonarlint/intellij/common/util/UrlUtilsTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+private const val BASE_URL = "http://localhost:8080/url"
+
+class UrlUtilsTest {
+
+    @Test
+    fun `should not change url when null map`() {
+        val same = UrlUtils.addParameters(BASE_URL, null)
+
+        Assertions.assertThat(same).isSameAs(BASE_URL)
```
damien-urruty-sonarsource: It's fine like this but FYI you could use static imports (in Java) or import the method (in Kotlin) to make it shorter to read:

```
import org.assertj.core.api.Assertions.assertThat

assertThat(same).isSameAs(BASE_URL)
```

You will find it quite often in the codebase

src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -159,9 +144,27 @@ import org.sonarsource.sonarlint.core.rpc.protocol.common.SonarCloudRegion
 import org.sonarsource.sonarlint.core.rpc.protocol.common.TokenDto
 import org.sonarsource.sonarlint.core.rpc.protocol.common.UsernamePasswordDto
 import org.sonarsource.sonarlint.plugin.api.module.file.ModuleFileEvent
+import java.io.IOException
```
damien-urruty-sonarsource: Same problem with imports here

src/main/java/org/sonarlint/intellij/promotion/UtmParameters.kt
```diff
@@ -0,0 +1,81 @@
...
+private val CONNECTION_PANEL_PARAMETERS = BASE_PARAMETERS +
+    (CONTENT to "create-new-connection-panel")
+private val NOTIFICATION_PARAMETERS = BASE_PARAMETERS +
+    (CONTENT to "notification")
+
+/**
+ * Sets of parameters used by Google Analytics to track where the clicks are coming from.
+ * Each enum value corresponds to a place where a link can be clicked so that it has its own set of passed parameters.
+ */
+enum class UtmParameters(val trackingParams: Map<String, String>) {
```
nquinquenel: I think we could improve `trackingParams`. Instead of a `Map<String, String>` that does not give much information about the content, we could introduce a `UtmKey` enum object as the key of the Map. We actually expect the key to be one of the four const values declared above.

And we could even go one step further, looking at the code, it seems that we always expect to have all of the 4 UTM keys to be filled in. I think we could introduce a new data class object with the four UTM attributes, which would replace this Map, and simplify the `toDto`.

src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -159,9 +144,27 @@ import org.sonarsource.sonarlint.core.rpc.protocol.common.SonarCloudRegion
...
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.logging.Filter
+import java.util.logging.Level
+import java.util.logging.Logger
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.issue.CheckStatusChangePermittedParams as issueCheckStatusChangePermittedParams
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.issue.CheckStatusChangePermittedResponse as issueCheckStatusChangePermittedResponse
 
+private val UTM = UtmParameters.CREATE_SQC_TOKEN.toDto()
```
nquinquenel: I think we can avoid this top-level declaration. This value is only used in a single method, and creating the `toDto` is not costly - we also don't call `helpGenerateUserToken` very often. I think having it as a top-level declaration creates noise for limited value. This class is already quite big and the UTM attribute does not relate to it

src/test/java/org/sonarlint/intellij/promotion/UtmParametersTest.kt
```diff
@@ -0,0 +1,44 @@
...
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+package org.sonarlint.intellij.promotion
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class UtmParametersTest {
```
nquinquenel, line 26: We name our test classes `***Tests` and not `Test` 😄

nquinquenel (commented):
Logic is great, although I think we can still improve the design and make it cleaner!

src/main/java/org/sonarlint/intellij/promotion/UtmParameters.kt
Krosovok: I see. The thing is that in all other places we do need the map itself but we can make another method like `toMap`/`asMap`.

Krosovok: I tried to think how to make it work and it kinda feels like creating more noise. With map it's easy to move common parts to `BASE_PARAMETERS` and others and with fields we would have to copy paste half of the params.

Introducing enum totally makes sense until I come back to the fact that most usages would require us to move it to `Map<String, String>` anyway. So it feels like adding unnecessary steps. I would totally agree if there was some usage of these fields in the code but here I would prefer them as it is now.

common/build.gradle.kts
Krosovok: Probably idea recognizes it as kotlin file and I didn't notice it adding it. :sweat_smile:

src/main/java/org/sonarlint/intellij/promotion/UtmParameters.kt
nquinquenel: What do you think of the following proposal?

```
private data class UtmParamSet(
    val medium: String = "referral",
    val source: String = "sq-ide-product-intellij",
    val content: String,
    val term: String
) {
    fun toMap() = mapOf(
        "utm_medium" to medium,
        "utm_source" to source,
        "utm_content" to content,
        "utm_term" to term
    )
}

enum class UtmParameters(
    private val content: String,
    private val term: String
) {
    NEW_CONNECTION_PANEL("create-new-connection-panel", "explore-sonarqube-cloud-free-tier"),
    DETECT_PROJECT_ISSUES("notification", "detect-project-issues-signup-free"),
    SPEED_UP_ANALYSIS("notification", "speed-up-project-analysis-signup-free"),
    ANALYZE_CI_CD("notification", "analyze-project-ci-cd-pipeline-downloads"),
    DETECT_SECURITY_ISSUES("notification", "detect-security-issues-files-signup-free"),
    ENABLE_LANGUAGE_ANALYSIS("notification", "enable-project-analysis-signup-free"),
    CREATE_SQC_TOKEN("create-new-connection-panel", "create-sqc-token");

    private val params = UtmParamSet(content = content, term = term)
    val trackingParams get() = params.toMap()

    fun toDto(): HelpGenerateUserTokenParams.Utm = HelpGenerateUserTokenParams.Utm(
        params.medium, params.source, params.content, params.term
    )
}

```

This allows for a strongly-typed solution, enforcing that the right values are present

common/src/test/java/org/sonarlint/intellij/common/util/UrlUtilsTests.kt
```diff
@@ -0,0 +1,78 @@
...
+    fun `should not change url when no params`() {
+        val same = UrlUtils.addParameters(BASE_URL, emptyMap())
+
+        assertThat(same).isSameAs(BASE_URL)
+    }
+
+    @Test
+    fun `should add params where there are none`() {
+        val withParameters = UrlUtils.addParameters(
+            "$BASE_URL?", mapOf(
```
nquinquenel:
```suggestion
            BASE_URL, mapOf(
```
The question mark makes it look like it requires the URL to have a "?" if there are parameters, I think it's better to remove this in the test to really assert that everything is handled correctly

nquinquenel (approved):
LGTM!

One last comment, and otherwise there is this discussion about ktor library that is used, I think we could remove it. It's ok not to do it in this ticket, but then we could maybe create a small task ticket about this for a later hardening.

Krosovok:
> LGTM!
> 
> One last comment, and otherwise there is this discussion about ktor library that is used, I think we could remove it. It's ok not to do it in this ticket, but then we could maybe create a small task ticket about this for a later hardening.

Who should create a ticket?

PR #1449, opened 2025-07-01
SLI-2123 Run analysis in a cancellable task

nquinquenel:


src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.kt
```diff
@@ -96,17 +106,20 @@ class AnalysisSubmitter(private val project: Project) {
...
+            val filesByModule = files.groupBy { selectedFile ->
                 findModuleForFile(selectedFile, project)
-            }.forEach { (module, files) ->
+            }
+            val taskState = createGlobalTaskIfNeeded("Analyzing Files", filesByModule.size)
+            filesByModule.forEach { (module, files) ->
                 module?.let {
                     getService(project, PromotionProvider::class.java).handlePromotionOnAnalysis()
-                    getService(BackendService::class.java).analyzeFileList(module, files).thenAcceptAsync { response ->
+                    getService(BackendService::class.java).analyzeFileList(module, files).thenAccept { response ->
```
eray-felek-sonarsource, line 116: Why did we have to change to thenAccept? Since this is a long operation and also we are going to the backend here, I see a potential risk of freeze. Are we sure it would not happen?

nquinquenel: We are already in a background thread at this step, there's no risk of blocking the UI thread. The problem having an async operation here is that it makes the tracking a difficult operation, because thenAcceptAsync proceeds response in a new thread (for loop might finish before all responses are handled)

src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.kt
```diff
@@ -176,6 +192,16 @@ class AnalysisSubmitter(private val project: Project) {
         }
     }
 
+    private fun createGlobalTaskIfNeeded(title: String, moduleSize: Int): GlobalTaskProgressReporter? {
+        return if (moduleSize > 1) {
+            val task = GlobalTaskProgressReporter(project, title, moduleSize)
+            task.updateText("SonarQube: Analysis 1 Out Of $moduleSize")
```
eray-felek-sonarsource: we could add "modules" at the end : 
Out of $moduleSize modules*

PR #1452, opened 2025-07-03
SLI-2117 Show feedback when 'Search in list...' fails

eray-felek-sonarsource:


src/main/java/org/sonarlint/intellij/tasks/ServerDownloadProjectTask.java
```diff
@@ -45,20 +46,54 @@ public ServerDownloadProjectTask(Project project, ServerConnection server) {
...
       return computeOnPooledThreadWithoutCatching(project, "Download Projects Task",
-        () ->
-        waitForFuture(indicator, getService(BackendService.class)
-          .getAllProjects(server))
-          .getSonarProjects()
-          .stream().collect(Collectors.toMap(SonarProjectDto::getKey, p -> p)));
+        () -> {
+          try {
+            var result = waitForFuture(indicator, getService(BackendService.class).getAllProjects(server));
+            var projects = result.getSonarProjects()
```
nquinquenel: I think you can chain the calls and avoid this `projects` variable

src/main/java/org/sonarlint/intellij/tasks/ServerDownloadProjectTask.java
```diff
@@ -45,20 +46,54 @@ public ServerDownloadProjectTask(Project project, ServerConnection server) {
...
+          }
+        });
     } catch (Exception e) {
       SonarLintConsole.get(project).error("Failed to download list of projects", e);
-      throw e;
+      return DownloadResult.failure("Failed to download list of projects. Reason: " + e.getMessage());
     }
   }
 
+  public static class DownloadResult {
```
nquinquenel, line 68: You almost modified the whole class, I think it could be an opportunity to rewrite the whole class to kotlin if you feel like it, maybe next time

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
```diff
@@ -157,11 +157,16 @@ private void onConnectionSelected() {
...
-    ApplicationManager.getApplication().assertIsDispatchThread();
-
     var downloadTask = new ServerDownloadProjectTask(project, selectedConnection);
     try {
-      return ProgressManager.getInstance().run(downloadTask);
+      var result = ProgressManager.getInstance().run(downloadTask);
+      if (result.isSuccess()) {
+        return result.getProjects();
+      } else {
+        getService(project, SonarLintConsole.class).error(result.getErrorMessage());
```
nquinquenel: I think you are already logging the error in some cases during the task. Maybe remove this one, or completely remove the other logging to avoid duplicates.

src/main/java/org/sonarlint/intellij/tasks/ServerDownloadProjectTask.java
```diff
@@ -45,20 +46,54 @@ public ServerDownloadProjectTask(Project project, ServerConnection server) {
...
-          .getAllProjects(server))
-          .getSonarProjects()
-          .stream().collect(Collectors.toMap(SonarProjectDto::getKey, p -> p)));
+        () -> {
+          try {
+            var result = waitForFuture(indicator, getService(BackendService.class).getAllProjects(server));
+            var projects = result.getSonarProjects()
+              .stream().collect(Collectors.toMap(SonarProjectDto::getKey, p -> p));
+            return DownloadResult.success(projects);
+          } catch (Exception e) {
```
nquinquenel, line 58: Somehow it looks like many try/catch blocks. I count 2 of them in this method, and another one in the method calling this task. I don't know if it could be reduced.

nquinquenel (approved):
LGTM. Some minor comments.
Can you also clean up the verification metadata file as well please?

eray-felek-sonarsource:
> LGTM. Some minor comments. Can you also clean up the verification metadata file as well please?

I think we can clean up the verification data when we update to the final released SLCORE before we release otherwise it would be duplicated effort

src/main/java/org/sonarlint/intellij/tasks/ServerDownloadProjectTask.java
eray-felek-sonarsource: Next time :)

src/main/java/org/sonarlint/intellij/tasks/ServerDownloadProjectTask.java
eray-felek-sonarsource: You are right, I removed one of the try catch, the other ones are needed since different levels of exceptions can happen on computeOnPooledThread and waitForFuture, if we just keep the general exception on computeOnPooledThread level I noticed we are not catching the waitForFuture level one with the catch block

src/main/java/org/sonarlint/intellij/config/project/SonarLintProjectBindPanel.java
eray-felek-sonarsource: Automatically removed one log since I removed the one try catch block that was not needed

nquinquenel:
> > LGTM. Some minor comments. Can you also clean up the verification metadata file as well please?
> 
> I think we can clean up the verification data when we update to the final released SLCORE before we release otherwise it would be duplicated effort

I think it's better to take the time to do it everytime you update a dependency. Otherwise, how do you know which checksum are outdated and which are not?

PR #1455, opened 2025-07-04
SLI-2080 Cache branch matching information

eray-felek-sonarsource:
[SLI-2080](https://sonarsource.atlassian.net/browse/SLI-2080)

[SLI-2080]: https://sonarsource.atlassian.net/browse/SLI-2080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/git/src/main/kotlin/org/sonarlint/intellij/git/GitRepo.kt
+++ b/git/src/main/kotlin/org/sonarlint/intellij/git/GitRepo.kt
@@ -42,7 +42,7 @@ private var cache: Pair<CacheKey, String?>? = null
 class GitRepo(private val repo: GitRepository, private val project: Project) : VcsRepo {
     override fun electBestMatchingServerBranchForCurrentHead(mainBranchName: String, allBranchNames: Set<String>): String? {
         val currentBranch = repo.currentBranchName
-        val head = repo.currentRevision ?: return null // Could be the case if no commit has been made in the repo
+        val head = repo.currentRevision
         val repoBaseDir = getGitDir()
 
         val cacheKey = CacheKey(currentBranch, head, repoBaseDir, mainBranchName, allBranchNames)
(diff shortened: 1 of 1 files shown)


git/src/main/kotlin/org/sonarlint/intellij/git/GitRepo.kt
```diff
@@ -29,14 +29,34 @@ import java.nio.file.Path
...
+)
+
+private var cache: Pair<CacheKey, String?>? = null
+
 class GitRepo(private val repo: GitRepository, private val project: Project) : VcsRepo {
     override fun electBestMatchingServerBranchForCurrentHead(mainBranchName: String, allBranchNames: Set<String>): String? {
-        return try {
-            val currentBranch = repo.currentBranchName
+        val currentBranch = repo.currentBranchName
+        val head = repo.currentRevision ?: return null // Could be the case if no commit has been made in the repo
```
nquinquenel: You moved this before the check `if (currentBranch != null && currentBranch in allBranchNames)`, I think the behavior is not the same anymore?

git/src/main/kotlin/org/sonarlint/intellij/git/GitRepo.kt
```diff
@@ -29,14 +29,34 @@ import java.nio.file.Path
...
+
+        val cacheKey = CacheKey(currentBranch, head, repoBaseDir, mainBranchName, allBranchNames)
+
+        cache?.let { (cachedKey, cachedResult) ->
+            if (cachedKey == cacheKey) {
+                return cachedResult
+            }
+        }
+
+        val result = try {
```
nquinquenel: The return does not exit the method, the computation in the try is always executed and the cached info is not used

nquinquenel (changes requested):
I think the caching does not work here

git/src/main/kotlin/org/sonarlint/intellij/git/GitRepo.kt
eray-felek-sonarsource: You are right in case of if (currentBranch != null && currentBranch in allBranchNames) it would be the case. I changed it

git/src/main/kotlin/org/sonarlint/intellij/git/GitRepo.kt
```diff
@@ -29,15 +29,41 @@ import java.nio.file.Path
...
+        val repoBaseDir = getGitDir()
+
+        val cacheKey = CacheKey(currentBranch, head, repoBaseDir, mainBranchName, allBranchNames)
+
+        cache?.let { (cachedKey, cachedResult) ->
+            if (cachedKey == cacheKey) {
+                return cachedResult
             }
-            val head = repo.currentRevision ?: return null // Could be the case if no commit has been made in the repo
+        }
```
nquinquenel, line 54: This part is still useless, the value is returned to the void

nquinquenel (changes requested):
I think the logic is still not working. Adding a test could help verify this.

nquinquenel (approved):

eray-felek-sonarsource:
> Would be nice to add some test

Yes, I tested manually but currently we have no automated tests for GitRepo so I created [this](https://sonarsource.atlassian.net/browse/SLI-2136) ticket for it which should include testing the caching mechanism

PR #1457, opened 2025-07-07
Update to released version of SLCORE and update change notes

nquinquenel:


src/main/resources/META-INF/plugin.xml
```diff
@@ -65,7 +65,7 @@
 
     <change-notes><![CDATA[
       <ul>
-        <li>10.27 - 4 new rules and 1 new QuickFix for Java.</li>
+        <li>10.27 - Fix an issue where analysis would never finish. Internal refactoring of the analysis workflow. 4 new rules and 1 new QuickFix for Java.</li>
```
eray-felek-sonarsource, line 68: Due to internal refactoring we have 2 features changed (Power Save Mode and Analysis being triggered only on current file). Maybe we should also mention this?

nquinquenel: I think the refactoring implies many changes, I'd rather not list them all. In the end, it should work similarly in terms of functionality

PR #1468, opened 2025-07-22
SLI-2136 GitRepo tests

Krosovok:
[SLI-2136](https://sonarsource.atlassian.net/browse/SLI-2136)

[SLI-2136]: https://sonarsource.atlassian.net/browse/SLI-2136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/git/src/test/kotlin/org/sonarlint/intellij/git/GitRepoTests.kt
+++ b/git/src/test/kotlin/org/sonarlint/intellij/git/GitRepoTests.kt
@@ -81,7 +81,7 @@ class GitRepoTests : AbstractLightTests() {
     }
 
     @Test
-    fun `should return current branch if it's in allBranchNames`() {
+    fun `should return current branch if it's known`() {
         val gitRepository = initRepoWithCommit()
         gitRepository.update()
         val tested = GitRepo(gitRepository, this.project)
@@ -92,7 +92,7 @@ class GitRepoTests : AbstractLightTests() {
     }
 
     @Test
-    fun `should return current non-main branch if it's in allBranchNames`() {
+    fun `should return current non-main branch if it's known`() {
         val gitRepository = initRepoWithCommit()
         newBranch(EXPECTED)
         gitRepository.update()


Krosovok:
@eray-felek-sonarsource , @damien-urruty-sonarsource please take a look too!

I wrote some tests that allow to verify behavior of `GitRepo` and fixed a little bug in branch selection in the process. It seems that `git rev-list` command's arguments are not symmetric and the command always returned `0` before I swapped parameters. Now it actually calculates distances.

damien-urruty-sonarsource:
@Krosovok I'm leaving for 2 weeks and there are other things I'd like to finish, sorry

git/src/test/kotlin/org/sonarlint/intellij/git/GitRepoTests.kt
```diff
@@ -0,0 +1,301 @@
...
+        gitRepository.update()
+        val tested = GitRepo(gitRepository, this.project)
+
+        val result = tested.electBestMatchingServerBranchForCurrentHead(MAIN_BRANCH, setOf())
+
+        assertThat(result).isNull()
+    }
+
+    @Test
+    fun `should return current branch if it's in allBranchNames`() {
```
nquinquenel: The name of the test is not very clear, `allBranchNames` is an implementation detail in the class, the name of the class should be easy to understand

git/src/test/kotlin/org/sonarlint/intellij/git/GitRepoTests.kt
```diff
@@ -0,0 +1,301 @@
...
+        // replace to avoid flaky test behaviour, otherwise it can mess up verify counts.
+        project.replaceService(GitBranchIncomingOutgoingManager::class.java,
+            mock(GitBranchIncomingOutgoingManager::class.java),
+            testRootDisposable)
+    }
+
+    @AfterEach
+    override fun tearDown() {
+        vcsManager.unregisterVcs(GitVcs.getInstance(project))
+        Files.list(root).forEach { child: Path -> child.toFile().deleteRecursively() }
```
nquinquenel: We tend not to put the types, unless it helps

git/src/test/kotlin/org/sonarlint/intellij/git/GitRepoTests.kt
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    @Test
+    fun `should return null for empty repo`() {
+        val gitRepository = initRepo()
+        gitRepository.update()
+        val tested = GitRepo(gitRepository, this.project)
```
nquinquenel: Minor details but we usually don't put `this.` unless required

git/src/test/kotlin/org/sonarlint/intellij/git/GitRepoTests.kt
```diff
@@ -0,0 +1,301 @@
...
+        verify(repoSpy, times(2)).currentRevision
+        // invoked when no cache
+        verify(repoSpy, times(2)).branches
+    }
+
+    private fun newGitRepository(): GitRepository {
+        vcsManager.setDirectoryMapping(root.toString(), GitVcs.NAME)
+        val file = LocalFileSystem.getInstance().refreshAndFindFileByNioFile(root)
+        val repository = GitUtil.getRepositoryManager(project).getRepositoryForRoot(file)!!
+        return repository
```
nquinquenel: You can directly return without assigning to `repository`

git/src/main/kotlin/org/sonarlint/intellij/git/GitRepo.kt
```diff
@@ -95,8 +96,8 @@ class GitRepo(private val repo: GitRepository, private val project: Project) : V
 
     private fun distance(project: Project, repository: GitRepository, from: String, to: String): Int? {
         val mergeBase = GitHistoryUtils.getMergeBase(project, repository.root, from, to) ?: return null
-        val aheadCount = getNumberOfCommitsBetween(repository, from, mergeBase.asString()) ?: return null
-        val behindCount = getNumberOfCommitsBetween(repository, to, mergeBase.asString()) ?: return null
+        val aheadCount = getNumberOfCommitsBetween(repository, mergeBase.asString(), from) ?: return null
+        val behindCount = getNumberOfCommitsBetween(repository, mergeBase.asString(), to) ?: return null
```
nquinquenel, line 100: I think the name of the parameters in `getNumberOfCommitsBetween` does not really fit. You are passing `from` in the `to` parameter for example, could you make it clearer?

nquinquenel (approved):
Looking good! Just a few minor comments for clarity

git/src/test/kotlin/org/sonarlint/intellij/git/GitRepoTests.kt
Krosovok: How else would you name it? The thing that the test checks is basically "...present in list that we pass as a parameter". I can call it something else than it's name but how that would be clearer? I can call call them like `remote branches list` or `server branches` but that would be even more implementation details (and on SLCore side no less).

nquinquenel: `should return current branch if it's known` something like that?

PR #1478, opened 2025-07-31
SLI-2162 display dependency risks

nquinquenel:
[SLI-2162](https://sonarsource.atlassian.net/browse/SLI-2162)

[SLI-2162]: https://sonarsource.atlassian.net/browse/SLI-2162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Krosovok:
Accepted Sonar issue with a comment: https://next.sonarqube.com/sonarqube/project/issues?pullRequest=1478&issueStatuses=OPEN,CONFIRMED&open=4e25e51e-4ff2-40e3-b7a3-4e2806fd35d1&id=org.sonarsource.sonarlint.intellij:sonarlint-intellij

src/main/java/org/sonarlint/intellij/ui/factory/PanelFactory.kt
```diff
@@ -53,6 +64,47 @@ class PanelFactory {
...
+            c2: JComponent,
+            proportionProperty: String,
+            defaultSplit: Float,
+        ): JBSplitter {
+            val splitter = OnePixelSplitter(splitVertically(project), proportionProperty, defaultSplit)
+            splitter.setFirstComponent(c1)
+            splitter.setSecondComponent(c2)
+            splitter.setHonorComponentsMinimumSize(true)
+
+            val listener: ToolWindowManagerListener? = object : ToolWindowManagerListener {
```
nquinquenel: You can remove the type, it's not needed. Also the "?" is not needed as well (in case you keep it).

src/main/java/org/sonarlint/intellij/ui/factory/PanelFactory.kt
```diff
@@ -53,6 +64,47 @@ class PanelFactory {
...
+            parentDisposable: Disposable,
+            c1: JComponent,
+            c2: JComponent,
+            proportionProperty: String,
+            defaultSplit: Float,
+        ): JBSplitter {
+            val splitter = OnePixelSplitter(splitVertically(project), proportionProperty, defaultSplit)
+            splitter.setFirstComponent(c1)
+            splitter.setSecondComponent(c2)
+            splitter.setHonorComponentsMinimumSize(true)
```
nquinquenel: You can use `apply` from kotlin instead

src/main/java/org/sonarlint/intellij/ui/factory/PanelFactory.kt
```diff
@@ -53,6 +64,47 @@ class PanelFactory {
...
+
+            val listener: ToolWindowManagerListener? = object : ToolWindowManagerListener {
+                override fun stateChanged(toolWindowManager: ToolWindowManager) {
+                    splitter.setOrientation(splitVertically(project))
+                    parentComponent.revalidate()
+                    parentComponent.repaint()
+                }
+            }
+            project.messageBus.connect(parentDisposable)
+                .subscribe(ToolWindowManagerListener.TOPIC!!, listener!!)
```
nquinquenel: No need to put `!!`

src/main/java/org/sonarlint/intellij/ui/factory/PanelFactory.kt
```diff
@@ -53,6 +64,47 @@ class PanelFactory {
...
+            Disposer.register(parentDisposable) {
+                parentComponent.remove(splitter)
+                splitter.dispose()
+            }
+
+            return splitter
+        }
+
+        private fun splitVertically(project: Project): Boolean {
+            val toolWindow: ToolWindow? = ToolWindowManager.getInstance(project).getToolWindow(TOOL_WINDOW_ID)
```
nquinquenel: You can remove the type here as well

nquinquenel (commented):
I think it looks mostly good, I left a few comment, and also:
- I think the ScaTest is not running, if it's not, then we should not include it in the PR and remove it
- I think the messaging is not clear about which version is supported or not with SQ:S. What happen if you connect with an old version for example?

Krosovok:
> I think it looks mostly good, I left a few comment, and also:
> 
> * I think the ScaTest is not running, if it's not, then we should not include it in the PR and remove it
> * I think the messaging is not clear about which version is supported or not with SQ:S. What happen if you connect with an old version for example?

1. Removed ScaTests for now.
2. I took a quick look on SLCore and it is checked there but there is no API for providing plugins with SCA availability details. So as I can't fix it now I added it to scope of a separate ticket: https://sonarsource.atlassian.net/browse/SLI-2192

Krosovok (approved):

PR #1481, opened 2025-08-05
MMF-4645 Add SCA to change notes.

Krosovok:
[MMF-4645](https://sonarsource.atlassian.net/browse/MMF-4645)

[MMF-4645]: https://sonarsource.atlassian.net/browse/MMF-4645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -65,7 +65,7 @@
 
     <change-notes><![CDATA[
       <ul>
-        <li>10.29 - Early access for MISRA rules. Less FPs and FNs for Java. Introduction of a new tab 'Dependency Risks' for displaying results of SCA analysis if it is enabled. Allow changing status of separate vulnerabilities.</li>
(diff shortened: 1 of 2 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -65,7 +65,7 @@
 
     <change-notes><![CDATA[
       <ul>
-        <li>10.29 - Early access for MISRA rules. Less FPs and FNs for Java.</li>
+        <li>10.29 - Early access for MISRA rules. Less FPs and FNs for Java. Introduction of a new tab 'Dependency Risks' for displaying results of SCA analysis if it is enabled. Allow changing status of separate vulnerabilities.</li>
```
nquinquenel: We need to display from which version it's available, also we should not mention "SCA" directly, it's difficult to understand.

PR #1484, opened 2025-08-06
SLI-1531 Update to IntelliJ Platform Gradle Plugin (2.x)

nquinquenel:


README.md
```diff
@@ -125,24 +133,22 @@ How to develop in IntelliJ
...
-To run against a specific IDE, the `runIdeDirectory` property can be used as such:
+To run against a specific IDE, you can use the `runIdeDirectory` property to point to the directory of the IDE you want to run.
 
-    ./gradlew :runIde -PrunIdeDirectory=<IDE_PATH>
+For example:
 
-Keep in mind that the `clean` task will wipe out the content of `build/idea-sandbox`,
-so you will need to repeat some setup steps for that instance, such as configuring the JDK.
+```bash
+./gradlew :runLocalIde -runIdeDirectory=<path_to_ide>
```
eray-felek-sonarsource: -PrunIdeDirectory

clion-common/src/test/java/org/sonarlint/intellij/clion/common/AnalyzerConfigurationTest.java
```diff
@@ -1,151 +0,0 @@
...
-import java.util.List;
-import java.util.Map;
-import java.util.function.Predicate;
-import org.junit.jupiter.api.Test;
-
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
-
-class AnalyzerConfigurationTest {
```
eray-felek-sonarsource, line 34: Why did we remove this test?

.cirrus.yml
```diff
@@ -62,13 +61,13 @@ eks_builder_container: &BUILDER_CONTAINER_DEFINITION
...
   builder_image: docker-builder-v*
-  builder_instance_type: t3.small
+  builder_instance_type: t3.medium
   builder_subnet_id: ${CIRRUS_AWS_SUBNET}
 
 ec2_instance_definition: &WINVM_DEFINITION
   region: eu-central-1
-  type: c5.4xlarge
-  image: sonarlint-intellij-v*
+  type: c5.9xlarge
```
damien-urruty-sonarsource: This is huge, do we need that big of a machine?

.cirrus.yml
```diff
@@ -221,7 +224,7 @@ qa_task:
   eks_container:
     <<: *BUILDER_CONTAINER_DEFINITION
     cpu: 6
-    memory: 16G
+    memory: 22G
```
damien-urruty-sonarsource: Same remark :o

build.gradle.kts
```diff
@@ -236,111 +111,260 @@ dependencies {
...
+    excludes(
+        listOf("**/*.jar", "**/*.png", "**/README")
+    )
+    strictCheck = true
+}
+
+intellijPlatform {
+    pluginConfiguration {
+        ideaVersion {
+            sinceBuild = "242"
```
damien-urruty-sonarsource: This is really a big thing, we are making quite a big leap in supported versions. We need to at least check with PMs if it's OK. There is no way to keep a longer backward compatibility? Is it because of the new Gradle plugin or something else?

src/main/java/org/sonarlint/intellij/ui/tree/SecurityHotspotTree.java
```diff
@@ -32,7 +32,7 @@
 import com.intellij.util.EditSourceOnDoubleClickHandler;
 import com.intellij.util.EditSourceOnEnterKeyHandler;
 import javax.annotation.CheckForNull;
-import javax.annotation.Nullable;
+import org.jetbrains.annotations.Nullable;
```
damien-urruty-sonarsource: Why changing all those annotations, and what is the impact?

src/test/java/org/sonarlint/intellij/tasks/ConnectionTestTaskTests.kt
```diff
@@ -43,6 +44,7 @@ class ConnectionTestTaskTests : AbstractSonarLintLightTests() {
     }
 
     @Test
+    @Disabled("Flaky, timeout")
```
damien-urruty-sonarsource: Is it still flaky or was it at some point? Should we track it with a ticket?

damien-urruty-sonarsource (approved):
Given the number of changes, I made a cursory review. I did not find any blocker, but getting another review from someone else might help. The only concern for me is raising the minimal supported version to a quite recent version, we should make sure this decision is OK with all stakeholders. You will also need to bump the major version and document it in the release notes.

One way to reduce the PR size could have been to bump the minimal supported version in one PR, and change the Gradle plugin in another, but it's always easier to say after the fact.

Finally, I want to give a huge kudos for following this topic for so long and finally bringing it to completion! It was not an easy one.

.cirrus.yml
nquinquenel: Looking at the Develocity scan, the Windows task reaches the 32gb of memory allocated to the c5.4xlarge instance, during the execution of the tests. I tried with the next one, allowing double the memory to see if it's any better. I'm not exactly sure if there's a big benefit, I'm fine lowering it back.

build.gradle.kts
nquinquenel: PMs agreed on increasing the minimum supported version. I think it would be worth verifying the number of users before fully committing to it. It's technically within our 1 year policy of supported versions.
The reason I did the increase is that I faced strange compatibility issues when migrating the Gradle plugin. Some Kotlin classes existing on 1.8+ were found at compile time, but not at runtime. I suspect a mistake on my end regarding the configuration. I can try extracting the logic about this minimum version update into another PR, and see how it goes.

src/main/java/org/sonarlint/intellij/ui/tree/SecurityHotspotTree.java
nquinquenel: I think we should be consistent about annotations we use, and we had a mix of both JetBrains Nullable, and JSR305 Nullable. Here I made the switch to JetBrains, which feel more natural to use as we are in their IDE. Not sure there's any impact, however we still have other JSR305 annotations for package-info, CheckForNull, etc. So maybe we should stick to the non-JetBrains one in the end 🤷🏻

build.gradle.kts
nquinquenel: I was able to keep compatibility on 2022.3, the struggle was due to the [Kotlin Standard Library](https://plugins.jetbrains.com/docs/intellij/using-kotlin.html#kotlin-standard-library) that we excluded 🤔
I will proceed with reverting the changes of the ITs

eray-felek-sonarsource: I agree with the Nicolas here, I think we should check percentage of user's affected and we should increase the minimal version if it is less then %10 since it is in our 1 year policy. Trying to maintain IT's for less than %10 would not worth it I think

eray-felek-sonarsource (approved):
LGTM, great job! As discussed I think we should also keep track of Windows improvements from IntelliJ platform with a ticket to check if they fixed the runIde issue where it takes too long to launch

eray-felek-sonarsource (approved):

.cirrus.yml
nquinquenel: With the decrease to minimal supported version 2023.1, Windows tests are running much faster. I'm also reverting this change as a result as we do not reach the memory limit anymore

src/main/java/org/sonarlint/intellij/ui/tree/SecurityHotspotTree.java
nquinquenel: Moved back everything to javax annotation as a result of my previous comment

src/test/java/org/sonarlint/intellij/tasks/ConnectionTestTaskTests.kt
nquinquenel: I re-enabled the test for now - not exactly sure why it was the case during the migration, but now that the PR is stable and building, it seems better

build.gradle.kts
nquinquenel: I managed to support 2023.1. 2022.3 represented only 2% of the users, that's an acceptable number.

PR #1487, opened 2025-08-07
SLI-2188 Order possible transition options in the same way as on web UI

Krosovok:
[SLI-2188](https://sonarsource.atlassian.net/browse/SLI-2188)

[SLI-2188]: https://sonarsource.atlassian.net/browse/SLI-2188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarlint/intellij/ui/resolve/ChangeDependencyRiskStatusPanel.kt
+++ b/src/main/java/org/sonarlint/intellij/ui/resolve/ChangeDependencyRiskStatusPanel.kt
@@ -37,17 +37,28 @@ import org.sonarlint.intellij.ui.options.OptionPanel
 import org.sonarlint.intellij.ui.options.addComponents
 import org.sonarsource.sonarlint.core.client.utils.DependencyRiskTransitionStatus
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.DependencyRiskTransition
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.DependencyRiskTransition.ACCEPT
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.DependencyRiskTransition.CONFIRM
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.DependencyRiskTransition.REOPEN
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.DependencyRiskTransition.SAFE
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.tracking.DependencyRiskDto
 
 val TRANSITIONS_WITH_OPTIONAL_COMMENT: EnumSet<DependencyRiskTransition> = EnumSet.of(CONFIRM, REOPEN)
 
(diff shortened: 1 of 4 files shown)


Krosovok:
That's how it looks now. Showing the current status as well (as disabled and grayed-out). But You can't confirm the status unless you change it (and provide comment if needed). This way it is:

1. Aligned with web UI;
2. Gives user information what the current status is (because otherwise we don't provide this information and it is important);
3. Provide better understanding of which statuses exist.

image

src/main/java/org/sonarlint/intellij/ui/resolve/ChangeDependencyRiskStatusPanel.kt
```diff
@@ -37,45 +37,78 @@ import org.sonarlint.intellij.ui.options.OptionPanel
...
+        return OptionPanel(
+            transition.name,
+            transition.title,
+            transition.description
+        ).also {
+            it.statusRadioButton.addActionListener(this)
+            it.statusRadioButton.actionCommand = transition.name
+            it.isEnabled = transition in allowedTransitions
+            it.setSelected(transition == initialSelection)
+        }
```
nquinquenel, line 109: I think you can reply by `apply` and remove all the `it.`

src/main/java/org/sonarlint/intellij/ui/resolve/ChangeDependencyRiskStatusPanel.kt
```diff
@@ -100,7 +133,20 @@ class ChangeDependencyRiskStatusPanel(
             add(JBLabel("<a href=\"$link\">Formatting Help</a>:  *Bold*  ``Code``  * Bulleted point").apply { setCopyable(true) })
         }
     }
-    private fun isReady(newValue: DependencyRiskTransition?) = newValue != null && isReady()
+    private fun isReady(newValue: DependencyRiskTransition?): Boolean {
+        return toOtherEnum(newValue) in allowedTransitions
+            && isReady()
+    }
+
+    private fun toOtherEnum(value: DependencyRiskTransition?) : DependencyRiskTransitionStatus? {
```
nquinquenel:
```suggestion
    private fun toOtherEnum(value: DependencyRiskTransition?): DependencyRiskTransitionStatus? {
```

nquinquenel: Name feels a bit strange, maybe `toTransitionChangeEnum`?

src/main/java/org/sonarlint/intellij/ui/resolve/ChangeDependencyRiskStatusPanel.kt
Krosovok: I tried, but then `...addActionListener(this)` is ambiguous. I don't know if we can differentiate between local `this` and global `this`. And even if we can this is a bit more straightforward.

PR #1494, opened 2025-08-19
SLI-2150 Implement the unified Current File tab

nquinquenel:
[SLI-2150](https://sonarsource.atlassian.net/browse/SLI-2150)

- New unified view
- New filtering panel

This is a first version handling the essential. Improvements will come afterwards, as well as the removal of the other tabs, in order to make incremental changes.

[SLI-2150]: https://sonarsource.atlassian.net/browse/SLI-2150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/sonarlint/intellij/actions/filters/IncludeResolvedFindingsAction.java
```diff
@@ -50,18 +44,11 @@ public boolean isSelected(AnActionEvent event) {
...
+    if (p != null && type == Finding.class) {
         isResolved = flag;
-        getService(p, SonarLintToolWindow.class).filterSecurityHotspotTab(isResolved);
-      } else if (type == LiveIssue.class) {
-        isResolved = flag;
-        getService(p, SonarLintToolWindow.class).filterCurrentFileTab(isResolved);
-      } else if (type == LocalTaintVulnerability.class) {
-        isResolved = flag;
-        getService(p, SonarLintToolWindow.class).filterTaintVulnerabilityTab(isResolved);
+        //getService(p, SonarLintToolWindow.class).filterCurrentFileTab(isResolved);
```
Krosovok, line 50: If you intend to get back to it later, please add a TODO.

src/main/java/org/sonarlint/intellij/finding/sca/LocalDependencyRisk.kt
```diff
@@ -34,9 +37,64 @@ class LocalDependencyRisk(serverDependencyRisk: DependencyRiskDto) {
...
+
+    override fun getRuleDescriptionContextKey() = null
+
+    override fun file() = null
+
+    override fun isValid() = true
+
+    override fun isOnNewCode() = true
+
+    override fun isResolved() = status in listOf(DependencyRiskDto.Status.SAFE, DependencyRiskDto.Status.ACCEPT, DependencyRiskDto.Status.FIXED)
```
Krosovok: Will `listOf()` be called here for each `isResolved()` call? I would move it out to initialize it once elsewhere. And would use `EnumSet` to utilize optimized `contains()` while at it.

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFileDisplayManager.kt
```diff
@@ -0,0 +1,126 @@
...
+        
+        // Try to keep the same selection if possible
+        if (prev != null) {
+            for (i in newOptions.indices) {
+                if (newOptions[i].equals(prev, ignoreCase = true)) {
+                    filtersPanel.severityCombo.selectedIndex = i
+                    break
+                }
+            }
+        }
```
Krosovok: Would't just `filtersPanel.severityCombo.selectedItem = prev` work here? The only significant difference that I see is that it would not be case insensitive but it doesn't seem to matter here.

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFileFindingsFilter.kt
```diff
@@ -0,0 +1,206 @@
...
+ * <h3>Filtering Strategy:</h3>
+ * <p>Uses a multi-stage filtering approach:</p>
+ * <ul>
+ *   <li><strong>Data Loading:</strong> Retrieves raw findings from various caches and holders</li>
+ *   <li><strong>Type-Specific Filtering:</strong> Applies filters appropriate to each finding type</li>
+ *   <li><strong>Unified Criteria:</strong> Uses {@link FilterCriteria} for consistent filter application</li>
+ *   <li><strong>Result Aggregation:</strong> Returns filtered results in a structured {@link FilteredFindings} container</li>
+ * </ul>
+ */
+class CurrentFileFindingsFilter(private val project: Project) {
```
Krosovok: Fro some reason Idea refuses to render this JavaDoc for me. It's weird because the markup seems right.
image

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFileFindingsFilter.kt
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    private fun filterIssueBySeverity(issue: LiveIssue, criteria: FilterCriteria): Boolean {
+        if (criteria.severityFilter == "All") return true
+        return if (criteria.isMqrMode) {
+            issue.getHighestImpact() != null && issue.getHighestImpact()?.name.equals(criteria.severityFilter, ignoreCase = true)
+        } else {
+            issue.userSeverity != null && issue.userSeverity?.name.equals(criteria.severityFilter, ignoreCase = true)
+        }
+    }
```
Krosovok: How about flipping equals here to avoid null checks? `criteria.severityFilter` is not nullable, right?
This will also allow to avoid the `equals` repeating as we always call it on the same string. `if` can be moved to a parameter or to a previous line.

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFileFindingsFilter.kt
```diff
@@ -0,0 +1,206 @@
...
+
+    private fun filterIssueByText(issue: LiveIssue, criteria: FilterCriteria): Boolean {
+        if (criteria.textFilter.isEmpty()) return true
+        val filterText = criteria.textFilter.lowercase(Locale.getDefault())
+        return (issue.message != null && issue.message.lowercase(Locale.getDefault()).contains(filterText))
+            || issue.getRuleKey().lowercase(Locale.getDefault()).contains(filterText)
+            || issue.file().name.lowercase(Locale.getDefault()).contains(filterText)
```
Krosovok: Isn't there a utility method for that? I found `Strings.CI.contains()` for instance. Or is locale passing here crucial?

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFileFindingsFilter.kt
```diff
@@ -0,0 +1,206 @@
...
+import org.sonarlint.intellij.finding.issue.vulnerabilities.LocalTaintVulnerability
+import org.sonarlint.intellij.finding.issue.vulnerabilities.TaintVulnerabilitiesCache
+import org.sonarlint.intellij.finding.sca.DependencyRisksCache
+import org.sonarlint.intellij.finding.sca.LocalDependencyRisk
+
+/**
+ * Filter criteria for findings
+ */
+data class FilterCriteria(
+    val severityFilter: String = "All",
```
Krosovok: I would move it a constant named something like `NO_FILTER`.

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFileFindingsFilter.kt
```diff
@@ -0,0 +1,206 @@
...
+            || risk.packageVersion.lowercase(Locale.getDefault()).contains(filterText)
+            || (risk.vulnerabilityId != null && risk.vulnerabilityId.lowercase(Locale.getDefault()).contains(filterText))
+    }
+
+    private fun filterDependencyRiskByQuickFix(risk: LocalDependencyRisk, criteria: FilterCriteria): Boolean {
+        return !criteria.quickFixFilter || risk.isAiCodeFixable()
+    }
+
+    private fun filterByStatus(finding: Finding, criteria: FilterCriteria): Boolean {
+        return criteria.statusFilter == "All" || (criteria.statusFilter == "Open" && !finding.isResolved()) || (criteria.statusFilter == "Resolved" && finding.isResolved())
```
Krosovok: I would convert this long line to a `when (criteria.statusFilter)` clause.

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFileFindingsPanel.kt
```diff
@@ -0,0 +1,339 @@
...
+
+/**
+ * Base panel class for displaying findings in tree structures with common functionality and infrastructure.
+ * 
+ * <h3>Design & Architecture:</h3>
+ * <p>This abstract base class provides the foundation for displaying findings in tree-based UI components.
+ * It implements a factory pattern for creating different types of tree configurations and manages the 
+ * common infrastructure needed by all findings display panels.</p>
+ */
+open class CurrentFileFindingsPanel(val project: Project) : SimpleToolWindowPanel(false, false), Disposable, DataProvider {
```
Krosovok: Judging be the JavaDoc it should be `abstract`, not just `open`.

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFilePanel.kt
```diff
@@ -0,0 +1,497 @@
...
+        super.setContent(splitter)
+    }
+
+    private fun setupToolbar() {
+        setToolbar(listOf(
+            SonarLintActions.getInstance().analyzeCurrentFileAction(),
+            SonarLintActions.getInstance().analyzeChangedFiles(),
+            SonarLintActions.getInstance().analyzeAllFiles(),
+            ActionManager.getInstance().getAction("SonarLint.toolwindow.Cancel"),
+            ActionManager.getInstance().getAction("SonarLint.toolwindow.Configure"),
```
Krosovok: Is there a reason not to get them from `SonarLintActions` as well?

src/main/java/org/sonarlint/intellij/ui/currentfile/tree/SingleFileDependencyRiskTreeModelBuilder.kt
```diff
@@ -0,0 +1,125 @@
...
+            SummaryUiModel(
+                riskSeverity(it.severity),
+                backgroundColorsByImpact[impactSeverity],
+                borderColorsByImpact[impactSeverity]
+            )
+        } ?: SummaryUiModel()
+    }
+
+    private fun getRiskWithHighestSeverity(): LocalDependencyRisk? {
+        return latestRisks.minByOrNull { DEPENDENCY_RISK_SEVERITY_ORDER.indexOf(it.severity) }
```
Krosovok, line 122: It kinda feels off when to get `Highest` we use `min...`. I'd either reverse order in the list or just use ordinal of enum (which is not a good practice but as we can't add there a meaningful comparison method I think it can acceptable). Ideally should have made the enum itself `Comparable`. But now that it's in other project it's not feasible.

src/main/java/org/sonarlint/intellij/ui/currentfile/tree/SingleFileHotspotTreeModelBuilder.kt
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    override fun findFindingByKey(key: String): LiveSecurityHotspot? {
+        return findHotspotNode(key)?.hotspot
+    }
+
+    override fun getSummaryUiModel(): SummaryUiModel {
+        val severity = latestHotspots
+            .map { it.vulnerabilityProbability }
+            .minByOrNull { VULNERABILITY_PROBABILITIES.indexOf(it) }
```
Krosovok, line 99: Same min/max here.

src/main/java/org/sonarlint/intellij/ui/currentfile/tree/SingleFileIssueTreeModelBuilder.kt
```diff
@@ -0,0 +1,160 @@
...
+        } ?: SummaryUiModel()
+    }
+
+    private fun getIssueWithHighestSeverity(): LiveIssue? {
+        return latestIssues.minByOrNull { SEVERITY_ORDER.indexOf(it.userSeverity) }
+    }
+
+    private fun getIssueWithHighestImpact(): LiveIssue? {
+        return latestIssues.minByOrNull { IMPACT_ORDER.indexOf(it.getHighestImpact()) }
+    }
```
Krosovok, line 143: Same here.

src/main/java/org/sonarlint/intellij/ui/currentfile/tree/SingleFileTaintTreeModelBuilder.kt
```diff
@@ -0,0 +1,147 @@
...
+            }
+        } ?: SummaryUiModel()
+    }
+
+    private fun getIssueWithHighestSeverity(): LocalTaintVulnerability? {
+        return latestTaints.minByOrNull { TAINT_SEVERITY_ORDER.indexOf(it.severity()) }
+    }
+
+    private fun getIssueWithHighestImpact(): LocalTaintVulnerability? {
+        return latestTaints.minByOrNull { TAINT_IMPACT_ORDER.indexOf(it.getHighestImpact()) }
```
Krosovok, line 144: Same here.

src/main/java/org/sonarlint/intellij/ui/currentfile/tree/SingleFileDependencyRiskTreeModelBuilder.kt
Krosovok: Seeing in how many places such comparisons are maybe it's better to move it somewhere? I know that there is also a bunch of similarly-looking enums to add on top of that. But I think all this ordering/comparing logic of Serverity/Impact should have some separate place.

src/main/java/org/sonarlint/intellij/actions/filters/IncludeResolvedFindingsAction.java
nquinquenel: I'm adding a TODO 👍🏻 
This will be refactored in the next batch of changes

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFileFindingsFilter.kt
nquinquenel: Interesting, I also notice the same when looking at the rendering. I will remove tags that break this

eray-felek-sonarsource (commented):
I think we could group check box filters and drop down ones together currently date sort comes after fix suggestion. We could also right align the filter.  I think we should also remove "No" text when there aren't any hotspots/vulnerabilities/dependency issues. Also Show/Hide issue functionality feels a bit hidden and not sure what's would be to use case of it given that we already have a filter to hide issues that we are not interested in

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFileFindingsFilter.kt
nquinquenel: Good point! I'll switch to `Strings.CI.contains(x, y)`

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFileFindingsFilter.kt
nquinquenel: 👍🏻 
I'll also introduce a new Enum for the resolve states

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFileDisplayManager.kt
nquinquenel: Actually I think I'll drop this part, it doesn't make sense. We try to match MQR vs Standard mode but I don't think it will ever work. If we switch to the other mode, just fallback to the default value

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFileFindingsPanel.kt
nquinquenel: Good catch

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFilePanel.kt
nquinquenel: No big reasons here. Just need to be careful, all the actions are not declared in `ActionManager`. I'll switch to the standard `SonarLintActions` call

src/main/java/org/sonarlint/intellij/ui/currentfile/CurrentFileFindingsFilter.kt
nquinquenel: I'll use `Strings.CI.equals`, more convenient, as suggested by your other comment!

src/main/java/org/sonarlint/intellij/ui/currentfile/tree/SingleFileDependencyRiskTreeModelBuilder.kt
nquinquenel: I'll pass on this one - I can't find a good solution

nquinquenel:
> I think we could group check box filters and drop down ones together currently date sort comes after fix suggestion. We could also right align the filter. I think we should also remove "No" text when there aren't any hotspots/vulnerabilities/dependency issues. Also Show/Hide issue functionality feels a bit hidden and not sure what's would be to use case of it given that we already have a filter to hide issues that we are not interested in

Thanks for the feedback! I'll implement some of these in the next batch of changes 👍🏻

PR #1509, opened 2025-08-28
SLI-2227 Handle report on multiple modules

nquinquenel:
[SLI-2227](https://sonarsource.atlassian.net/browse/SLI-2227)

When a report is triggered on multiple modules, there's a new tab created for each module. Instead, we should group all the results into one tab.

[SLI-2227]: https://sonarsource.atlassian.net/browse/SLI-2227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarlint/intellij/callable/ShowReportCallable.kt
+++ b/src/main/java/org/sonarlint/intellij/callable/ShowReportCallable.kt
@@ -27,7 +27,9 @@ import org.sonarlint.intellij.editor.CodeAnalyzerRestarter
 import org.sonarlint.intellij.ui.UiUtils.Companion.runOnUiThread
 import org.sonarlint.intellij.ui.report.ReportTabManager
 
-class ShowReportCallable(private val project: Project, private val batchId: String = generateBatchId()) : AnalysisCallback {
+class ShowReportCallable(private val project: Project) : AnalysisCallback {
+
+    private val batchId: String = generateBatchId()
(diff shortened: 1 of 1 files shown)


src/main/java/org/sonarlint/intellij/callable/ShowReportCallable.kt
```diff
@@ -20,35 +20,33 @@
...
 import org.sonarlint.intellij.analysis.AnalysisResult
 import org.sonarlint.intellij.common.util.SonarLintUtils
 import org.sonarlint.intellij.editor.CodeAnalyzerRestarter
 import org.sonarlint.intellij.ui.UiUtils.Companion.runOnUiThread
+import org.sonarlint.intellij.ui.report.ReportTabManager
 
-class ShowReportCallable(private val project: Project) : AnalysisCallback {
-
-    private var results: AnalysisResult? = null
+class ShowReportCallable(private val project: Project, private val batchId: String = generateBatchId()) : AnalysisCallback {
```
damien-urruty-sonarsource: It seems there is no invocation of `ShowReportCallable` that passes a `batchId` argument. So you could simply have it as a val int he class

damien-urruty-sonarsource (approved):

PR #1512, opened 2025-08-28
SLI-2224 SLI-2221 SLI-2215 SLI-2213 SLI-2211 SLI-2230 SLI-2228 Hardening 11.0

nquinquenel:


src/main/java/org/sonarlint/intellij/trigger/SonarLintCheckinHandler.java
```diff
@@ -110,7 +110,7 @@ public ReturnResult beforeCheckin(@Nullable CommitExecutor executor, PairConsume
 
       if (analysisIdsByCallback == null) {
         SonarLintConsole.get(project).debug("Pre-commit analysis cancelled because analysis did not start");
-        return ReturnResult.CANCEL;
+        return ReturnResult.COMMIT;
```
Krosovok, line 113: Just to double check - is this change intentional? To make sure it's not some side effect of refactoring or something.

nquinquenel: Yes, it's intentional! When no files are to be analyzed (only deleted files for example), we don't trigger an analysis, so we should proceed with the commit instead of canceling it, that was the issue reported by some users

nquinquenel: In general, I changed the pre-commit handling so that it always ask for users to continue or to abort, even in case of failure. We don't return a CANCEL state directly anymore

PR #1522, opened 2025-09-03
SLI-2238 Fix sorting in the Findings tab

nquinquenel:


nquinquenel:
cc @Krosovok for visibility

src/main/java/org/sonarlint/intellij/editor/RangeBlinker.java
```diff
@@ -40,14 +41,15 @@ public class RangeBlinker {
...
+  private final Alarm myBlinkingAlarm;
   private final TextAttributes myAttributes;
   private final List<RangeHighlighter> myAddedHighlighters = new ArrayList<>();
 
-  public RangeBlinker(Editor editor, final TextAttributes attributes, int timeToLive) {
+  public RangeBlinker(Editor editor, final TextAttributes attributes, int timeToLive, Disposable parentDisposable) {
     myAttributes = attributes;
     myEditor = editor;
     myTimeToLive = timeToLive;
+    myBlinkingAlarm = new Alarm(parentDisposable);
```
damien-urruty-sonarsource: Good catch! I forgot that we create one Alarm per location, which seems huge. The parent disposable being the project, all alarms will be disposed only when the project is closed. I think this is a memory leak, it might be pretty bad, to be checked. We could try to dispose the alarms as soon as they are not useful. Or maybe rework the logic and have a single alarm for the project, making all ranges blink at the same time

damien-urruty-sonarsource (approved):
Not directly related to your changes, but there might be something important to address

src/main/java/org/sonarlint/intellij/editor/RangeBlinker.java
nquinquenel: Right, I refactored this!

PR #1526, opened 2025-09-08
SLI-2200 Support Flight Recorder mode

nquinquenel:
[SLI-2200](https://sonarsource.atlassian.net/browse/SLI-2200)

[SLI-2200]: https://sonarsource.atlassian.net/browse/SLI-2200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/sonarlint/intellij/core/BackendService.kt
```diff
@@ -1131,4 +1168,38 @@ class BackendService : Disposable {
...
+    fun disableFlightRecorder() {
+        getService(SonarLintGlobalSettingsStore::class.java).state?.let {
+            val newSettings = SonarLintGlobalSettings(it)
+            newSettings.isFlightRecorderEnabled = false
+            getService(SonarLintGlobalSettingsStore::class.java).save(newSettings)
+
+            ApplicationManager.getApplication().messageBus.syncPublisher(GlobalConfigurationListener.TOPIC)
+                .applied(it, newSettings)
+
+            restartBackendService(true)
```
jblievremont: This looks like almost duplication from `enableFlightRecorder`, the only difference being a boolean value. Maybe this could be refactored?

nquinquenel: Good point, I also made it less verbose, we have utility methods to access the global settings

PR #1527, opened 2025-09-12
SLI-378 Password migration

Krosovok:
I introduced migration of credentials that were saved with outdated `PasswordUtil`. New `PasswordSafe` is used now. To implement this change, I moved all credentials logic to a separate `CredentialsService`.

1) When IDE with credentials stored in an old-fashioned way (either dogfood or connection ones), they are migrated to a new storage.
2) Old data is wiped and forcefully saved to erase the data from configuration files.
3) All interactions with credentials is done with new API. Old API is left for migration purposes (to read old data).
4) Made a new `CredentialsChangeService` to handle changes in credentials (as they are no longer part of settings themselves). SLCore handles it separately anyways.


Krosovok:
@damien-urruty-sonarsource I would appreciate if you can take a look too.

build.gradle.kts
```diff
@@ -93,6 +93,7 @@ dependencies {
         pluginComposedModule(runtimeOnly(project(":git")))
         bundledPlugins("com.intellij.java", "Git4Idea")
         testFramework(TestFrameworkType.Platform)
+        testFramework(TestFrameworkType.JUnit5)
```
nquinquenel: Do we need this in the end?

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettingsStore.java
```diff
@@ -45,13 +54,60 @@ public SonarLintGlobalSettings getState() {
...
+    modificationCount = 0;
+  }
+
+  @Override
+  public void initializeComponent() {
+    migrateCredentials();
+  }
+
+  private void migrateCredentials() {
+    CredentialsService credentialsService = SonarLintUtils.getService(CredentialsService.class);
```
nquinquenel:
```suggestion
    var credentialsService = SonarLintUtils.getService(CredentialsService.class);
```

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettingsStore.java
```diff
@@ -45,13 +54,60 @@ public SonarLintGlobalSettings getState() {
...
+    var serverConnections = settings.getServerConnections();
+
+    if (serverConnections.stream().anyMatch(credentialsService::hasOldCredentials)) {
+      Runnable migration = () -> performCredentialsMigration(serverConnections, credentialsService);
```
nquinquenel, line 70:
```suggestion
      var migration = () -> performCredentialsMigration(serverConnections, credentialsService);
```

src/main/java/org/sonarlint/intellij/sharing/AutomaticSharedConfigCreator.kt
```diff
@@ -151,7 +154,8 @@ class AutomaticSharedConfigCreator(
         val currBindingSuggestion = getSettingsFor(project).isBindingSuggestionsEnabled
         try {
             getSettingsFor(project).isBindingSuggestionsEnabled = false
-            val serverConnectionBuilder = ServerConnection.newBuilder().setDisableNotifications(false).setToken(String(tokenField.password))
+            val serverConnectionBuilder = ServerConnection.newBuilder().setDisableNotifications(false)
+//                .setToken(String(tokenField.password))
```
nquinquenel: What is this commented code?

nquinquenel (approved):
Looks good, I still need to test on Windows. Let's try to merge this so we can dogfood it.

src/main/java/org/sonarlint/intellij/config/global/SonarLintGlobalSettingsStore.java
Krosovok: Doesn't compile. Says for lambda it should be explicit.

PR #1621, opened 2026-01-27
SLI-2464 Do not refresh new code definition on each finding raised

nquinquenel:


voz:
@nquinquenel can we close it or define path forward if we need to? Thx for helping to keep things tidy.

PR #1710, opened 2026-04-01
SLI-2569 SLI-2589 Avoid crash on pre-commit and avoid infinite polling

nquinquenel:


src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.kt
```diff
@@ -107,23 +110,41 @@ class AnalysisSubmitter(private val project: Project) {
...
+
+    private fun awaitPreCommitAnalysisStart(module: Module, files: List<VirtualFile>, callback: CheckInCallable, analysisIds: MutableList<UUID>) {
+        getService(project, PromotionProvider::class.java).handlePromotionOnPreCommitCheck()
+        val future = getService(BackendService::class.java).analyzeFileList(module, files)
+        try {
+            future[5, TimeUnit.SECONDS].analysisId?.let { analysisId ->
+                getService(project, RunningAnalysesTracker::class.java).track(AnalysisState(analysisId, callback, module))
+                analysisIds.add(analysisId)
             }
+        } catch (e: TimeoutException) {
```
nquinquenel, line 126: The main change in this class is this wrapper with the try/catch around the future, to avoid propagating the exception. Otherwise it's just clean up in different methods

src/main/java/org/sonarlint/intellij/analysis/AnalysisSubmitter.kt
```diff
@@ -107,23 +110,41 @@ class AnalysisSubmitter(private val project: Project) {
...
+    }
+
+    private fun reportPreCommitAnalysisStartFailure(callback: CheckInCallable, error: Throwable, message: String) {
+        SonarLintConsole.get(project).error(message, error)
+        callback.onError(error)
+    }
+
+    private fun preCommitAnalysisResult(callback: CheckInCallable, analysisIds: List<UUID>): Pair<CheckInCallable, List<UUID>>? = when {
+        analysisIds.isNotEmpty() -> Pair.of(callback, analysisIds)
+        !callback.analysisSucceeded() -> Pair.of(callback, analysisIds)
```
damien-urruty-sonarsource: You could group those 2 cases I think

src/main/java/org/sonarlint/intellij/trigger/SonarLintCheckinHandler.java
```diff
@@ -119,9 +120,13 @@ public ReturnResult beforeCheckin(@Nullable CommitExecutor executor, PairConsume
         var future = new CompletableFuture<Boolean>();
         var startTime = System.currentTimeMillis();
         var checkTask = runPreCommitAnalysis(future, indicator, analysisIds, startTime);
-        // Avoid busy-waiting
-        scheduledExecutor.scheduleAtFixedRate(checkTask, 0, 100, TimeUnit.MILLISECONDS);
-        return waitForFuture(indicator, future);
+        var polling = scheduledExecutor.scheduleAtFixedRate(checkTask, 0, 100, TimeUnit.MILLISECONDS);
```
damien-urruty-sonarsource: I think this code is more complex than it should, or maybe I miss something? First, the comment was wrong, because there is busy waiting, so good call for removing it. But I would argue that the executor is not useful, and it brings some indirection. In `waitForFuture`, we are checking for the indicator success flag every 100ms, and in the executor, we are checking if we need to resolve the future every 100ms. We need busy-waiting here, because their API is synchronous, couldn't we just wait in a loop until the analysis ID is completed?

src/main/java/org/sonarlint/intellij/trigger/SonarLintCheckinHandlerFactory.java
```diff
@@ -19,16 +19,30 @@
...
 
 public class SonarLintCheckinHandlerFactory extends CheckinHandlerFactory {
 
   @NotNull @Override public CheckinHandler createHandler(CheckinProjectPanel panel, CommitContext commitContext) {
     var project = panel.getProject();
-    return new SonarLintCheckinHandler(project, panel);
+    var handler = new SonarLintCheckinHandler(project, panel);
+    var workflowHandler = panel.getCommitWorkflowHandler();
+    if (workflowHandler instanceof Disposable workflowDisposable) {
+      Disposer.register(workflowDisposable, handler);
```
damien-urruty-sonarsource: If we get rid of the executor, it would also simplify this code

damien-urruty-sonarsource (approved):
LGTM, if you have time you could clean up even a little bit more IMO

PR #1721, opened 2026-04-10
BUILD-10988 Migrate GitHub Actions workflows to sonar-*-public runners

hedinasr:
## BUILD-10988: Migrate sonarlint-intellij to self-hosted runners

Replace `github-ubuntu-latest-s`/`github-ubuntu-latest-m` runners with SonarSource self-hosted runners across all workflows:

- **`sonar-xs`** — all lightweight jobs that don't need a container: `build-number`, `aws-auth`, `promote`, `dogfood`, `plugin-verifier-nightly`, the `shadow_scans` `aws-auth` job, and all PR/review event workflows (`PullRequestClosed`, `PullRequestCreated`, `RequestReview`, `SubmitReview`, `notify-failure`).
- **`sonar-m-docker`** — the container-based jobs: `build-plugin`, `test-and-sonar`, `qa`, and the `shadow_scans` `scan` job. These require Docker-in-Docker / a pre-baked toolchain image.
- **`sonar-m`** — `verify-plugin` and the `shadow_scans` `iris` job. These use a container image but do not need Docker-in-Docker.

Also updates the Develocity endpoint from `develocity-public.sonar.build` to `develocity.sonar.build`, since `sonar-m-docker` is on the internal network segment.

Part of the Milestone 5 migration effort tracked under [BUILD-10864](https://sonarsource.atlassian.net/browse/BUILD-10864).

## Container-user change: `--user 0` on sonar-m-docker

`sonar-m-docker` runs a **rootless** Docker-in-Docker daemon (configured in [`github-runners-infra`](https://github.com/SonarSource/github-runners-infra/blob/master/infra/applications/github-runners/values/runner.yaml.gotmpl)). The daemon runs as host UID 1001 and applies a user-namespace remapping to every container it spawns (`runner:100000:65536` in subuid/subgid).

Because of that remapping, the UID seen inside a job container is not the same as the UID on the node. Only container UID `0` maps back to the rootless daemon's own host UID (1001), which is the UID that owns the shared `_work/_temp` volume. Any other in-container UID lands in the remapped range (`100000`+) and therefore cannot write to files the runner created.

The progression on this PR:

1. First attempt used the image's default user (`--user runner`, UID 1654 in this image). The `actions/checkout` step failed immediately on writing the `save_state` file:
   ```
   Error: EACCES: permission denied,
     open '/__w/_temp/_runner_file_commands/save_state_<uuid>'
   ```
2. Second attempt used `--user 1001` to "match the ARC runner UID". Same EACCES — 1001 *inside* a rootless-DinD child container is not 1001 on the host (it's the remapped `101000`).
3. Final fix: `--user 0`. In a rootless userns, container UID 0 = host UID 1001, which owns `_work/_temp`, so `save_state` succeeds and subsequent Gradle/test steps own the caches they create.

**Why `--user 0` is not "running as root":** the kernel enforces the userns mapping. Container UID 0 in this runner has no path to host UID 0 — it is host UID 1001 with DAC bypass only inside its own namespace.

```
### /proc/self/uid_map  (container_uid  host_uid  count)
         0       1001          1
         1     100000      65536
```

## Privilege drop for SonarQube integration tests

`--user 0` creates a conflict with Elasticsearch 8.x, which has an **unconditional** root check (`can not run elasticsearch as root`) with no bypass property. The `es.insecure.allow.root` JVM property from ES 7.x was removed in ES 8.0.

This affects QA matrix jobs that start a SonarQube instance via the Orchestrator library (ConfigurationTests, ConnectedAnalysisTests, OpenInIdeTests on IC-2024.2).

**Fix:** create a non-root user (`sqtester`) and run only the Gradle integration test step via `su`, while all other steps (checkout, IDE launch, artifact upload) remain as UID 0:

```yaml
- name: Prepare non-root user for SonarQube tests
  run: |
    useradd -m -s /bin/bash sqtester
    chmod -R a+rwX ${GITHUB_WORKSPACE}
    chmod a+rx /root
    chmod -R a+rwX /root/.gradle
    chmod -R a+rx /github/home/.local/share/mise

- name: Run integration tests
  env:
    GRADLE_USER_HOME: /root/.gradle
  run: |
    su sqtester -p -s /bin/bash -c "export PATH='${PATH}' && ./gradlew :its:check ..."
```

Key detail: `su -p` preserves environment variables (credentials, GRADLE_USER_HOME) but still resets `PATH`, so an explicit `export PATH` is needed to retain mise-installed tools (mvn, java).

[BUILD-10864]: https://sonarsource.atlassian.net/browse/BUILD-10864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mikolaj-matuszny-ext-sonarsource (approved):

nquinquenel (approved):

nquinquenel (approved):
👌🏻

PR #1723, opened 2026-04-10
DEX-23 Update image without ReSharper, and update change notes

nquinquenel:


.github/workflows/build.yml
```diff
@@ -60,7 +60,7 @@ jobs:
   build-plugin:
     runs-on: github-ubuntu-latest-m
     container:
-      image: 275878209202.dkr.ecr.eu-central-1.amazonaws.com/sonarlint-intellij:20260318124419
+      image: 275878209202.dkr.ecr.eu-central-1.amazonaws.com/sonarlint-intellij:20260410161051
```
damien-urruty-sonarsource, line 63: Just out of curiosity, how do you find this information usually? Is it provided by Eng XP?

damien-urruty-sonarsource (approved):
LGTM!

.github/workflows/build.yml
nquinquenel: For now it is provided by Eng XP, but they will try making an improvement so we can see the version ourselves. Currently we don't have access to the account to check prod images, and the CI that promotes the image to prod only prints staging version.

PR #1733, opened 2026-04-20
SLI-2621 Update bouncycastle dependency

sophio-japharidze-sonarsource:


nquinquenel (approved):
LGMT, I remoevd the bouncycastle dep, it's not used anymore

PR #1739, opened 2026-04-24
BUILD-10988 Use warp-custom-ubuntu-24-04

mikolaj-matuszny-ext-sonarsource:
Use warpbuild docker runer instead of old sonar-m-docker runner.


nquinquenel (approved):
LGTM, thanks!

PR #1753, opened 2026-04-30
SLI-2644 Prepare bug fix version 12.2.1

kirill-knize-sonarsource:


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=12.3-SNAPSHOT
+version=12.2.1
```
nquinquenel: Keep the -SNAPSHOT

gradle/verification-metadata.xml
```diff
@@ -1270,6 +1270,11 @@
             <sha256 value="fcbbf1c18c9047e6917d384f2064e21b42a58fc85fb0523a9e2e015eff7c619d" origin="Generated by Gradle"/>
          </artifact>
       </component>
+      <component group="com.fasterxml" name="oss-parent" version="76">
+         <artifact name="oss-parent-76.pom">
+            <sha256 value="109e6c526eb2dbda3a1b21700fec4be2ddab24005ab9f1f80aa934591a1d8beb" origin="Generated by Gradle"/>
+         </artifact>
+      </component>
```
nquinquenel, line 1277: Technically we remove manually the previous checksum but it's ok if you don't - we can clean up later

kirill-knize-sonarsource: Removed

gradle.properties
kirill-knize-sonarsource: Recovered

src/main/java/org/sonarlint/intellij/ui/risks/tree/LocalDependencyRiskRenderer.kt
```diff
@@ -64,10 +64,10 @@ object LocalDependencyRiskRenderer : NodeRenderer<LocalDependencyRisk> {
...
             details += " [${node.cvssScore}] ${node.vulnerabilityId}"
         }
-        details = if (node.type == DependencyRiskDto.Type.PROHIBITED_LICENSE) {
-            "$details (Prohibited License)"
-        } else {
-            "$details (Vulnerability)"
+        details = when (node.type) {
+            DependencyRiskDto.Type.PROHIBITED_LICENSE -> "$details (Prohibited License)"
+            DependencyRiskDto.Type.MALWARE -> "$details (Malware)"
+            else -> "$details (Vulnerability)"
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: @kirill-knize-sonarsource I think the bot has a point here. Couldn't we explicitely mention all types in the when, so compilation fails the next time a new value is added to the enum?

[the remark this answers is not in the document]
kirill-knize-sonarsource: Done.

src/main/java/org/sonarlint/intellij/ui/risks/tree/LocalDependencyRiskRenderer.kt
```diff
@@ -64,10 +64,10 @@ object LocalDependencyRiskRenderer : NodeRenderer<LocalDependencyRisk> {
...
         }
-        details = if (node.type == DependencyRiskDto.Type.PROHIBITED_LICENSE) {
-            "$details (Prohibited License)"
-        } else {
-            "$details (Vulnerability)"
+        details = when (node.type) {
+            DependencyRiskDto.Type.VULNERABILITY -> "$details (Vulnerability)"
+            DependencyRiskDto.Type.PROHIBITED_LICENSE -> "$details (Prohibited License)"
+            DependencyRiskDto.Type.MALWARE -> "$details (Malware)"
         }
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: `DependencyRiskDto.Type` is bundled with the plugin at a fixed version — it is not loaded dynamically from the server. When sonarlint-core adds a new type, we bump the dependency and recompile. The exhaustive when without else will then fail to compile, forcing us to handle the new case explicitly. That is exactly the desired behaviour, as demonstrated in this PR. A `NoWhenBranchMatchedException` at runtime is not possible here because the compile-time and runtime enum are always the same jar.

nquinquenel (approved):

PR #1806, opened 2026-06-17
SLI-2733 Update vulnerable dependencies

kirill-knize-sonarsource:
----
## Summary by Gitar

- **Dependency security patches:**
  - Force update `org.springframework` to `6.2.19` to address CVE-2026-41850 and CVE-2026-41851.
  - Upgrade `io.opentelemetry:opentelemetry-api` to `1.62.0` to resolve CVE-2026-45292.
- **Build infrastructure:**
  - Add `resolutionStrategy` to `build.gradle.kts` and `gradle/module-conventions.gradle` to enforce secure dependency versions.
  - Patch `org.codehaus.plexus:plexus-utils` to `3.6.1` in `its/build.gradle.kts` to mitigate CVE-2025-67030.


nquinquenel (commented):
Not a big fan of this: 
- It's hardcoded values that are hard to maintain (when should we remove them? how not to forget? etc)
- It's a new pattern that we do not manage in SLI
- Usually, this is how I would handle a dependency risk
	- Is it easy to fix?
		- Yes -> Fix it
		- No -> Are we at risk?
			- Yes -> Fix it with extra effort (this is where I would accept this PR)
			- No -> Accept it with an explanation in SonarQube

PR #1809, opened 2026-06-22
SLI-2724 Configure Renovate post-upgrade task to regenerate lock files

nquinquenel:
----
## Summary by Gitar

- **Renovate configuration:**
  - Added `postUpgradeTasks` to regenerate Gradle lockfiles after dependency updates for `gradle` and `gradle-wrapper`.
  - Configured custom commands to write locks across all project modules using `./gradlew --write-locks`.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -5,5 +5,24 @@
   ],
   "ignorePaths": [
     "its/projects/**"
+  ],
+  "packageRules": [
+    {
+      "description": "Regenerate Gradle lockfiles after dependency updates",
+      "matchManagers": [
+        "gradle",
+        "gradle-wrapper"
+      ],
+      "postUpgradeTasks": {
+        "commands": [
+          "./gradlew --write-locks dependencies :common:dependencies :clion:dependencies :clion-resharper:dependencies :nodejs:dependencies :git:dependencies :rider:dependencies :test-common:dependencies :its:dependencies"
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -5,5 +5,24 @@
...
+  "packageRules": [
+    {
+      "description": "Regenerate Gradle lockfiles after dependency updates",
+      "matchManagers": [
+        "gradle",
+        "gradle-wrapper"
+      ],
+      "postUpgradeTasks": {
+        "commands": [
+          "./gradlew --write-locks dependencies :common:dependencies :clion:dependencies :clion-resharper:dependencies :nodejs:dependencies :git:dependencies :rider:dependencies :test-common:dependencies :its:dependencies"
```
damien-urruty-sonarsource, line 18: Isn't `dependencies` calling all subprojects' `dependencies` tasks? All this list might be redundant

damien-urruty-sonarsource (approved):

.github/renovate.json
nquinquenel: `dependencies` only runs on the root project in a multi-project build. It's explained here [Gradle dependency locking docs](https://docs.gradle.org/current/userguide/dependency_locking.html):
> In a multi-project setup, note that dependencies is executed only on one project, typically the root project.

PR #1821, opened 2026-07-03
SLI-2765 Pre-commit analysis should not block EDT

nquinquenel:
----
## Summary by Gitar

- **Refactored pre-commit analysis:**
  - Wrapped `analyzeFilesPreCommit` in a modal task to ensure non-blocking execution on the EDT.
  - Refactored `SonarLintCheckinHandler` to handle analysis results asynchronously using `runModalTaskWithResult`.
- **Code cleanup:**
  - Simplified `SonarLintCheckinHandler` logic using `switch` expressions and ternary operators.
  - Removed unnecessary `computeOnPooledThread` call in `AnalysisSubmitter` for file module resolution.

### DIFF
--- a/src/main/java/org/sonarlint/intellij/trigger/SonarLintCheckinHandler.java
+++ b/src/main/java/org/sonarlint/intellij/trigger/SonarLintCheckinHandler.java
@@ -48,11 +48,11 @@ import javax.annotation.Nullable;
 import javax.swing.JCheckBox;
 import javax.swing.JComponent;
 import javax.swing.JPanel;
-import org.apache.commons.lang3.tuple.Pair;
 import org.sonarlint.intellij.actions.SonarLintToolWindow;
 import org.sonarlint.intellij.analysis.AnalysisResult;
 import org.sonarlint.intellij.analysis.AnalysisSubmitter;
 import org.sonarlint.intellij.analysis.RunningAnalysesTracker;
+import org.sonarlint.intellij.callable.CheckInCallable;
 import org.sonarlint.intellij.cayc.CleanAsYouCodeService;
 import org.sonarlint.intellij.common.ui.SonarLintConsole;
 import org.sonarlint.intellij.common.util.SonarLintUtils;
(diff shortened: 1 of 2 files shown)


src/main/java/org/sonarlint/intellij/trigger/SonarLintCheckinHandler.java
```diff
@@ -100,17 +101,23 @@ public ReturnResult beforeCheckin(@Nullable CommitExecutor executor, PairConsume
...
-
-      if (analysisIdsByCallback == null) {
+      var preCommitResult = runModalTaskWithResult(project, "SonarQube for IDE Pre-Commit Analysis",
+        indicator -> {
+          var analysisIdsByCallback = getService(project, AnalysisSubmitter.class).analyzeFilesPreCommit(affectedFiles);
+          if (analysisIdsByCallback == null) {
+            return null;
+          }
+          var completed = waitForPreCommitAnalyses(indicator, analysisIdsByCallback.getRight());
+          return Pair.of(analysisIdsByCallback, completed);
```
damien-urruty-sonarsource: Nitpick: Instead of using a pair to return 2 data, that are used together in a same condition later, couldn't we combine them here? We could also introduce a PreCommitResult record that contains the right data right away

src/main/java/org/sonarlint/intellij/trigger/SonarLintCheckinHandler.java
```diff
@@ -48,6 +48,7 @@
 import javax.swing.JCheckBox;
 import javax.swing.JComponent;
 import javax.swing.JPanel;
+import org.apache.commons.lang3.tuple.Pair;
```
damien-urruty-sonarsource: We could also convert the class to Kotlin and use their Pair type

damien-urruty-sonarsource (approved):

src/main/java/org/sonarlint/intellij/trigger/SonarLintCheckinHandler.java
nquinquenel: Added a record, leaving the need for Pair

PR #1837, opened 2026-08-03
SLI-2803 Remove the SQC region selection checkbox

damien-urruty-sonarsource:
----
## Summary by Gitar

- **Configuration:**
    - Removed the SonarQube Cloud region selection checkbox and associated settings from global options
    - Updated server connection management and project binding panels to always display regions for multiple connections

### DIFF
--- a/its/src/test/kotlin/org/sonarlint/intellij/its/utils/ProjectBindingUtils.kt
+++ b/its/src/test/kotlin/org/sonarlint/intellij/its/utils/ProjectBindingUtils.kt
@@ -50,15 +50,17 @@ object ProjectBindingUtils {
         }
     }
 
-    fun enableConnectedMode(projectKey: String, connectionName: String) {
+    fun enableConnectedMode(projectKey: String, connectionName: String, connectionType: ConnectionType) {
         with(remoteRobot) {
             idea {
                 dialog("Project Settings") {
+                    val connectionLabelWithPrefix = connectionType.uiLabelPrefix + connectionName
                     checkBox("Bind project to SonarQube (Server, Cloud)").select()
                     comboBox("Connection:").click()
                     remoteRobot.find<ContainerFixture>(byXpath("//div[@class='CustomComboPopup']")).apply {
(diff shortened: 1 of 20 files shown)


its/src/test/kotlin/org/sonarlint/intellij/its/utils/ProjectBindingUtils.kt
```diff
@@ -50,15 +50,17 @@ object ProjectBindingUtils {
...
             idea {
                 dialog("Project Settings") {
+                    val connectionLabelWithPrefix = connectionType.uiLabelPrefix + connectionName
                     checkBox("Bind project to SonarQube (Server, Cloud)").select()
                     comboBox("Connection:").click()
                     remoteRobot.find<ContainerFixture>(byXpath("//div[@class='CustomComboPopup']")).apply {
-                        waitFor(Duration.ofSeconds(5)) { hasText(connectionName) }
-                        findText(connectionName).click()
+                        // a prefix is added only if there are multiple SQC connection. Check both with and without label
+                        waitFor(Duration.ofSeconds(5)) { hasText(connectionLabelWithPrefix) || hasText(connectionName) }
```
vnaskos-sonar, line 62: It seems that there is no test (unit or integration) with more than one SQC connections, so practically `hasText(connectionLabelWithPrefix)` is always false. It's a test gap that existed before this PR, we could add some tests to assert that the prefixes are applied correctly.

vnaskos-sonar (approved):
All good just one comment for a possible test gap

its/src/test/kotlin/org/sonarlint/intellij/its/utils/ProjectBindingUtils.kt
damien-urruty-sonarsource: There is no test where we intentionally create 2 connections to check if it works. But there is a test suite where multiple connections are created by different tests, and we can end up having more than one connection. I had to fix this snippet to make the test pass. Otherwise I agree with you, we could test this case more intentionally, but I won't do it now

